PR #1, opened 2015-03-05
Documentation

FiniteSource:
Created conventions for contributing to improve readability and ease of maintenance.  At this point a lot of conventions are scrambled e.g. config/configuration.php.  Further guidelines should be created, but this is a good starting point that can be contributed to.


marcelfolaron:
NICE!

PR #87, opened 2020-01-11
Create docker-compose.yml

Cthulhunonamous:
Relative to docker run command;
* Omitted network
* MySQL not exposing ports to host

I don't believe the above are necessary to bring a minimal instance up.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,23 @@
+version: '3'
+services:
+  leantime_web:
+    image: leantime/leantime:latest
+    restart: always
+    environment:
+      - DB_HOST=mysql_leantime
+      - MYSQL_USER=admin
+      - MYSQL_PASSWORD=321.qwerty
+      - MYSQL_DATABASE=leantime
+    depends_on:
+      - mysql_leantime
+    ports:
+      - 80:80
+  mysql_leantime:
+    image: 'mysql:5.7'
+    restart: always
+    environment:
+      - MYSQL_USER=admin
+      - MYSQL_PASSWORD=321.qwerty
+      - MYSQL_ROOT_PASSWORD=321.qwerty
+      - MYSQL_DATABASE=leantime
+    command: --character-set-server=utf8 --collation-server=utf8_unicode_ci


marcelfolaron:
Hi,

Thanks for adding this. I think it makes sense to keep a buildable compose file, but I am wondering if we should expand that to allow a development environment with a local volume. 
Something like this: https://blog.codeship.com/using-docker-compose-for-php-development/

Your thoughts?

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
+services:
+  leantime_web:
+    image: leantime/leantime:latest
+    restart: always
+    environment:
+      - DB_HOST=mysql_leantime
+      - MYSQL_USER=admin
+      - MYSQL_PASSWORD=321.qwerty
+      - MYSQL_DATABASE=leantime
```
MelvinLoos: I would suggest environment variables which can then savely be substitued through an .env file like in my setup: 
https://github.com/MelvinLoos/leantime-docker-compose
```suggestion
      - MYSQL_USER=${DB_USER}
      - MYSQL_PASSWORD=${DB_PASSWORD}
      - MYSQL_DATABASE=${DB_DATABASE}
```

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    ports:
+      - 80:80
+  mysql_leantime:
+    image: 'mysql:5.7'
+    restart: always
+    environment:
+      - MYSQL_USER=admin
+      - MYSQL_PASSWORD=321.qwerty
+      - MYSQL_ROOT_PASSWORD=321.qwerty
+      - MYSQL_DATABASE=leantime
```
MelvinLoos: Same here:
```suggestion
      - MYSQL_DATABASE=${DB_DATABASE}
      - MYSQL_USER=${DB_USER}
      - MYSQL_PASSWORD=${DB_PASSWORD}
```

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
+services:
+  leantime_web:
+    image: leantime/leantime:latest
+    restart: always
```
MelvinLoos: For local dev/testing `restart: always` is quite annoying since it would then start these containers again on machine boot... but then again for production this might be preferable so not sure but putting it out there to discus.

marcelfolaron: I think the primary purpose of the compose file in the main repo should be a dev environment. As such I agree that this should probably be removed.

marcelfolaron:
Docker compose is now available in our docker repo:
https://github.com/Leantime/docker-leantime
Let's continue iterating there

PR #107, opened 2020-02-19
Added dutch translation

Hesselni:
Added dutch translation based on feature-i18nSupport reference file 'en.ini'.


marcelfolaron:
This is awesome! Thank you! I am not quite done adding all translation strings but I'll try to keep new ones to a minimum and add them to the bottom of the file.

resources/language/nl.ini
```diff
@@ -0,0 +1,489 @@
+language.code       = "nl"
+language.direction  = "ltr"
+language.isRTL = "false"
+language.dateformat = "d/m/Y"
```
marcelfolaron: Slashes can be replaced with hyphens or periods if needed. (so d.n.Y or d-m-Y) is a valid format

resources/language/nl.ini
```diff
@@ -0,0 +1,489 @@
...
+language.jsdateformat = "dd/mm/yy"
+language.timeformat = "h:i A"
```
marcelfolaron: This is the american timeformat with am/pm. For 24h timeformat you can use:
H:i (https://www.php.net/manual/en/function.date.php)

resources/language/nl.ini
```diff
@@ -0,0 +1,489 @@
...
+language.jstimeformat = "h:mm A"
```
marcelfolaron: Same as above but in Javascript it's:
HH:mm

resources/language/nl.ini
Hesselni: Good point! I will adjust it.

resources/language/nl.ini
Hesselni: 24h format looks better indeed.

resources/language/nl.ini
Hesselni: Check! I'll adjust it

PR #142, opened 2020-03-04
supports SMTP service without TLS.

evanlabs:


src/core/class.mailer.php
```diff
@@ -78,6 +78,7 @@ public function __construct()
                 $this->mailAgent->SMTPAuth = true;                               // Enable SMTP authentication
                 $this->mailAgent->Username = $config->smtpUsername;                 // SMTP username
                 $this->mailAgent->Password = $config->smtpPassword;                           // SMTP password
+                $this->mailAgent->SMTPAutoTLS = $config->smtpAutoTLS;                 // Enable TLS encryption automatically if a server supports it
```
marcelfolaron: Since we'll have a few configs that don't have that flag yet we should check if it is set at all. 
```
 $this->mailAgent->SMTPAutoTLS = $config->smtpAutoTLS ?? true; 
```

evanlabs: SMTPAutoTLS defaults to true, but sometimes we want to change to false to disable AutoTLS

marcelfolaron: I understand, but the variable `$config->smtpAutoTLS` may not be defined if it is not set in the configuration.php (since it is a new configuration variable many people will not have it right away. Configuration files are also not updates as part of an upgrade process.

evanlabs: Thank you, just realized.

PR #195, opened 2020-04-03
Spanish translation

gonztirado:
My team and I are using your awesome application for planning our tasks. As a reward, we have added this Spanish translation to the Leantime project.

We've used a crowdin.com trial account. If you already have a localization saas that you already use, add me if you want as a Spanish translator.


marcelfolaron:
Thank you so much!!! 

We have been looking for different localization management platforms but haven't made a decision yet. Crowdin looks good to me. I'll add you to it once we have it set up!

PR #223, opened 2020-04-27
some german translations added

ramsonline:
hi, i am pretty sure that i will not have the time to work on this translation in the future. However some hours were spend into this translation already. I hope you can use this already.


marcelfolaron:
Thank you so much! This is great.

PR #301, opened 2020-07-22
Fixed the issue of the logo.png loading through http instead of https…

cstrachan88:
… when appURL is set to https.

This fix involves changing the references to the logoPath from using the `$_SESSION` variable to using the `$config` variable. Examples below:

`<a href="<?=BASE_URL ?>" style="background-image:url('<?php echo htmlentities($_SESSION["companysettings.logoPath"]); ?>')">&nbsp;</a>`

in the header for content.php, login.php, and twoFA.php vs

`<a href="<?=BASE_URL ?>" style="background-image:url(<?php echo $config->logoPath; ?>">&nbsp;</a>`

in the header for install.php, update.php, and resetPassword.php (actually, these pages use both - the `$config` variable in the header to grab the logo.png, and the `$_SESSION` variable further down the page). There is also a typo in the pages using the `$config` variable - no closing parenthesis on the background-image:url setting. Also inconsistent use of quotes for this value.

When loaded, the above resolve to

`<a href="https://URL" style="background-image:url('http://URL/images/logo.png')">&nbsp;</a>`

and

`<a href="https://URL" style="background-image:url(/images/logo.png">&nbsp;</a>`

respectively.

Changing all instances to use the `$config` variable (and adding the missing parenthesis) resolves the mixed-content issue of having the logo be loaded through http.

### DIFF
--- a/src/content.php
+++ b/src/content.php
@@ -25,7 +25,7 @@
 
         <div class="logo" style="<?php if(isset($_SESSION['menuState']) && $_SESSION['menuState'] == 'closed') echo 'margin-left:-260px;'; ?>">
             <a class="barmenu <?php if(!isset($_SESSION['menuState']) || $_SESSION['menuState'] == 'open') echo 'open'; ?>" href="javascript:void(0);"></a>
-            <a href="<?=BASE_URL ?>" style="background-image:url('<?php echo htmlentities($_SESSION["companysettings.logoPath"]); ?>')">&nbsp;</a>
+            <a href="<?=BASE_URL ?>" style="background-image:url(<?php echo $config->logoPath; ?>)">&nbsp;</a>
         </div>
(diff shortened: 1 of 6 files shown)


marcelfolaron:
Sorry for the delay. Was this tested with the logo upload function? I'll have to run a few local tests there are a couple scenarios that were causing problems with the config variable:

Cases to be tested:
Regular install
1. Logopath in config file absolute OR relative
2. Logo uploaded via settings screen

Leantime set up in subfolder install domain.com/leantime/
1. Logopath in config file absolute OR relative
2. Logo uploaded via settings screen

marcelfolaron:
We can't use the config variable as it would prevent users from changing the logo via the UI.
I did find a couple issues that are now fixed. Will close this PR. Thank you for your help!

PR #303, opened 2020-07-22
[BUG] - Files uploaded are not removed from /userfiles/ when deleted …

santibm:
Fixing the file removing bug


marcelfolaron:
Ah, thank you! Great catch!

Issue #304, opened 2020-07-23
[BUG] Login goes to incorrect URL


lamping7:
**Describe the bug**
I have this setup with the official docker image behind nginx. Nginx is serving this up at `http://myhost.myuniversity.edu/projects` which is what I have `LEAN_APP_URL` and the corresponding `appUrl` set to. When trying to login, the `requestUrl` is `'/'`, taking you to `myhost.myuniversity.edu` when you click the "Login" button, which is a 404 for obvious reasons. You can then change the address in the bar to the correct URL (`http://myhost.myuniversity.edu/projects`) and you're logged-in and get to the dashboard. All other functions seem to work as expected after this.

**To Reproduce**
 - Self-explanatory

**Expected behavior**
 - Self-explanatory

**Leantime Version**
2.1.4

**Server**
nginx

**PHP / MySQL Version**
docker image with mariadb 10.4.13

**Additional context**
I have a fix in a coming PR.

marcelfolaron:
Left a comment on the PR. Did you test redirects from emails?

lamping7:
This is still a problem.

marcelfolaron:
I think this is resolved now. Will close. Let me know if you are still running into issues

PR #305, opened 2020-07-23
login: fix redirectUrl

lamping7:
See #304 

Let me know if you have an alternate approach for this that you'd prefer. I made a couple minor formatting changes as well for style consistency.

### DIFF
--- a/src/login.php
+++ b/src/login.php
@@ -70,10 +70,10 @@
                     </div>
                 </div>
                 <div class="regcontent"  style="margin-left: 90px;">
-                    <form id="login" action="<?php echo $redirectUrl?>" method="post">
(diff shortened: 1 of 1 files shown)


lamping7:
I do not have this setup with emails and am not completely sure what you mean. Is there code you can point me to or more info on the use case example you're mentioning? 

Using the _server request_uri for this doesn't work for anyone with a sub folder path, but there may be a better solution than what I've done here.

marcelfolaron:
Added a fix in #370 . Instead of updating redirect url inline I changed it at the top of the file where we handle the redirect logic

lamping7:
👍🏻 Sounds good. I'll give it a try later this week. Thanks.

lamping7:
FYI I finally upgraded to 2.1.7. The login still didn't work and I used my fix in this PR to resolve.

PR #320, opened 2020-08-15
a new admin menu - ui optimisation

carsteneu:
Changes in behaviour and ui of the admin elements - now there is a possibility to open the items in an explicit admin menu
it was somehow not logical that e.g. the user management has menuitems of a current project - this commit will change this.


carsteneu:
@marcelfolaron - would you consider changes of this kind and especially these in the future or does it not fit into your ideas? Otherwise I would not submit such pull requests...

It looks like that:

screenshot-2020 10 20-15_35_50-1603200950300

marcelfolaron:
I like that how do you get to that screen though? Clicking on the profile?

carsteneu:
No, click on a submenu entry of the profile. If you click one of them, it will open this view.

marcelfolaron:
Oh I see, maybe we can add a "back to project" link on the tip to make it
easier to get back?

carsteneu:
Yeah - thats a good idea...  that can be added easely i think

marcelfolaron:
Closing this for now, since there are multiple conflicts and a bit more work to do

PR #323, opened 2020-08-24
Misc features

sogln:
If, for example, the content in the idea overview is larger than can be displayed in the short column view, it would be helpful if the entire content could be displayed without the editor function by tapping the idea heading. 

Link texts in columns of large tables which are outside the columns are displayed outside the columns.


marcelfolaron:
Not sure if you meant to just open a bug. But it seems these commits are from my branch :) Not ready yet. 
Will close this one

PR #325, opened 2020-09-02
Feature/japanese translation

lachelier:
Hi, Thank you for the great system.
I translated it into Japanese.
It may need a few corrections, but you can use it as a draft.


marcelfolaron (approved):

marcelfolaron:
Thank you!

PR #347, opened 2020-11-02
NEW Additonal Project Settings added - now you can add in the Project…

carsteneu:
… Template additional Fields which will be directly usable in all project related situations
includes a db Update and for this we have to update the version to 2.1.5 - but this ist just for testing purpose - feel free to change...

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -17,7 +17,7 @@ class settings {
 
 	public $appVersion = "2.1.4";
 
-    public $dbVersion = "2.1.4";
+    public $dbVersion = "2.1.5";
 
 	/**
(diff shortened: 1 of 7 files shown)


marcelfolaron:
Hey, 
I just now realized that this Pull Request added a new field to the project table while it should have used the settings table. 
We use the table zp_settings to store settings across users and projects using keys in the format: projecsettings.PROJECTID.SETTING_IDENTIFIER = value

I created #375 for the refactor. Would you @carsteneu have time to refactor that? 
I can try to get it done as well, but it might take a little longer.

Thanks.

PR #348, opened 2020-11-02
NEW - Reorder comments

carsteneu:
NEW - Possibility to change the order of the comments...
NEW - now the order sets also the position of the input textarea
NEW - Order of comments now available for comments in tickets

And adds pull request #347 - so you can use it as a patch

… Template additional Fields which will be directly usable in all project related situations
includes a db Update and for this we have to update the version to 2.1.5 - but this ist just for testing purpose - feel free to change...

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -17,7 +17,7 @@ class settings {
 
 	public $appVersion = "2.1.4";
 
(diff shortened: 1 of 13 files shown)


marcelfolaron (approved):
This is awesome! Thank you!

PR #352, opened 2020-11-30
German Translation

Skaldar:
- Completed german translation

- Added informal version of german translation


marcelfolaron (approved):
Thank you for the translation!

Issue #355, opened 2020-12-11
Milestones progress calculations works inverse dependency direction [BUG]


w1naenator:
Milestones progress calculations works inverse direction of dependency (if dependencies used).
Milestones don't "understand" that all To-Dos done if dependencies used.

olenoerby:
To add data to this issue, if a milestone B is dependent on this milestone A, 
milestone A won't be marked as 100% done before all it's To-Do's are marked as done and milestone B is marked as done as well.
![image](https://user-images.githubusercontent.com/6160947/102546768-7077ed80-40b8-11eb-999a-f8c20ab66467.png)

w1naenator:
Seams last github master branch contain improved version, now, if all milestones marked as done, to-dos progress displayed correctly.
![milestone](https://user-images.githubusercontent.com/2381733/102550847-3cf08f80-40c7-11eb-90a4-ac7d751a1b32.png)
![todos](https://user-images.githubusercontent.com/2381733/102551142-affa0600-40c7-11eb-8631-405c58d03105.png)
![milsetone_settings](https://user-images.githubusercontent.com/2381733/102551152-b4262380-40c7-11eb-9028-39cd95eb1021.png)

olenoerby:
I am running docker leantime/latest version 2.1.4, so this is possible I might be behind on version.

marcelfolaron:
I see the issue now. The progress on dependency does seem to be reversed,

PR #370, opened 2021-01-26
Fix misc issues

marcelfolaron:


marcelfolaron:
Fixing 361 and #355 as well as logo bug (231, 301) in subfolder installations

PR #374, opened 2021-01-30
FIX Fixed the problem of missing psettings when initializing zp_projects

yokazek:
I'm new to PR, so I don't know if this is the right branch name and how to do it, but I had a problem with a new install, so I fixed it.

It seems that psettings is missing when initializing zp_projects.


marcelfolaron (approved):

marcelfolaron:
Thanks I am merging this to unblock users but we'll have to remove psettings from the DB. I missed when that was committed

Issue #312, opened 2020-07-31
[BUG] In the new user form in the project assignment part the project name doesn't pass into view 


LightTemplar:
**Describe the bug**
In the new user form I have projects to assign, but I can't distinguish between them, because their client name takes all place in the view. 
![image](https://user-images.githubusercontent.com/453783/89047936-69418480-d34f-11ea-88f0-b3426c1e72f0.png)

**Expected behavior**
I don't know. May be two lines per project, or some hint, when mouse is over it? Or let me change the view width. The main problem is how to find which line is which project?

**Leantime Version**
 2.1.4

**PHP / MySQL Version**
It's complicated. Is there some debug info, accessible per click?

marcelfolaron:
Thanks for submitting. I'll try to get a line break in there.

carsteneu:
Maybe it is solved by commit be4cd49cdc1315f59f4aa7cbed2fd85ed396d316??

marcelfolaron:
Thanks, it should help to see the project name first, but once project names are too long we run into a similar issue. I'll try to see if we can wrap it or just change the UI to use checkboxes

carsteneu:
Yeah - but i think project names are most of the time not that similar as company names... we have hundreds of projects - not even 2 are similar in the first letters ... i think this would be a very special situation if project names are that similar.

LightTemplar:
@NoobTW, I have just tried it and haven't seen any multiline names. Was it somehow removed from the code?

marcelfolaron:
Fixed in 2.1.9

PR #441, opened 2021-08-17
feat: make project multiselect multiline

NoobTW:
Show multiline if the "project name / project client" is too long.

Before:

image

After:

image


NoobTW:
#312

PR #454, opened 2021-09-02
Update class.editUser.php

b90g:
https://github.com/Leantime/leantime/blob/4978c6a9c5ff647e15cc65b143f922b4b14f13e4/src/domain/users/controllers/class.editUser.php#L68

i changed it that way and at least now the page is loading again.

did the 2.1.8 work for anybody?

### DIFF
--- a/src/domain/users/controllers/class.editUser.php
+++ b/src/domain/users/controllers/class.editUser.php
@@ -65,7 +65,7 @@ namespace leantime\domain\controllers {
 								'hours' => ($_POST['hours']),
 								'wage' => ($_POST['wage']),
 								'clientId' => ($_POST['client']),
-								'password' = ($row['password']),
+								'password' => ($row['password']),
 							);
 
 							$changedEmail = 0;


marcelfolaron (approved):

marcelfolaron:
Thank you!

dirkbez:
> https://github.com/Leantime/leantime/blob/4978c6a9c5ff647e15cc65b143f922b4b14f13e4/src/domain/users/controllers/class.editUser.php#L68
> 
> i changed it that way and at least now the page is loading again.
> 
> did the 2.1.8 work for anybody?

Version 2.1.8 (downloaded on 2021-12-16 from the website) still has the error.

JorgeCarousel:
In my case using 2.1.8 the line was already that way, but I still copied and pasted it, but nothing happened, I still see a blank page when I try to edit any user.

unknown797:
> In my case using 2.1.8 the line was already that way, but I still copied and pasted it, but nothing happened, I still see a blank page when I try to edit any user.

The solution is 'password' => ($row['password']), 

It needs that > added after =

PR #455, opened 2021-09-04
Added Hebrew translation

yarons:
Hey, this is the full Hebrew translation but I do have some comments:
1. Days of the week, Months, Decimal separator, Thousands separators, currency, these are all part of CLDR, no need to translated them manually.

2. What do you think about a localization platform? I personally love Weblate as it's completely open source and provides free hosting for open source projects.

If you think it's a good idea I'd love to assist with the migration, we will have broader translators audience and this project might see some new languages.


marcelfolaron (approved):

marcelfolaron:
Thank you! Yes, I'd love to get on a better translation platform!

yarons:
Great! How do you want to meet?

PR #458, opened 2021-09-10
Added missing month July in french language

parazitenew:
I noticed that July was missing in calendar (french language), and an "undefined" month was displaying after December, that caused a gap between month number and month index (since an array is indexed starting 0), I just added "Juillet" in the months array inside fr-FR.ini to correct the issue, for french users.


BENETNATH (approved):
That's a mandatory fix :)

BENETNATH (approved):

hartigdan (commented):
lgtm

Issue #485, opened 2021-10-18
[BUG] - Menu doesn't have a scrollbar, can't list all the projects in small monitors


matheus-rossi:
*If you have problems with the installation please use our [community forum](https://community.leantime.io)*

**Describe the bug**
When we have a lot of projects, in the same client, the menu doesn't show them all.
As a workaround i'm changing the zoom of the page to view them

**To Reproduce**
[![Animacao.gif](https://s8.gifyu.com/images/Animacao.gif)](https://gifyu.com/image/eTvT)

**Expected behavior**
See all the projects, maybe with a scrollbar ?

**Leantime Version**
2.1.8

**Server**
Apache

PR #486, opened 2021-10-18
Fix Menu doesn't have a scrollbar bug #485

mohabmes:
Fix https://github.com/Leantime/leantime/issues/485


matheus-rossi:
tested the fix in my base, working 100%

Issue #502, opened 2021-11-05
[FEATURE] see all my pending tasks 


previmeteo:
**Is your feature request related to a problem? Please describe.**
have a single "my tasks" page with the entire list of my pending tasks wheras they are in different projects
**Describe the solution you'd like**
view like /tickets/showAll but for all the active projects
**Additional context**
none

clemrdh:
It is implemented in the current v3.2 release ?

PR #515, opened 2021-11-25
Add show All  Tickets Per Project

davask:
Proposed solution for that ticket https://github.com/Leantime/leantime/issues/502


marcelfolaron:
Awesome, I'll take a look this week. Sorry for the delay on that.

marcelfolaron:
Hi @davask,

Thank you for your PR. I like the idea to show projects + tickets but I think the table in the current layout would get very large quite quickly. Instead I would probably group the tasks by project as it is set up on /tickets/showAll (Check the grouping options). We could still show project information in the grouped header.

Thank you

javaindian:
Hi @marcelfolaron

Any roadmap when it could be incorporated? I couldn't find it in v2.2.

I tried to manually do this. My bad, i couldn't succeed. Hence rolled back.

Thanks

corentiny1:
Same here, it's important to see all tasks in the same place and don't have to go through all the projects to try to get an overview. Maybe a kanban view with tasks labeled by the name of the project ?

broskees:
I'm happy to work on this if you have an idea for implementation that works for you. @marcelfolaron

clemrdh:
Still not implemented ? It would be very usefull

PR #540, opened 2021-12-27
set UTF-8 for both SMTP and php Mail

Beexx:
UTF-8 charset was set only in SMTP mode :

```
if($config->useSMTP === true) {
   ...
   $this->mailAgent->CharSet = 'UTF-8';
}
```

I've moved it above the "if" statement so that it applies to both SMTP and php Mail(). Character encoding for accentuated characters is now correct even with the php Mail() builtin function (it was not the case before).


marcelfolaron:
Good catch!

marcelfolaron (approved):

PR #542, opened 2021-12-30
Improve fr-FR.ini further 

Beexx:
**fr-FR.ini** : 
I've corrected a few misspelled words, and I've added some missing strings.
I've also kept the English original terms for some concepts, like "Backlog" and "Lean Canvas". As those terms are part of the Lean Canvas methodology, it looked weird to translate them in french as "Arriéré" and "Canvas Lean"...
I also translated "milestone" as "jalon", which is the accurate translation in french for this. If you prefer "milestone" (the english word), please go back to it. Milestone is also understood in french.


marcelfolaron:
Thank you!

marcelfolaron (approved):

PR #711, opened 2022-05-17
Prevent access to installation page after installation

### LINKED ISSUE #704: [BUG] Installation page still accessible after installation
**Describe the bug**
After using the `/install` route to do the initial setup - which works fine -, I can still open the installation page afterwards. Either on `/install` or `/index.php?install=true`. Though I didn't test whether the installation can be overridden by "installing again", this is a high vulnerability in my opinion.

**To Reproduce**
1. Install Leantime
2. Go to `/install` & fill out the form
3. Go to `/install` again

**Expected behavior**
Going to `/install` after initial setup should either show a 404 page or just redirect to the `/`.

**Leantime Version**
2.1.9

**Server**
OpenResty

**PHP / MySQL Version**
PHP: 7.4
MySQL: 8.0.25-15 (Percona Server (GPL), Release '15', Revision 'a558ec2')

**Additional context**
Some additional checks should be implemented here:

https://github.com/Leantime/leantime/blob/e8376104b1f44a22e2a7ffa8717302dd80ace086/src/core/class.application.php#L66

Edit:
**Possible fixes**

- Add a row to the database table `zp_settings`: `installed` = `true`
- Check if a local file (e.g. `LEANTIME_INSTALLED`) exists

axelrindle:
Closes #704 

This change will redirect users to the main page when trying to access `/install` and the installation process has already finished.

### DIFF
--- a/src/install.php
+++ b/src/install.php
@@ -4,6 +4,11 @@ $config = new leantime\core\config();
 $settings = new leantime\core\settings();
 $install = new leantime\core\install($config, $settings);
 
+if ($install->checkIfInstalled()) {
+    header('Location: '. BASE_URL);
+    exit;
+}
+
 ?>
 <!DOCTYPE html>
 <html dir="<?php echo $language->__("language.direction"); ?>" lang="<?php echo $language->__("language.code"); ?>">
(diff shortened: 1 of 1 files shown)


marcelfolaron:
Thank you!

PR #730, opened 2022-06-07
Lossless compression of images

phulstaert:
Lossless compression of images with png optimizer tools.


phulstaert:
Repository went from ~30MB to ~18MB. This effect is felt by every page served.

marcelfolaron (approved):

marcelfolaron:
Oh wow, thank you. It does make me realize that there are a bunch of files in there that we do not need anymore. 
I'll try to clean that up.

PR #781, opened 2022-07-12
Fix array typing issue in reports

TareqK:


TareqK:
fixes the error ``PHP Fatal error:  Uncaught Error: Cannot use object of type leantime\domain\models\reports as array in /var/www/html/src/domain/sprints/services/class.sprints.php:160``

marcelfolaron (approved):

PR #842, opened 2022-08-03
Zip not readable in Windows

phulstaert:
Issues is the "leantime/." path in the zip command. Just leantime like in the tar is good.
Also 
- Removed the createReleasePackage.sh file that is in the repository now
- Added the -p paramter to mkdir release (only if directory does not already exists)
- Remove leftover leantime directory before pulling the latest from Github (should never happen, but we both know Murphy).
- Added documentation

Open questions
==========
Is the .idea directory needed for release?

### DIFF
--- a/createReleasePackage.sh
+++ b/createReleasePackage.sh
@@ -1,31 +1,33 @@
 #!/bin/sh
 
-mkdir release
+# Creating folder for release build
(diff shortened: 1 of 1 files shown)


marcelfolaron (approved):
Thank you!!!

PR #843, opened 2022-08-03
Easy version in createReleasePackage.sh

phulstaert:
This pulls the version out of the appSettings.php file, so you don't have to add it as a parameter.
You can still add a parameter; this will be added after the version 
(createReleasePackage "-alpha" would result in "Leantime-v2.2.4-alpha.zip" (and tar.gz)

This needs to be tested on a mac before being accepted.


marcelfolaron:
Oh nice! Thank you!

PR #847, opened 2022-08-05
first steps towards PSR-12 compliance

phulstaert:
With a PSR-12 checker as my guide, I went to fix the most obvious formatting issues. Still have a lot to do regarding the soft limit of 120 lines and a more impactful missing of PascalCase. Have tested it by going over all tabs and pages. No errors found. (also should not happen since this is only formatting)


marcelfolaron:
Which tool did you use to create that? Can we re-run it a little later? I have a few structural changes that would probably end up creating a bunch of conflicts with this PR

phulstaert:
Yes, we can defenetly run it at a later time again.
I used PHP_CodeSniffer. 

```
#Install by downloading the phar file
wget https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar

#Run by this command
php ~/phpcb.phar --standard=PSR12 ./
```

After some trail and error (a few hours) I went to read the manual (because real men don't read manuals :-))
And it appears there is a version that fixes the easy errors for you.

```
#Install by downloading the phar file
wget https://squizlabs.github.io/PHP_CodeSniffer/phpcbf.phar

#Run by this command
php ~/phpcbf.phar --standard=PSR12 ./
```
After the understandable facepalm, I run this command on the code. Reviewed every type of correction it does and tested the code. It all works fine.
There are however errors/warnings it does not fix on its own, but it displays a message about it.

It may not be much, my opinion is that it is a step in the right direction.

PR #848, opened 2022-08-05
Add support for port in pdo mysql

bankzst:
Description:
I recently had issue with database configuration that is not using default 3306 port. So, I tried adding the port support in these files.

**configuration.sample.php**
- new port variable.
- new environment variable for port.

**class.db.php**
- new port variable
- add port to existing pdo connection string

**class.install.php**
- new port variable
- add port to existing pdo connection string

**backup.php**
- add port flag to mysqldump command

pl. sorry if I made any mistake, this is my first github PR, all advice are welcome.

### DIFF
--- a/config/configuration.sample.php
+++ b/config/configuration.sample.php
@@ -19,6 +19,7 @@ class config
     public $dbUser = ''; //Database username
     public $dbPassword = ''; //Database password
     public $dbDatabase = ''; //Database name
+    public $dbPort = '3306'; //Database port
 
     /* Fileupload */
     public $userFilePath = 'userfiles/'; //Local relative path to store uploaded files (if not using S3)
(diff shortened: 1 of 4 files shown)


marcelfolaron (changes requested):
Thank you, this is great. One thing though: We'll need to add the default value into the individual classes. A lot of people won't update their config file during a regular system update, so I assume that would break their connection. 

Probably just a matter of setting the property to 3306 in the individual classes (class.db.php, backup.php, class.install.php) and then do a `$config->dbPort ?? $this->port`

phulstaert (commented):
Seems ok; I'll update the formatting to PSR-12 later.

bankzst:
> Thank you, this is great. One thing though: We'll need to add the default value into the individual classes. A lot of people won't update their config file during a regular system update, so I assume that would break their connection.
> 
> Probably just a matter of setting the property to 3306 in the individual classes (class.db.php, backup.php, class.install.php) and then do a `$config->dbPort ?? $this->port`

I already added the default value into both class.db.php and class.install.php. But I have one question about backup.php

In the configuration.sample.php, I added the `$this->dbPort` with default value 3306 along with other database variables. And this value will be initiated on constructor with the help of _configEnvironmentHelper_ if LEAN_DB_PORT env variable is not set by using value from `$this->dbPort`.

Isn't this enough to make the value always exist for `$config->dbPort` in backup.php?

PR #860, opened 2022-08-11
Clean up application, auth and session handling

marcelfolaron:
- Use of central layout files
- Move login, install into actual domain modules and remove individual files
- application and frontcontroller updated to remove custom install, login, resetpassword update etc files
- Allow public controllers
- Streamline auth handling


src/core/class.application.php
```diff
@@ -46,61 +57,59 @@ public function __construct(config $config,
...
+            $telemetryResponse = $this->reportService->sendAnonymousTelemetry();
 
-        ob_start();
+            // Check if trying to access twoFA code page, or if trying to access any other action without verifying the code.
+            if($_SESSION['userdata']['twoFAEnabled'] && $_SESSION['userdata']['twoFAVerified'] === false){
 
-        if($this->login->logged_in()===false) {
+                if($this->frontController::getCurrentRoute() !== "twoFA.verify"
+                    && $this->frontController::getCurrentRoute() !== "auth.logout"
+                    && $this->frontController::getCurrentRoute() !== "api.i18n") {
```
funkywaddle, line 84: `if(!in_array($this->frontController::getCurrentRoute(), ['twoFA.verify','auth.logout','api.i18n']))` might be cleaner than 3 separate if checks. Not 100% sure though

src/core/class.application.php
```diff
@@ -174,47 +178,68 @@ public function overrideThemeSettings() {
...
+        }
         //Only run this if the user is not logged in (db should be updated/installed before user login)
-        if($this->login->logged_in()===false) {
+        if($this->auth->logged_in()===false) {
 
-            if($this->settingsRepo->checkIfInstalled() === false && isset($_GET['install']) === false){
-                if(!isset($_GET['act']) || $_GET['act'] != 'api.i18n') {
-                    header("Location:" . BASE_URL . "/install");
-                    exit();
+            if(!isset($_SESSION['isInstalled']) || $_SESSION['isInstalled'] === false) {
```
funkywaddle, line 215: isn't this already covered by the fact that the first if didn't return?

marcelfolaron: Good catch. You are right

src/core/class.session.php
```diff
@@ -102,6 +105,14 @@ public static function getInstance()
...
 
+        if (self::$instance === null) {
+
+            self::$instance = new self();
+            return self::$instance::$sid;
+
+        }
+
+
         return self::$sid;
```
funkywaddle, line 116: What's the difference between `self::$instance::$sid` and `self::$sid`?

src/core/class.template.php
```diff
@@ -7,7 +7,9 @@
 
 namespace leantime\core {
 
+    use JetBrains\PhpStorm\NoReturn;
```
funkywaddle, line 10: am I missing where this is getting used?

src/domain/install/controllers/class.index.php
```diff
@@ -0,0 +1,100 @@
...
+                    'email' => ($params['email']),
+                    'password' => $params['password'],
+                    'firstname' => ($params['firstname']),
+                    'lastname' => ($params['lastname']),
+                    'company' => ($params['company'])
+                );
+
+                if (isset($params['email']) == false || $params['email'] == '') {
+                    $this->tpl->setNotification("notification.enter_email", "error");
+                } else {
```
funkywaddle, line 71: I believe `if->elseif->elseif` is slightly more efficient and easier to read than a series of nested `else->if->else->if` statements

src/core/class.template.php
marcelfolaron: Removed

PR #903, opened 2022-08-22
Language update

phulstaert:
Removed duplicate entries from language files
Added missing entries to other languages (using en_US as default)
Noticed that french has a "text.project_success" no other language had... placed a remark on GitHub.


marcelfolaron (approved):
Thank you!

PR #922, opened 2022-08-26
Optimize loops + formating + useldap as boolean

phulstaert:
- reduce the amount of configEnvironmentHelper from 42 to 20 in a lot of cases (do not read S3 settings if useS3 is not true; same for ldap and mail)
- formatting closer to PSR-12
- useLdap set to boolean instead of undefined (=>string)

### DIFF
--- a/config/configuration.sample.php
+++ b/config/configuration.sample.php
@@ -1,67 +1,66 @@
-<?php /** @noinspection PhpMultipleClassDeclarationsInspection */
+<?php
+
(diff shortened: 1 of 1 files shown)


phulstaert:
FYI: the dashboard loads in average in 17ms on my machine (only the body, not css, JS, etc..) With these changes, this is gets to 15ms because I don't use LDAP, S3 nor mail.

marcelfolaron (approved):

marcelfolaron (approved):

PR #1021, opened 2022-10-12
Proposed corrections for various bugs

diderich:
If a class from a vendor package matches the naming convention of `leantime`, loading the class fails because the system assumes that it is a `leantime` class.

=> Update code to only load a `leantime` class if the file exists.

There exist a type bug in the comments repository

=> correct it.

Chevron buttons is create/edit user did not show.

=> Added missing 'pa' class.


marcelfolaron:
Thankl you!

PR #1046, opened 2022-10-18
Fixed editor role cannot create or edit to-do in showKanban

sinarajabpour1998:
Fixed a bug that user with editor role cannot create, edit or move cards in showKanban


marcelfolaron:
Hey, thanks for the PR. 
The changes you made were specific exceptions for admins, owners and managers only. 
Editors as user role should not receive the original role. In practical terms it is possible for a user to be editor on project A but readonly in project B. This is by design and allows managers the flexibility to add people as readonly to some projects and as editors to others.

PR #1050, opened 2022-10-19
Fix calendar issue (update)

diderich:
The original fix I suggested (and was implemented) is only partially correct. It incorrectly sets `$tickets` to empty if not tickets exist for `thisWeek` and `later`.


marcelfolaron (approved):
Thank you! Good catch!

PR #1075, opened 2022-10-25
Added two config options: 'smtpAuth' and 'smtpSSLNoverify'

ruhnet:
The 'smtpAuth' option allows user/password authentication to be disabled when using SMTP (it is still enabled by default). The 'smtpSSLNoverify' option allows disabling certificate checking for SMTP SSL/TLS connections, so that self-signed or expired certificates will be accepted. This in effect disables authentication but keeps encryption on a [semi] secure connection. Since SSL/TLS is not universally supported for email delivery, and some mail servers offer SSL/TLS, but only with old or self-signed certificates, adding the capability to Leantime makes sense. The default is still to verify and require a properly secure and authenticated connection when SSL/TLS is being used.


marcelfolaron:
Thank you for the PR. Looks good so far but can you add checks to see if the variables are set? A lot of people don't update the configuration file and we don't have a good mechanism to inject configuration file changes yet.

ruhnet:
Ah good point---ok I will implement that to make sure there won't be any problems.

ruhnet:
I've added the checks, so it should be safe if a user upgrades without changing their existing config file.

marcelfolaron:
Thank you!

PR #1082, opened 2022-10-27
Make top right menus UI consistent

diderich:
Add a caret-down next to the new company management / administration menu on the top right to show that it is a drop-down menu, rather than a function, consistent with the profile menu next to it.

Remove the name from the profile menu item (only show the image) to make it consistent with the overall menu items that only show icons, rather than text.

An alternative would be to have a text for all items (at it was the case in some old version). I prefer the new set-up.


marcelfolaron:
Thanks, that makes sense to me as well (and yeah while text is more accessible, the icons are just so much cleaner. )

PR #1134, opened 2022-11-14
unfinished start of twig POC

broskees:


src/core/class.frontcontroller.php
```diff
@@ -328,18 +333,38 @@ public static function getCurrentRoute()
...
+         */
         public static function redirect($url, $http_response_code = 303): void
         {
 
             header("Location:".trim($url),true, $http_response_code);
             exit();
         }
 
+        /**
+         * Sets the response code for the current request
```
marcelfolaron, line 354: Thank you for adding the comments!

marcelfolaron:
Could we get rid of the entire old template class and just use twig for all of that or do you see a reason to keep it?

broskees:
> Could we get rid of the entire old template class and just use twig for all of that or do you see a reason to keep it?

Yeah I mean I think thats where we want to get to. The only reason why I left it there is because I wanted old templates to work in the mean time.

YasinAshorian (approved):

PR #1188, opened 2022-12-10
Changed timesheet labels to default format.

### LINKED ISSUE #1185: [BUG] Wrong Date in timetracking chart with different locale
**Describe the bug**
Problem with the Time Tracking when language is set to German. In the Ticket, the month and day are in the wrong order. It should be the 6th and 7th of December but it is June and July. So I guess the data for the Chart is Wrong.

**To Reproduce**
1. Go to company settings and change the Language to German
2. Go to a Ticket which has Time tracked

**Expected behavior**
The Chart should show the correct date values

**Leantime Version**
V2.2.11

image
**Example**: The Date is _6th of December_ but in the graph it is shown as _12th of June_

Niki2k1:
Fixes #1185 

Chart.js does not do localization. So passing the date in a different form than the en_US format does not work.
Changed to en-US default format.

### DIFF
--- a/src/domain/tickets/templates/submodules/timesheet.sub.php
+++ b/src/domain/tickets/templates/submodules/timesheet.sub.php
@@ -71,7 +71,7 @@
         foreach ($this->get('ticketHours') as $hours){
             $sum = $sum + $hours['summe'];
 
-            echo"labels.push('".date($this->__("language.dateformat"),  strtotime($hours['utc']))."');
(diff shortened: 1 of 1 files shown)


marcelfolaron:
It was actually quite simple. Chart.js does localize via moment js and it is set but in the label push we parsed the date coming from the database as local date which we shouldn't have: See https://github.com/Leantime/leantime/pull/1190/files

Thank you

Issue #1319, opened 2023-01-20
LDAP options clean-up


sgissi:
**Describe the bug**
The LDAP settings need cleanup as current settings have no effect and can mislead users. I'll be happy to provide a clean-up patch.

**Leantime Version**
2.3.5 / master

---

**ldapUserDomain**
Becomes userDomain at [LDAP class](https://github.com/Leantime/leantime/blob/master/app/domain/ldap/services/class.ldap.php#L65). It was used at [Auth class](https://github.com/Leantime/leantime/blob/master/app/domain/auth/services/class.auth.php#L208) but commented now. Not used anywhere else.

**bindUser / bindPassword**
Defined at both [configuration.sample](https://github.com/Leantime/leantime/blob/master/config/configuration.sample.php#L70-L71) and [defaultConfiguration](https://github.com/Leantime/leantime/blob/master/app/core/class.defaultConfiguration.php#L69-L70). It seems tied to the old LDAP implementation. It is not currently present on LDAP class. All current LDAP binds use user-provided credentials.

**baseDn**
Defined but not used anywhere. Probably used in the old LDAP implementation. All LDAP searches are bound by the ldapDN setting instead.

marcelfolaron:
Hi, thanks. You are right, some clean up is required. 
Happy to review the patch if you have time. I won't be able to get to this until later this week

sgissi:
Thanks Marcel. I opened #1329 to clean it up.

The [docs](https://github.com/Leantime/docs/blob/master/installation/configuration.md#ldap-configuration-beta) says LDAP is in Beta. Let me know when you consider it stable and I'll be happy to update the documentation to match the current LDAP configuration.

PR #1329, opened 2023-01-23
LDAP: Cleanup unused parameters

sgissi:
Removes LDAP parameters from configuration that are unused and could mislead users.

Signed-off-by: Silvio Gissi <silvio@gissilabs.com>

### DIFF
--- a/app/core/class.defaultConfiguration.php
+++ b/app/core/class.defaultConfiguration.php
@@ -63,11 +63,7 @@ class config
     public $ldapType = 'OL';                              //Select the correct directory type. Currently Supported: OL - OpenLdap, AD - Active Directory
     public $ldapHost = '';                                //FQDN
     public $ldapPort = 389;                               //Default Port
-    public $baseDn = '';                                  //Base DN, example: DC=example,DC=com
     public $ldapDn = '';                                  //Location of users, example: CN=users,DC=example,DC=com
-    public $ldapUserDomain = '';                          //Domain after ldap, example @example.com
-    public $bindUser = '';                                //ldap user that can search directory. (Should be read only)
-    public $bindPassword = '';
     //Default ldap keys in your directory.
     //Works for OL
     public $ldapKeys = '{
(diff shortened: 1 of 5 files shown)


marcelfolaron (approved):

b90g (commented):
- Documentation is missing how leantime is supposed to login into LDAP
- How was this unused?!
- how does lean time now login to ldap to check user credentials?..

sgissi:
Hi @b90g, good point about documentation, I noticed that in #1319. LDAP is listed as "Beta" in the documentation. Once Marcel and team considers it stable, I'll be happy to propose a PR for the docs.

The unused parameters ended up like that as LDAP implementation changed over time:

- All searches are bound by LdapDN, so BaseDN is not used.
- User domain was joined with the username to make the e-mail, however the actual email is an user attribute.
- Bind user and password are not used anymore because the user supplies their own credentials at both Leantime login and import pages.

The idea behind the cleanup was exactly to avoid luring users into setting things that have no effect. Please let me know if you caught any issues as a result of the clean-up and I will propose a fix.

b90g:
Thanks for the quick reply.

I see. My understanding of LDAP is very little and i was relying on the concept of a bind authentication user to much. (nextcloud, zammad, wekan..)

I will try to understand the config where in the user authenticates against ldap directly.

sgissi:
No worries, I spent quite some time in the code to figure them out. The only call to ldap_bind comes from the LDAP Service class, in the [bind function](https://github.com/Leantime/leantime/blob/master/app/domain/ldap/services/class.ldap.php#L115).

The two places that I have found calls the LDAP Service bind function are:
- [Import](https://github.com/Leantime/leantime/blob/master/app/domain/users/controllers/class.import.php#L63) at the users controller - Using the username and password from the form
- [Login](https://github.com/Leantime/leantime/blob/master/app/domain/auth/services/class.auth.php#L206) at the auth service - Using the username and password from the login screen

I hope I didn't miss anything in the search!

b90g:
UTC+1 and tired:

I notice the default ldap client config requires valid certificates.
I use univention coporate servers openldap implementation which has selfsigned certificates.

Maybe an option to override the invalid certificate would help? not sure if this applies

sgissi:
That is a great catch @b90g. Actually the way we use ldap_connect is passing host and port, that ends up with only unencrypted connections and is [deprecated](https://www.php.net/manual/en/function.ldap-connect.php). The correct way is to pass the URL (ldap://host:port or ldaps://host:port). Besides that, if you are using a self-signed certificate, you will have to point to a copy of the public CA file (https://www.php.net/manual/en/ldap.constants.php#constant.ldap-opt-x-tls-cacertfile).

Do you mind opening a new issue to track that feature request? This PR is closed and will not have the right visibility.

PR #1444, opened 2023-02-17
remove sub-url install as default from .htaccess

TareqK:
this is to prevent confused users and dysfunctional docker images


TareqK:
closed since its done in #1445

PR #1467, opened 2023-02-25
Changes to remove dynamic properties.

acul009:
I tried to add some of the dynamic properties that are used to the actual class definitions.
This might not be compatible with all your supported PHP-versions.

If you tell me the lowest PHP-version you want to support, I could update the code.

### DIFF
--- a/app/core/class.application.php
+++ b/app/core/class.application.php
@@ -16,6 +16,7 @@ class application
     private services\projects $projectService;
     private repositories\setting $settingsRepo;
     private services\reports $reportService;
+    private services\api $apiService;
 
     private IncomingRequest $incomingRequest;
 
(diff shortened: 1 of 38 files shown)


acul009:
If you want to continue my work, I used the following regex for most of it:

\$this->(.+) = new (.+)\(.*

and replaced it with

private $2 \$$1;

marcelfolaron:
Thank you! We are going with the official PHP long term support cadence: https://www.php.net/supported-versions.php
So right now the lowest is 8.0 but we'll EOL that in November.

acul009:
Wow, than I could have made some of the type definitions a bit tighter.

Would you accept pull requests which specify types on currently generic attributes?
While that might produce more visible errors, in a previous project those definitions caught some errors, that would have been devastating and helped increase code stability in the long term.

marcelfolaron:
Yes, definitely. Tightening up types would help a lot!

marcelfolaron (approved):

marcelfolaron:
Thank you so much!

PR #1475, opened 2023-03-01
Feature/new leantime command to test email sending

jfp-gravity:
I was flummoxed with testing the email sending system and couldn't find an easy way to do it within the app without ending up with trouble, so I cooked up a rapid diagnosis command to add to the leantime command-line tool, and that allowed me to solve the issue, so I figured I would contribute it back.

It temporarily increases logging levels of the SMTP Mailer library to return as much debug feedback as possible.

### DIFF
--- a/app/command/class.testEmailCommand.php
+++ b/app/command/class.testEmailCommand.php
@@ -1,0 +1,60 @@
+<?php
+
+
+namespace leantime\command;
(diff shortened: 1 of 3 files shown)


marcelfolaron (approved):

marcelfolaron:
Thank you. This is awesome and so helpful!

PR #1491, opened 2023-03-10
Basic OpenID Connect (OIDC) Support

acul009:
Hi,

this is my current prototype of OIDC authentication.
I tested this configuration with Google, Github (example in env) and authentik.

There are still some things which would be good to have:

1. OIDC state creation and verification - protection against CSRF
2. Support for parsing jwks signing keys from e and n - x5c already works.  Google provides these only on https://www.googleapis.com/oauth2/v1/certs so you need to verride the JWKS url
3. The login only works on OIDC providers which provide the email. The system will log you in using the email - any provider which does not verify the user email will therefore create quite the easy exploit.

It might also requite some nicer formatting and some code cleanup, but you have some working code to build upon.

If you need some more functionality or some bug fixes before merging, just tell me :)


app/domain/auth/templates/login.tpl.php
```diff
@@ -33,6 +33,14 @@
         <div class="">
             <input type="submit" name="login" value="<?php echo $this->language->__("buttons.login"); ?>" class="btn btn-primary"/>
         </div>
+        <?php if($this->get('oidcEnabled')) { ?>
```
marcelfolaron, line 36: Can we move this to the top to afterRegContent? 
The reason I ask is we have the Google Auth plugin built by @m8than 
He added the button to the top (https://github.com/m8than/googleSSO-leantime/blob/f0c44fa29c6817b30ed3427ba1ac966d7cad521c/register.php#L107) and we should keep the login providers in the same place.

app/domain/oidc/controllers/class.callback.php
```diff
@@ -0,0 +1,31 @@
...
+
+    public function get($params)
+    {
+        $code = $_GET['code'];
+        $state = $_GET['state'];
+        try {
+            $this->oidc->callback($code, $state);
+        } catch(Exception $ex) {
+            echo '<pre>';
+            echo $ex->getMessage();
```
marcelfolaron, line 26: Let's write this into the error logs using `error_log($ex)`

app/domain/oidc/services/class.oidc.php
```diff
@@ -0,0 +1,398 @@
+<?php
```
marcelfolaron, line 1: @TareqK Can you review this file in particular. I don't have enough knowledge with OIDC. To me it all looks good.

marcelfolaron (changes requested):
This is great. Just 2 requests

app/domain/oidc/services/class.oidc.php
TareqK: It seems fine to me, should work on most providers. However, we need to add PSK verification since a lot of providers use rs256 PSK

marcelfolaron:
Thanks for the changes. Can you resolve the conflicts real quick and I'll merge the PR.

acul009:
I've only resolved SVN merge conflicts before. I might have messed up something, so please check before mergin.
Sorry for the trouble.

marcelfolaron:
Thank you lgtm

Issue #1556, opened 2023-04-09
[BUG] leantime 2.3.15 : UI translation not working anymore


jmassiaux77:
After 2.3.14 to 2.3.15 leantime update, UI translation don't work anymore (except in the dropdown selection where translation is well working).

- I have tried with several langage (french, deutch, spanish) and each time with logout+login and navigator cache flush.
- I have tried with old (2.3.14) fr-FR.ini langage file but same problem.

<img width="977" alt="Capture d’écran 2023-04-09 à 11 59 54" src="https://user-images.githubusercontent.com/123368774/230766516-f51e4ecb-a218-4c4c-9574-e28b0cc5bb9d.png">

jmassiaux77:
2.3.16 patch tested : it's well working 👍

PR #1558, opened 2023-04-09
Bug Fixes

marcelfolaron:
- Clean up effort keys
- Fix language loading issue #1556


tedmarcin:
[Remove ROOT from path since it was added to the constant](https://github.com/Leantime/leantime/pull/1558/commits/ea1bd62a3fc3205c9fa35963bce02f9000c2806e)

Most translations works after patch , but not pl_PL Polski (pl)? Any idea why?

PR #1636, opened 2023-05-05
Password hash at User Repository level

sia-mfierro:
Hi,
I found a case (editing a user via RPC call) where the password was not hashed properly.

I thought that it might be a good idea to simply have the hashing happen in the user repository, so that one doesn't have to remember about it.

I shouldn't have broken anything, but I haven't been able to test the code, so I highly advise  a review.

Thanks for the amazing software.


marcelfolaron:
Thank you. That is really helpful and makes sense.

PR #1656, opened 2023-05-19
Conditional change password by edituser 

cholletk:
Hi,
I was facing issue when editing a user from an Owner account. This was leading to a password change (due to a double hashing of password) and the modified user was not anymore able to connect. This is probably a side effect of the PR #1636 

I made some checks to not update the password if not provided to editUser function and removed the value from the editform.

I was not able to do extended check, so it this change has to be taken with care


marcelfolaron:
Thank you. This is helpful.

PR #1679, opened 2023-06-08
Fixing issue with field project and todo in my timesheets view

cholletk:
Destroying chosen and rebuilding it when there is a change (issue with the trigger)


marcelfolaron:
Thank you!

PR #1692, opened 2023-06-13
Add 'ID' column to the task list

phavel:
I am sending a proposal to display a column with the task ID.


marcelfolaron (approved):
Thank you!

Issue #1528, opened 2023-03-30
[FEATURE] Comments, Discussions and Notifications - Need To Be More In Harmony With Expectations


compumatter:
**Is your feature request related to a problem? Please describe.**
Employees often do not find the right place to add discussions, can't find discussions, don't get notified on time with "time sensitive" notations.

**Describe the solution you'd like**
* I would like "Discussions" not hidden in Project Settings.  
* I would like to see Todo "comments" in its own tab and called Task Discussions. 
* I would like to see the @theirname used in Discussions also used in Task Discussions
* Whenever @theirname is used, @theirname should be notified of that discussion note immediately.
* Comments are currently being emailed to a team member only when that team member is logged in / active.  That means time-sensitive comments are often missed completely as team members are mobile and not logged into Leantime.  At the very least, configuration.php should have a setting that allows emails to be sent every X minutes via a cron or just sent out when comments are made.   Not when a team member is logged in / active.
* Discussion notes even when using @theirname are currently not sent even when the team members are logged in.  Perhaps this is a cron.  @theirname should get an email notification immediately and other team members via a configurable cron time.

**Additional context**
Overall just finding a disconnect on Discussions vs Comments so I've created a video to make more obvious my thoughts and our hopes in this regard.
* https://www.youtube.com/watch?v=rUg7c3aL4vA

gloriafolaron:
We've had a discussions around a few of these and the disconnect of some of the others is really helpful. This is definitely something that promotes use by teams and having them turn off notifications becomes counter-intuitive.

Thanks for putting this together so thoroughly.

Issue #1505, opened 2023-03-16
Email Notifications Feature/Update Request


Auren-Frost:
Hello, I have been using Leantime for task organization and a way to collaborate with people here at CompuMatter. I decided to turn email notifications on to inform me of any changes that have occurred every three hours.

Example email: [https://i.imgur.com/J9ZQ5kQ.png](https://i.imgur.com/J9ZQ5kQ.png)

These emails come in even when I am the only one making changes. It would be great if it either only sent updates when other people made changes or if there was an additional notification setting to turn off email notifications for your own changes. I do not really see the need of being notified for editing my own work.

gloriafolaron:
Thanks for adding this! This is one we've been talking about internally as well -- improving email notifications. Will definitely take a look.

PR #1698, opened 2023-06-14
#1505 #1528 Fix: Email notifications about changes are sent to the change owner.

phavel:
Correct usage of the 'notifyProjectUsers' method with the line 'return $user != $notification->authorId;', see: https://github.com/Leantime/leantime/blob/98f0bb4af4763c44a007942e6b8d6c40b79b3a1f/app/domain/projects/services/class.projects.php#L171

### DIFF
--- a/app/domain/projects/services/class.projects.php
+++ b/app/domain/projects/services/class.projects.php
@@ -135,7 +135,7 @@ namespace leantime\domain\services {
             //Only users that actually want to be notified and are active
             foreach ($users as $user) {
                 if ($user["notifications"] != 0 && strtolower($user["status"]) == 'a') {
-                    $to[] = $user["username"];
+                    $to[] = $user["id"];
                 }
             }
 


app/domain/projects/services/class.projects.php
```diff
@@ -135,7 +135,7 @@ public function getUsersToNotify($projectId)
             //Only users that actually want to be notified and are active
             foreach ($users as $user) {
                 if ($user["notifications"] != 0 && strtolower($user["status"]) == 'a') {
-                    $to[] = $user["username"];
+                    $to[] = $user["id"];
```
marcelfolaron, line 138: Did you test this thoroughly? 
I am trying to wrap my head around this and I know that the queue will check for email or ID. So that should be fine. Are there any other cases where we actually expect the email address instead of ID? 

I agree that we need to make sure the author check is correct which it is not right now (Great catch on that one!)

phavel: Unfortunately, I did not get to the proper testing. I've only known the system for a few days, I don't have a general knowledge of what all needs to be tested.
However, in the IDE I went through places that use this method and all places should handle the ID correctly (I hope I'm not mistaken).
Please check on your side.
I just know that with this modification, my mails go correctly when I add/change tasks.
Thank you

PR #1732, opened 2023-07-05
fix save bug on goal canvas

broskees:


app/domain/goalcanvas/controllers/class.editCanvasItem.php
```diff
@@ -281,7 +279,7 @@ public function post($params)
 
                     $this->projectService->notifyProjectUsers($notification);
 
-                    $this->tpl->redirect(BASE_URL . '/strategyPro/editCanvasItem/' . $_GET['id']);
```
marcelfolaron, line 284: Oops. Thank you!

marcelfolaron (commented):
Does that fix the save & close logic as well?

broskees:
> Does that fix the save & close logic as well?

I couldn't reproduce that. I tried every which way. For me Save & Close just works, unless you were talking about the "please include title" error. However, I also fixed the dashboard not displaying the various goal canvas items.

PR #1885, opened 2023-09-15
Usability improvements

marcelfolaron:
- Timer is now using htmx, no reload required
- Added timeline menu link back
- Table UX enhancements
- started emojiinput component
- file uploads are now part of main menu.


app/Domain/Tickets/Js/ticketsController.js
```diff
@@ -1222,7 +1222,54 @@ leantime.ticketsController = (function () {
...
+                        }, 0 );
+
+
+                    // Update footer by showing the total with the reference of the column index
+                    jQuery( api.column( 9 ).footer() ).html('Total');
+                    jQuery( api.column( 10 ).footer() ).html(plannedHours);
+                    jQuery( api.column( 11 ).footer() ).html(hoursLeft);
+                    jQuery( api.column( 12 ).footer() ).html(loggedHours);
+
+                },
```
broskees, line 1272: Any way we can easily move this to HTMX rather than adding more JavaScript?

app/Domain/Tickets/Templates/Submodules/timelineTabs.sub.php
```diff
@@ -0,0 +1,29 @@
...
+function findActive($route)
+{
+    if (str_contains(\Leantime\Core\Frontcontroller::getCurrentRoute(), $route)) {
+        return "active";
+    }
+    return "";
+}
+
+findActive("");
+?>
```
broskees, line 14: This makes me wonder if we can have composers for php templates. I see no reason why we can't. Let's start moving stuff to composers for partials.

On another note though, can this be a blade template?

app/Domain/Tickets/Templates/showKanban.tpl.php
```diff
@@ -146,39 +146,12 @@
...
-                                                                                                                                   echo sprintf($tpl->__("links.stop_work_started_at"), date($tpl->__("language.timeformat"), time()));
-                                                                                                                               }?></a>
-                                                                    <span class='working' <?php if ($clockedIn === false || $clockedIn["id"] === $row["id"]) {
-                                                                        echo"style='display:none;'";
-                                                                                          }?>><?php echo $tpl->__("text.timer_set_other_todo"); ?></span>
-                                                                </li>
-                                                            </ul>
-                                                        </div>
-                                                    <?php } ?>
+                                                    <?php echo app("blade.compiler")::render('@include("tickets::partials.ticketsubmenu", [
```
broskees, line 149: Missing closing parentheses

app/Domain/Tickets/Templates/showAll.tpl.php
```diff
@@ -359,57 +359,11 @@
...
-                                                        echo"style='display:none;'";
-                                                    } ?>
-                                                >
-                                                    <?php echo $tpl->__("text.timer_set_other_todo"); ?>
-                                                </span>
-                                            </li>
-                                        </ul>
-                                    </div>
-                                <?php } ?>
+                                <?php echo app("blade.compiler")::render('@include("tickets::partials.ticketsubmenu", [
```
broskees, line 362: Missing closing parentheses

broskees (changes requested):

app/Domain/Tickets/Js/ticketsController.js
marcelfolaron: hm unfortunately I don't think that is possible in this case. This is part of the table initialization. It creates a sum at the bottom of the table for datatables. With column visibility being a frontend only picker it would probably break. I can check when I create a table component in the future

app/Domain/Tickets/Templates/Submodules/timelineTabs.sub.php
marcelfolaron: Do we even need composers, the ugly way would be to just rename the file to blade and wrap the whole thing in @php right?

marcelfolaron: or alternatively just replace <?php with @php and @endphp

PR #1916, opened 2023-09-26
refactor environment class and move events execution to earlier for m…

broskees:
…ust use plugins


app/Core/Environment.php
```diff
@@ -388,16 +77,23 @@ class Environment implements ArrayAccess, ConfigContract
      */
     public function __construct(DefaultConfig $defaultConfiguration)
     {
+        unset($_SESSION['mainconfig']);
```
marcelfolaron: debug?

marcelfolaron (approved):

PR #1959, opened 2023-10-06
start of plugin marketplace

broskees:


app/Core/Bootloader.php
```diff
@@ -516,13 +519,14 @@ private function bindRequest(): void
...
-        if (isset($headers['Hx-Request'])) {
-            $incomingRequest = $this->app->instance(IncomingRequest::class, HtmxRequest::createFromGlobals());
-        } elseif (isset($headers['X-Api-Key'])) {
-            $incomingRequest = $this->app->instance(IncomingRequest::class, ApiRequest::createFromGlobals());
-        } else {
-            $incomingRequest = $this->app->instance(IncomingRequest::class, IncomingRequest::createFromGlobals());
-        }
+        $incomingRequest = $this->app->instance(
+            IncomingRequest::class,
+            match (true) {
```
marcelfolaron, line 524: I had refactored this because it was not working. For some reason it did not return an instance of API Request for api requests. But it did with the if statements above.

PR #1996, opened 2023-10-17
Fix List icons in German Language

Stiles96:
Change not working class from "fa-list" to "fa fa-list".


marcelfolaron (approved):
Thank you!

PR #2025, opened 2023-10-23
feat(login form): hide form if config is true

### LINKED ISSUE #1908: [FEATURE] Allow user to disable the default log in form
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I install Leantime, set OIDC log in, and log in with the user which has the same email address as admin, so that they can be linked, and in the future I can use OIDC to directly log in. However, after this the default log in form won't let me log in anymore, it becomes useless, I will suggest give user an option to disable the default log in form.

**Additional context**

eric-hansen:
Closes #1908

I don't use Docker so I can't run the acceptance test(s) locally.


marcelfolaron (approved):
Thank you! Looks great

PR #2032, opened 2023-10-24
User customizable datetime

### LINKED ISSUE #1945: [FEATURE] date-time formats
**Is your feature request related to a problem? Please describe.**
The problem is that date and time formats are related to selected locale in settings. For example some english terms are sometimes better to use than localized national translation. In the same time using non 24h format and m/d/y sequence is not good

**Describe the solution you'd like**
 It would be better if ISO yyyy-mm-dd or dd-mm-yyyy will be used or there will be option to choose date-time formats separately from language

eric-hansen:
Closes #1945 

The front-end date time looks like the correct JS format now (based on the pre-defined values in editOwn.php):

image

This also resolves the original issue that caused the PR to be reverted before as well.


app/Domain/Users/Controllers/EditOwn.php
```diff
@@ -254,5 +262,37 @@ public function post(): void
...
+         * @link https://www.php.net/manual/en/class.datetimeinterface.php#datetimeinterface.constants.types
+         * 
+         * @return [<string,string>] Format of ID => date-time string
+         */
+        private function getSupportedDateTimeFormats(): array
+        {
+            return [
+                'dates' => [
+                    $this->language->__("language.dateformat"),
+                    'Y-m-d',
```
marcelfolaron, line 277: Is this working for you? When I try to save an updated string I get a fatal error that the array index cannot be found on editOwn.

eric-hansen: Are you sure you have the latest changes?  I was able to switch formats just fine and render them both with PHP calls and JS.

marcelfolaron: Let me double check

marcelfolaron: Now I'm getting a fatal error:

[28-Oct-2023 11:25:09 America/Los_Angeles] 2023-10-28T11:25:09-07:00 [critical] Uncaught Target class [Leantime\Core\ServicesSetting] does not exist. (View: /Volumes/Source/Sites/leantime-oss/app/Views/Templates/layouts/entry.blade.php)
[28-Oct-2023 11:25:11 America/Los_Angeles] 2023-10-28T11:25:11-07:00 [critical] Uncaught Target class [Leantime\Core\ServicesSetting] does not exist. (View: /Volumes/Source/Sites/leantime-oss/app/Views/Templates/layouts/entry.blade.php)

marcelfolaron: Seems to be link 602 in language: 

        $settings = app()->make(ServicesSetting::class);

marcelfolaron: After instantiating the right class and trying to navigate to editOwn, I get this. Can you pull from master and check if everything is working on your end? 

image

eric-hansen: Sure I'll check that tomorrow.  Makes sense now that I think about it as to how you were having issues.

marcelfolaron: Thank you

eric-hansen: Pushed up 44a77f8 as somehow the template setters got removed.  I tested with the settings deleted from the DB (so a user who hasn't set this before) and they rendered properly, and I was able to save them.

marcelfolaron: Thanks. I made a couple fixes and created PR into your branch. 
1. It seems you were pulling directly from the db every time a dateformat needed to be output. I went ahead and cached the result in the session
2. Timeformat was not returned
3. Added missing language strings to en-US.

Please take a look and merge into your branch (which should update this PR here automatically.

PR #2037, opened 2023-10-25
Update README.md

bhargavshirin:


bhargavshirin:
some terms and names are case-sensitive and should be spelled as they are originally written

marcelfolaron (approved):

PR #2046, opened 2023-10-28
Update README.md

Killer2OP:
# Description
Added the tech stack badges in readme file.


marcelfolaron:
Thanks, but I don't see a lot of value in adding those badges. They take up quite a bit of room and the techstack is also available on the right side of github. Aside it feels rather incomplete given that we require a lot more in our techstack.

PR #2054, opened 2023-10-30
Align db versions with actual version number

marcelfolaron:


TimexPeachtree (approved):
Awesome 😎 this makes it more leaner.

PR #2090, opened 2023-11-15
Adds IIS-specific installation note to README.md

MeatyFresh:
This adds a step-by-step GUI-based guide on enabling the PATCH method in IIS, which is often not enabled by default (afaik.)

Without the PATCH method being accepted by the server some small functionalities within LeanTime don't work quite right, primarily the tutorial popups keep appearing on every page after each authentication/log on even if the user clicks the "close and don't show again" link, as this preference is set via a PATCH.

Unfortunately this is an IIS thing and I can't find a way to enable this in the codebase only (via web.config for example) without also touching the IIS config somehow. This method as written seems like to simplest to distribute to IIS users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,23 @@ There are two main ways to install LeanTime for production. The first of which i
 * Navigate to `<yourdomain.com>/install`
 * Follow instructions to install database and set up first user account
 
+##### IIS Installation Notes #####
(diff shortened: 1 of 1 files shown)


marcelfolaron (approved):
Thank you

PR #2107, opened 2023-11-21
Improved Slack and Mattermost notifications

jontorrado:
The current Slack and Mattermost notifications were not sending much information. For example, when moving a card, it just said "Task was updated", but nothing else was shown. So I tried Trello's integrations and updated the Leantime with more information. Previous notifications were like this (Spanish):

screenshot1

To this:

screenshot2

So, now:
- The summary just shows what happened, without the ticket ID
- The clickable area is just the task name
- The project is shown in bold
- The status of the card is shown below

More information may be added easily, but I don't think a client, for example, needs to know the priority or other internal labels.

Here you can also see a comment:

screenshot3

And the same goes for Mattermost:

screenshot4


marcelfolaron (approved):

marcelfolaron:
Thank you!

Issue #2114, opened 2023-11-24
[BUG] Date formatting method returns empty strings on API calls


jeppekroghitk:
**What set up are you using**
* [ ] Cloud Hosted Version
* [x] Self Hosted

**Describe the bug**
This problem is possibly more extensive than this case im presenting here, but here goes:

When calling the "getSprint" endpoint, the dates startDate and endDate are run through the method _getFormattedDateString_ in the language class. Ultimately, the _getCustomDateTimeFormat_ is called, to get the correct formatting for the date.

*Open your code and goto the getCustomDateTimeFormat method in Language.php:618*
https://github.com/Leantime/leantime/blob/f8091a9f4eb992101406059ea58faa1f9963941b/app/Core/Language.php#L603-L624

When calling from the API, the session is unset, which passes the first if statement.
Next line, the settings object is instantiated.
Next, the result array is built, with data from the ini file.
Next, the userId is set, because every API user has an userId, which enters the userId if statement.
The format are gotten from the settings, which is NULL for the API user.

Both $results['date'] and $results['time'] is set to FALSE, as user settings are NULL.

The session is set with this data.

From now on, every time an API call gets to this method, both date and time format is false and returned from the session at the beginning of the method.

To sum up, _getFormattedDateString_ ultimately returns date(false, {timestamp})
which evaluates to an empty string.

We need to come up with a solution to this problem, as every date returned from the API (given it's run through getFormattedDateString) will return empty strings. Furthermore, i don't like the fact that dates returned from functions are pre-formatted, or relies on a format defined in user settings or the .ini file. The format of a date should be finally defined where it's used (say the template file). Atleast when using the API.
Let's say that i build a lot of custom code that relies on the Leantime API, and the date format returned by the API is determined by what is defined in the .INI file. All my code could break, if the date format in the .ini file is updated.

**To Reproduce**
Call the getSprint($id) endpoint via the API.
See that the returned dates are empty strings.

**Expected behavior**
The API should return the dates correctly.

**Leantime Version**
2.4.7

**Server**
N/A

**PHP / MySQL Version**
N/A

**Additional context**
N/A

PR #2115, opened 2023-11-24
Added method for determining API context

jeppekroghitk:
Proposed fix for bug report:
https://github.com/Leantime/leantime/issues/2114

This PR proposes a (breaking) change to the API. The added method checks if a given method is called in an API context, and therefore return the date in an predetermined format, what can be relied on.

As of now, all dates formatted through getFormattedDateString via methods called from the API will return empty strings. This PR serves to fix this issue, and furthermore introduce a way to determine whether the code is called from an API context, which doesn't have the SESSION variables or database content in it's user row, that the system relies on.

Please see the connected bug report and this PR as a suggestion to how this issue could be fixed. Let me know if there is anything else i can do to help, or you want to talk this through.


app/Core/Language.php
```diff
@@ -80,6 +80,11 @@ class Language
      */
     public Theme $themeCore;
 
+    /**
+     * @var Apirequest $apiRequest
+     */
+    public Apirequest $apiRequest;
```
marcelfolaron, line 86: Can you update the name to be ApiRequest (with capital R) the acceptance tests fail because it can't find the file (linux systems are case sensitive)

marcelfolaron (changes requested):

app/Core/Language.php
jeppekroghitk: It is now fixed. Sorry bout that.

marcelfolaron (approved):

app/Core/Language.php
marcelfolaron: No worries, thanks

PR #2160, opened 2023-12-21
Add Ticket Filter Search Term

suntel-nagamine:
### Editing details
The search term that used to be there seems to have disappeared with the design modifications, so I added it to the end of the filter.


marcelfolaron:
Thank you!

PR #2169, opened 2024-01-01
prod updates

marcelfolaron:


app/Core/HttpKernel.php
```diff
@@ -48,7 +48,9 @@ public function handle($request)
             return $e->getResponse();
         } catch (\Throwable $e) {
             if (! app()->make(Environment::class)->debug) {
-                return new RedirectResponse(BASE_URL . "/errors/error500", 201);
+                echo "An error occured";
+                error_log($e);
+                //return new RedirectResponse(BASE_URL . "/errors/error500", 201);
             }
```
broskees, line 54: It would be better if we just checked for the redirect loop here like so:

```php
if (
    ! app()->make(Environment::class)->debug
    && Frontcontroller::getCurrentRoute() !== 'error.error500'
) {
    error_log($e);
    return new RedirectResponse(BASE_URL . "/errors/error500", 201);
}
```

app/Core/Template.php
```diff
@@ -100,44 +132,35 @@ class Template
...
-        private AuthService $login,
-
-        /** @var Roles */
-        private Roles $roles,
-
-        /** @var \Illuminate\View\Factory|null */
-        public ?Factory $viewFactory = null,
-
-        /** @var CompilerInterface|null */
-        private ?CompilerInterface $bladeCompiler = null,
```
broskees, line 131: I cleaned all this up before, why did we undo it?

app/Core/Theme.php
```diff
@@ -751,14 +752,16 @@ public function setSchemeColors($colorscheme)
...
 
         if (isset($this->colorSchemes[$colorscheme]["secondaryColor"])) {
             $secondary = $this->colorSchemes[$colorscheme]["secondaryColor"];
-            $_SESSION["usersettings.colorScheme.secondaryColor"] = $secondary;
         }
 
+        $_SESSION["usersettings.colorScheme.primaryColor"] = $primary;
+        $_SESSION["usersettings.colorScheme.secondaryColor"] = $secondary;
+
+
```
broskees, line 764: I think we have some merge conflicts issues. I fixed this before too so that it can't throw an error.

app/Domain/Calendar/Repositories/Calendar.php
```diff
@@ -53,12 +56,13 @@ class Calendar extends RepositoryCore
          *
          * @access public
          */
-        public function __construct(DbCore $db, LanguageCore $language, DateTimeHelper $dateTimeHelper)
+        public function __construct(DbCore $db, LanguageCore $language, DateTimeHelper $dateTimeHelper, Environment $config)
```
broskees, line 59: Time to break this into multiple lines.

broskees: Also, let's do `__construct(private DbCore $db)` when we touch constructors.

app/Domain/Calendar/Templates/ical.tpl.php
```diff
@@ -8,7 +8,7 @@
 $calendars = $tpl->get('calendar');
 $eol = "\r\n";
 $env = app()->make(Environment::class);
```
broskees, line 10: Is $env still being used?

app/Core/Template.php
marcelfolaron: not entirely sure, honestly. Reverting

app/Core/Theme.php
marcelfolaron: yeah that must have been a bad merge. Reverting. The change important here was line 342, and just setting an id if it can't find the theme. Theme is not so crucial that a failure to load the theme should stop the application. We just revert to the default theme.

app/Core/Template.php
```diff
@@ -130,14 +126,24 @@ public function __construct(
...
+            app()->call([$this, 'setupBlade']);
+            $this->attachComposers();
+            $this->setupDirectives();
+            $this->setupGlobalVars();
         }
-
-        app()->call([$this, 'setupBlade']);
-        $this->attachComposers();
-        $this->setupDirectives();
-        $this->setupGlobalVars();
```
broskees, line 140: This has gotta go back to the way it was too.

app/Domain/Calendar/Repositories/Calendar.php
```diff
@@ -49,23 +41,27 @@ class Calendar extends RepositoryCore
...
+            private DbCore $db,
+            private LanguageCore $language,
+            private DateTimeHelper $dateTimeHelper,
+            private Environment $config)
         {
             $this->db = $db;
             $this->language = $language;
             $this->entity = "calendar";
             $this->dateTimeHelper = $dateTimeHelper;
+            $this->config = $config;
```
broskees, line 62: You don't need any of these when you use the new constructor format. Also, $this->config doesn't exist.

app/Domain/Tickets/Hxcontrollers/Subtasks.php
```diff
@@ -23,40 +24,98 @@ class Subtasks extends HtmxController
...
      *
      * @param Timesheets $timesheetService
      * @return void
      */
-    public function init(Tickets $ticketService): void
+    public function init(Tickets $ticketService, Language $language): void
     {
         $this->ticketService = $ticketService;
+        $this->language = $language;
     }
```
broskees, line 39: We can clean this up too.

broskees (changes requested):

app/Domain/Tickets/Hxcontrollers/Subtasks.php
marcelfolaron: How so? this is not a constructor.

PR #2175, opened 2024-01-05
Changes/for custom fields and more

broskees:


app/Core/Support/Cast.php
```diff
@@ -36,14 +43,30 @@ public function castTo(string $classDest, array $constructParams = []): object
             $name = $property->getName();
 
             if (! isset($sourceObj->$name)) {
+                if ($property->hasDefaultValue()) {
```
marcelfolaron, line 46: Should have asked before but will $properties always be a countable/iterable var? Otherwise the foreach loop will fail

marcelfolaron (approved):

PR #2226, opened 2024-01-18
update db no matter what...

marcelfolaron:


marcelfolaron:
In this case I am actually swallowing any exceptions and continue the migration. It seems we got out of sync with those 2 columns and some folks have it others don't. I am just running them and if they fail, nothing will happen and your db version will be updated to 3.0.2

PR #2238, opened 2024-01-23
Download files issue when using OIDC login provider

sy-gpe:
Hello !

There is currently an issue with the file download script which returns the "No Access" image for file downloads when the users are logged in through an OIDC provider.

This seems to be related to the fact that the download script does not load the full application, and some settings are missing.

This patch loads the app settings in the bootstrap_minimal_app function.

This has been tested and is working on a production instance with OIDC login.

I don't know if this is the best way to implement this. Tell me if there are changes to do.

Thanks !

Best regards

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -4,6 +4,7 @@ use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Support\Str;
 use Illuminate\View\Factory;
 use Leantime\Core\Application;
+use Leantime\Core\AppSettings;
 use Leantime\Core\Bootloader;
 use Leantime\Core\Language;
 use Leantime\Core\Support\Build;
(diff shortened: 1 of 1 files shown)


marcelfolaron (approved):
@broskees Can you take a look as well?

marcelfolaron:
We recently made a change to defer setting the is bootstraped guard. Can you check if that latest change (in master) fixed the issue already?

sy-gpe:
Hello !

Thanks, I will setup a test installation and get back to you as soon as possible.

Best regards

sy-gpe:
Hello !

I just have tried on the master branch - commit 92363e4e53ea64456f4a6953bcc8e418e91497dd - and the issue is still present.

My patch still fixes it.

I will stay available if you have questions.

Best regards

marcelfolaron (approved):

PR #2241, opened 2024-01-25
Fix cropped textbox content effected by rounder textbox border.

bankzst:
**Description**

This PR provide fix of textbox size in leantime v3 for the new leantime UI that comes with rounder textbox border. This result in some of the textbox contents were cropped out. Please refer to screen capture below

- Project to-do status sort order textbox (existing status and add new status)

Before
image

After
image

- Date textbox (new project, project detail, ticker detail)

Before
image

After
image


marcelfolaron:
Thank you!

PR #2267, opened 2024-02-01
Added milestone and tags to all timesheet view

cableman:
Added milestone and tags column to all timesheet view (which also exports them in the CVS export, which then can be used to make pivot-tables using these fields).

Screenshot from 2024-02-01 14-58-10


marcelfolaron (approved):
Thank you!

PR #2302, opened 2024-02-09
Translations for formal and informal German

Elaktrato:
I tried to complete the German translations for both the informal and formal versions.


marcelfolaron:
Thank you!!

macidev:
will this be persistent for the future since crowdin serves the translations?

PR #2314, opened 2024-02-12
Addressing issue regarding timezone change

jeppekroghitk:
TLDR: 
Core/Support/DateTimeHelper:281 - convertDateTime()
converts 05/02/2024 00:00:00 to 2024-02-04 23:00:00 when converting from UTC to Europe/Copenhagen, slowly but surely shifting the dates when rendering the timesheet overview calendar.

Screenshot 2024-02-12 at 16 22 25

As of version 3.0.6, being in the Europe/Copenhagen timezone, navigating the "My Timesheets" page by jumping from week to week caused some issues in the table layout.

When navigating backwards one week, then forward one week multiple times, this happens:
(Notice how Monday the 12th Febuary becomes wednesday and then friday)

Initial:
Screenshot 2024-02-12 at 16 18 13

Backward 1 week:
Screenshot 2024-02-12 at 16 18 18

Forward 1 week:
Screenshot 2024-02-12 at 16 18 24

Backward 1 week:
Screenshot 2024-02-12 at 16 18 29

Forward 1 week:
Screenshot 2024-02-12 at 16 18 35

### DIFF
--- a/app/Core/Support/DateTimeHelper.php
+++ b/app/Core/Support/DateTimeHelper.php
@@ -293,19 +293,16 @@ class DateTimeHelper
         $dateTime = trim($dateTime);
 
         if ($this->isValidDateString($dateTime)) {
-            $fromDateTimeZone = new \DateTimeZone($fromTz);
+            $fromDateTimeZone = new DateTimeZone($fromTz);
 
             //If it is an API Request, don't do any formatting, return ATOM in target tz
             if ($this->apiRequest->isApiRequest()) {
                 $utcDate = new DateTime($dateTime, $fromDateTimeZone);
                 return is_object($utcDate) ? $utcDate->format(DateTime::ATOM) : false;
             }
+            $timestamp = DateTime::createFromFormat($fromFormat, $dateTime, new DateTimeZone($toTz));
 
-            $timestamp = DateTime::createFromFormat($fromFormat, $dateTime, $fromDateTimeZone);
-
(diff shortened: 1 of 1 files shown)


app/Core/Support/DateTimeHelper.php
```diff
@@ -293,19 +293,16 @@ public function convertDateTime(
...
         if ($this->isValidDateString($dateTime)) {
-            $fromDateTimeZone = new \DateTimeZone($fromTz);
+            $fromDateTimeZone = new DateTimeZone($fromTz);
 
             //If it is an API Request, don't do any formatting, return ATOM in target tz
             if ($this->apiRequest->isApiRequest()) {
                 $utcDate = new DateTime($dateTime, $fromDateTimeZone);
                 return is_object($utcDate) ? $utcDate->format(DateTime::ATOM) : false;
             }
+            $timestamp = DateTime::createFromFormat($fromFormat, $dateTime, new DateTimeZone($toTz));
```
marcelfolaron, line 303: It seems that the "fromTimezone" entry is not being taken into consideration anymore (unless it's an API Request). Which means this would fail for a any date save objects from other forms. 

The goal of this method is to convert a date from one timezone+format into a different timezone+format. It's deliberately made to be very flexible. We don't want to rely on the "default" timezone set by the server since different users can set different timezones. 

I think the issue with the timesheets is probably somewhere in the timesheet class rather than in the datetime helper.

PR #2318, opened 2024-02-12
Filter for notifications

marcelfolaron:
Adding a filter for notifications for pre processing (and turning off) individual notifications via plugins


marcelfolaron:
@jeppekroghitk @cableman You can use this filter to check what type of notification is being processes and then only allow mentions + assigns

PR #2319, opened 2024-02-12
Timesheet issue

marcelfolaron:
Timezone drifting on timesheet button due to native "format" function being used instead of our datetimehelper


cableman:
I have tests this by patching our live sites and it works.

PR #2325, opened 2024-02-14
Updated code style for API classes - part II

cableman:
#### Link to ticket

N/A

#### Description

This is the next part of the API classes code style fixes.

#### Screenshot of the result

N/A

#### Checklist

- [x] My code passes all test cases.
- [x] My code passes our static analysis suite.
- [x] My code passes our continuous integration process.

If your code does not pass the requirements on the checklist, you should add a comment explaining why this change 
should be exempt from the list.

#### Additional comments or questions

N/A

PR #2328, opened 2024-02-14
Added PR template for easier review process

cableman:
Don't known if the project want this, but it has helped me in other open source projects to have at PR template, that automatically fill the PR description with common content.

See https://github.com/Leantime/leantime/pull/2325 for an example of this template (not a very good filled out example, but an example :-) )


marcelfolaron:
Oh thank you. This is great!

Issue #2322, opened 2024-02-13
[BUG] Timesheet: time registrations on wrong date


lilosti:
**What set up are you using**
* [ ] Cloud Hosted Version
* [X] Self Hosted

**Describe the bug**
Adding a task and putting in 2 hours on monday on my timesheets, saving I see my registration on sunday in the same week.

**To Reproduce**
1. Go to 'Timesheets'
2. Select at task to register time on, register 2 hours on monday
3. Press save
4. See your 2 hours have not been saved on monday (but on sunday)

Try registering time on any day and have fun guessing what will happen...

**Expected behavior**
Time registered on monday is expected to stay on monday

**Leantime Version**
3.0.6

**Server**
Apache, nginx, IIS etc

**PHP / MySQL Version**
Please list the PHP and Mysql version you are using

**Additional context**

marcelfolaron:
Fixed in 3.1.0-beta

PR #2361, opened 2024-02-29
Timesheet fixes

cableman:
#### Link to ticket

https://github.com/Leantime/leantime/issues/2322

#### Description

Look into timezone handling and the conversion of dates. 

* Switched to using carbon (https://carbon.nesbot.com)  for date handling in timesheets
* Ensure date are in UTC in all PHP code and first converted on display in UI to local timezone and format.
* Added helper to `format()` to help during transition to carbon (as this is only change in timesheets for now)

#### Screenshot of the result

N/A

#### Checklist

Builds on top of PR https://github.com/Leantime/leantime/pull/2354 (which have been merged by now)

### DIFF
--- a/app/Domain/Tickets/Controllers/ShowTicket.php
+++ b/app/Domain/Tickets/Controllers/ShowTicket.php
@@ -2,6 +2,7 @@
 
 namespace Leantime\Domain\Tickets\Controllers {
 
+    use Carbon\Carbon;
     use Illuminate\Contracts\Container\BindingResolutionException;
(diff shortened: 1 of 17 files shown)


marcelfolaron:
Thank you. Appreciate all the work that went into this. 
We'll need to make sure to update the format() function and also all the other places using the datetimehelper. 
I think we can refactor the date() and time() methods to do that. We won't need the 'long' format as there is no need for it on the frontend (it's either a date field or a time field, never both).

PR #2386, opened 2024-03-11
Refactoring timesheets

marcelfolaron:
#### Description

Refactor of timesheets building on top of @cableman work
- new datetimehelper
- new format helper (now focusing only on string conversions) 
- new carbon macros
- weekly timesheets are now grouped in php
- 

#### Outstanding
- find and replace all format(..., 'short') with format(.. ,?) where ? is the enum value to be parsed
- final testing
- 

#### Checklist


app/Core/Support/FromFormat.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * For value containing both date and time in users preferred format and timezone separated by a space
+     */
+    case UserDateTime;
+
+    /**
+     * for values containing only the user formatted date and timezone. Adds start of day time to string
+     */
+    case UserDateStartOfDay;
```
cableman, line 25: Do we need these start and end of day?

app/Domain/Timesheets/Controllers/ShowMy.php
```diff
@@ -64,26 +67,26 @@ public function run(): Response
 
         // Use UTC here as all data stored in the database should be UTC (start in user's timezone and convert to UTC).
         // The front end javascript is hardcode to start the week on mondays, so we use that here too.
-        $fromData = Carbon::now($_SESSION['usersettings.timezone'])->startOfWeek(CarbonInterface::MONDAY);
-        $fromData->setTimeFrom(Carbon::now($_SESSION['usersettings.timezone']))->setTimezone('UTC');
+
+        //Get start of the week in current users timezone and then switch to UTC
+        $dateTimeHelper = new DateTimeHelper();
+        $fromDate = $dateTimeHelper->userNow()->startOfWeek(CarbonInterface::MONDAY);
```
cableman, line 73: Is the conversion to UTC missing here?

app/Domain/Timesheets/Controllers/ShowMyList.php
```diff
@@ -43,18 +45,20 @@ public function run(): Response
...
         }
 
-        $dateFrom = Carbon::now($_SESSION['usersettings.timezone'])->setTimezone('UTC')->startOfMonth();
+        // Use UTC here as all data stored in the database should be UTC (start in user's timezone and convert to UTC).
+        // The front end javascript is hardcode to start the week on mondays, so we use that here too.
+
+        //Get start of the week in current users timezone and then switch to UTC
+        $dateTimeHelper = new DateTimeHelper();
+        $dateFrom = $dateTimeHelper->userNow()->startOfMonth();
+        $dateTo = $dateTimeHelper->userNow()->endOfMonth();
```
cableman, line 54: Should these be converted to UTC, so only the template layer uses users timezone?

app/Domain/Timesheets/Repositories/Timesheets.php
```diff
@@ -276,37 +277,36 @@ public function getWeeklyTimesheets(int $projectId, Carbon $fromDate, int $userI
...
 
         $call = $this->dbcall(func_get_args());
 
         $call->prepare($query);
 
-        $startOfDay = $fromDate->format('Y-m-d');
+        if (!$fromDate->isUtc()) {
+            $fromDate->setTimezone("UTC");
+        }
+        $startOfDay = $fromDate->format('Y-m-d H:i:s');
```
cableman, line 305: If possible I think we should move the "format" into a const (to help prevent typos). Could be DateTimeHelper::DatabaseTimeFormat or something like that.

app/Domain/Timesheets/Services/Timesheets.php
```diff
@@ -231,13 +301,119 @@ public function getAll(Carbon $dateFrom, Carbon $dateTo, int $projectId = -1, st
...
+                //UTC. That way we can compare the dates coming from the db
+                $timesheetGroups[$groupKey] = array(
+                    "kind" =>  $timesheet["kind"],
+                    "clientName" => $timesheet["clientName"],
+                    "name" => $timesheet["name"],
+                    "headline" => $timesheet["headline"],
+                    "ticketId" => $timesheet["ticketId"],
+                    "day1" => array(
+                        "start" =>  $fromDate,
+                        "end" => $fromDate->addHours(23)->addMinutes(59),
```
cableman, line 338: We not use `endOfDay()` here?

app/Domain/Timesheets/Templates/editTime.tpl.php
```diff
@@ -98,7 +98,7 @@
 
 </select><br />
 <label for="date"><?php echo $tpl->__('label.date')?></label> <input type="text" autocomplete="off"
-    id="datepicker" name="date" value="<?php echo format($values['date'], 'short'); ?>" size="7" />
+    id="datepicker" name="date" value="<?php echo format($values['date'], "", \Leantime\Core\Support\FromFormat::DbDate)->date(); ?>" size="7" />
```
cableman, line 101: You should be able to add an `use` statement in the top of the file and then use the short and more readable version here.

cableman: We could also use named parameters here? To remove the empty string in these calls.

app/Domain/Timesheets/Templates/showMy.tpl.php
```diff
@@ -263,69 +273,44 @@
...
 
-                        <tr class="gradeA timesheetRow">
+                            <tr class="gradeA timesheetRow">
                             <td width="14%"><?php $tpl->e($timeRow["clientName"]); ?> // <?php $tpl->e($timeRow["name"]); ?></td>
                             <td width="14%"><?php $tpl->e($timeRow["headline"]); ?></td>
                             <td width="10%"><?php echo $tpl->__($tpl->get('kind')[$timeRow['kind']]); ?></td>
 
-                            <?php foreach ($rowKeys as $i => $key) { ?>
-                                <td width="7%" class="row<?php echo $days[$i]; ?><?php if ($currentDate->isToday()) {
+                            <?php for ($i = 1; $i < 8; $i++) {
```
cableman, line 288: Maybe a comment about these "for" loop that start at 1 and runs to 8? Would is be possible to make a foreach loop over the data (foreach are consider more safe)

cableman (commented):
Some feedback :-)

app/Core/Support/DateTimeHelper.php
```diff
@@ -12,24 +15,150 @@
...
+        }
+
+        return $this->datetime;
     }
 
+    /**
+     * Parses a database date string and returns a CarbonImmutable instance.
+     *
+     * @param string $dbDate The date string in the database format to parse.
+     * @return CarbonImmutable The parsed CarbonImmutable instance.
```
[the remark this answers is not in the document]
marcelfolaron: Good call out!

app/Core/Support/DateTimeHelper.php
```diff
@@ -12,24 +15,150 @@
...
+
+    private CarbonImmutable $datetime;
     /**
      * Initializes a new instance of the class.
      *
-     * @param Language   $language   The Language object to use for translations.
-     * @param ApiRequest $apiRequest The ApiRequest object to use for accessing the API.
      */
-    public function __construct(Language $language, ApiRequest $apiRequest)
+    public function __construct($time = null, $tz = null)
```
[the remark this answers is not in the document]
marcelfolaron: I wish I could. This class inherits from CarbonImmutable which doesn't have types parameters on the contructor and I can't overload the contructor cause PHP... I added them into the phpdoc block comment though

app/Core/Support/CarbonMacros.php
```diff
@@ -0,0 +1,92 @@
...
+    public string $userDateFormat;
+
+    public string $userTimeFormat;
+
+    public string $dbTimezone;
+
+    public string $dbFormat;
+
+
+    public function __construct(
```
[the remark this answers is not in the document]
marcelfolaron: Totally agree!

app/Domain/Timesheets/Controllers/ShowMyList.php
marcelfolaron: You are correct. Thanks for catching that, those should have been UTC

app/Domain/Timesheets/Repositories/Timesheets.php
marcelfolaron: Yeah I agree. Actually, I moved to using the helper methods and not having to worry about the format at all. 
It's either dbDateTime (UTC + iso format) or userDateTime (local timezone + user format). There is one exception where I need to be able to handle 24:00 time for input type="time" which requires it.

app/Domain/Timesheets/Services/Timesheets.php
marcelfolaron: Great question: In this case we are not actually looking for the end of the day but 24hours in the future. Since this is the UTC date the date time could be starting on Sunday 18:00. So then day1 should be Sunday 18:00 to Monday 17:59 UTC. 

The reason for this range is small timezone shifts will break the weekly timesheets. For example daylight savings time or just traveling to a different timezone might add a new UTC date to the db which does not fully align with the datetime in the previous zone:

Example:
Monday: adding 4 hours while in New York
Tuesday: Traveled to San Francisco (3 hours back) and forgot to add another 2 hours to Monday. So I add a new entry. 

Because dates are stored as start of day 00:00:00 in the users timezone we will have 2 rows in the db one for Monday, 4 hours at UTC-New York and one for 2 Hours at UTC-San Fran. 

The grouping  was not taking that scenario into consideration. Now it does and just uses the users current timezone start of day transformed to UTC and then adds 24 hours. 

* I hate Timezone Management *

app/Domain/Timesheets/Templates/showMy.tpl.php
marcelfolaron: Thanks for the call out. I refactored that.

marcelfolaron:
@cableman Thanks for the feedback. I made some updates and did some additional testing + bug fixing. Still need to update current Unit tests.

marcelfolaron:
Ready for round 2

marcelfolaron:
@broskees Hey, if you have some spare time and interest, would you take a look at this as well?

app/Core/Support/FromFormat.php
marcelfolaron: They are just a helper for the format() function. It allows us to pass in a string date and then just say start of day in our template files without having to instantiate a DateTimeHelper.

Issue #2389, opened 2024-03-11
[BUG] Link to settings are shown when a plugin is not enabled.


rimi-itk:
*If you have problems with the installation please use our [community forum](https://community.leantime.io)*

**What set up are you using**
* [ ] Cloud Hosted Version
* [ ] Self Hosted

**Describe the bug**
The plugin list shows the “Settings” link for all installed plugins. Only plugins that are *enabled* should have the Settings link.
![Screen Shot 2024-03-11 at 11 35 22](https://github.com/Leantime/leantime/assets/11267554/6cb24c23-f208-4cfc-8e79-4b20e3e9adf0)

**To Reproduce**
1. Go to `/plugins/myapps`
2. Install a plugin with settings.
3. Click “Settings” 
4. Notice “Page not found”

**Expected behavior**
Only enabled plugins should have the “Settings” link.

**Leantime Version**
https://github.com/Leantime/leantime/releases/tag/v3.0.7

marcelfolaron:
Fixed in 3.1.0-beta

PR #2390, opened 2024-03-11
Added check for plugin enabled on settings link

rimi-itk:
#### Link to ticket

https://github.com/Leantime/leantime/issues/2389

#### Description

Shows "Settings" link only on *enabled* plugins.

#### Screenshot of the result

Before: 
Screen Shot 2024-03-11 at 11 35 22

After:
Screen Shot 2024-03-11 at 11 36 22

#### Checklist


marcelfolaron (approved):
Oh that is a good one!

Issue #2392, opened 2024-03-11
[BUG] Leantime\Domain\Projects\Services\Projects::getProject(): Argument #1 ($id) must be of type int, string given


rimi-itk:
*If you have problems with the installation please use our [community forum](https://community.leantime.io)*

**What set up are you using**
* [ ] Cloud Hosted Version
* [x] Self Hosted

**Describe the bug**
[A recent change](https://github.com/Leantime/leantime/commit/80cd01b434c97b2bcfbaf6df67a6a3d606e3c15c) added type-hints on function parameters and this leads to some runtime errors.

**To Reproduce**
1. Make sure that your session has no current project set (i.e. `$_SESSION[]` is empty).
2. Try to sign in
3. The error
    ```
    Illuminate\View\ViewException:
    Leantime\Domain\Projects\Services\Projects::getProject(): Argument #1 ($id) must be of type int, string given, called in /app/app/Domain/Menu/Services/Menu.php on line 125 (View: /app/app/Domain/Menu/Templates/headMenu.blade.php) (View: /app/app/Domain/Menu/Templates/headMenu.blade.php) (View: /app/app/Domain/Menu/Templates/headMenu.blade.php)
    ```
    is reported.

**Expected behavior**
I expect to be able to sign in without having a current project.

**Leantime Version**
https://github.com/Leantime/leantime/releases/tag/v3.0.7

compumatter:
Same here.  Just upgraded my development version and got a 500 error 
![image](https://github.com/Leantime/leantime/assets/121611/1b5744e2-c393-4e95-baab-96bdb9f8eecd)

Specific error in logs:
[14-Mar-2024 14:02:53 America/Los_Angeles] TypeError: Leantime\Domain\Projects\Services\Projects::getProject(): Argument #1 ($id) must be of type int, string given, called in public_html/app/Domain/Menu/Services/Menu.php on line 125 and defined in public_html/app/Domain/Projects/Services/Projects.php:101

marcelfolaron:
Fixed in 3.1.0-beta

PR #2393, opened 2024-03-11
Issue 2392

rimi-itk:
#### Link to ticket

https://github.com/Leantime/leantime/issues/2392

#### Description

Adds a `getCurrentProjectId` function in `Leantime\Domain\Projects\Services\Projects` and uses it instead of `$_SESSION['currentProject']` in select places (the ones calling [Projects::getProject()](https://github.com/Leantime/leantime/blob/6635c4d3bb5b65adff5ea877eb6273b13784b0a9/app/Domain/Projects/Services/Projects.php#L101)).

#### Checklist


marcelfolaron (approved):
Thank you!

Issue #2397, opened 2024-03-12
[BUG] JSON-RPC API does not convert datetime strings to Carbon objects.


rimi-itk:
*If you have problems with the installation please use our [community forum](https://community.leantime.io)*

**What set up are you using**
* [ ] Cloud Hosted Version
* [x] Self Hosted

**Describe the bug**
Datetimes formatted as strings, e.g. `2024-05-23`, in API params are not converted (cast) to Carbon objects.

Example:

```shell
curl http://leantime.local.computer/api/jsonrpc/ \
    --header 'x-api-key: …' \
    --header 'content-type; application/json' \
    --data @- <<'JSON'
{
 "jsonrpc": "2.0",
 "method": "leantime.rpc.timesheets.getAll",
 "id": "…",
 "params": {
  "dateFrom": "2001-01-01",
  "dateTo": "2101-01-01"
 }
}
JSON

returns

```json
{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params","data":{}},"id":"…"}%
```

and the underlying exception says `Could not cast parameter: dateFrom. See server logs for more details.`

It's worth noting that the cause of “Invalid params” is not reported in the API response making this error very hard to resolve (and apparently nothing is written in the server logs).

**To Reproduce**
Perform an API request as shown above.

**Expected behavior**
Strings should be converted to Carbon objects when needes.

**Leantime Version**
https://github.com/Leantime/leantime/releases/tag/v3.0.7

marcelfolaron:
Fixed in 3.1.1

PR #2398, opened 2024-03-12
Handle date(time) parameters in API requests

rimi-itk:
#### Link to ticket

https://github.com/Leantime/leantime/issues/2397

#### Description

* Handles converting date(time) strings in API requests to `Carbon` objects if needed.

#### Checklist

Also contains a small improvement in the API request error reporting: https://github.com/Leantime/leantime/pull/2398/commits/ac400e9357f993e6db6e137ab0429ee71075fdd2

The actual conversion from string to `Carbon` object is performed in the [`cast` function](https://github.com/ITK-Leantime/leantime/blob/ac400e9357f993e6db6e137ab0429ee71075fdd2/app/helpers.php#L254-L257). It may be more appropriate to do it in [`Cast::castTo`](https://github.com/Leantime/leantime/blob/7a0856dd66a0e5c88e4dc6a4fbaae22cbaadea80/app/Core/Support/Cast.php#L33), but that would require letting the `Cast` constructor accept a `string` type argument in addition to the already allowed types (`array|object`) and further changes to handle a string in the rest of the class.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -1,5 +1,7 @@
 <?php
 
+use Carbon\Carbon;
+use Carbon\CarbonInterface;
 use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\View\Factory;
 use Leantime\Core\Application;
@@ -7,11 +9,11 @@ use Leantime\Core\AppSettings;
 use Leantime\Core\Bootloader;
 use Leantime\Core\Language;
 use Leantime\Core\Support\Build;
+use Leantime\Core\Support\Cast;
+use Leantime\Core\Support\DateTimeHelper;
 use Leantime\Core\Support\Format;
 use Leantime\Core\Support\FromFormat;
-use Leantime\Core\Support\Cast;
 use Leantime\Core\Support\Mix;
-use Carbon\Carbon;
 
 
 if (! function_exists('app')) {
@@ -232,6 +234,11 @@ if (! function_exists('cast')) {
 
         if (enum_exists($classOrType)) {
             return Cast::castEnum($source, $classOrType);
+        }
+
+        // Convert string to date if required.
+        if (is_string($source) && is_a($classOrType, CarbonInterface::class, true)) {
+            return new DateTimeHelper($source);
         }
 
         return (new Cast($source))->castTo($classOrType, $constructParams, $mappings);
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -251,6 +251,11 @@ function cast(mixed $source, string $classOrType, array $constructParams = [], a
             return Cast::castEnum($source, $classOrType);
         }
 
+        // Convert string to date if required.
+        if (is_string($source) && Carbon::class === $classOrType) {
+            return new Carbon($source);
```
marcelfolaron, line 256: I know my PR (https://github.com/Leantime/leantime/pull/2386) is still outstanding but we should cast this to DateTimeHelper (which inherits from CarbonImmutable). That way all of the Carbon Macros are available

rimi-itk: Cast to DateTimeHelper in https://github.com/Leantime/leantime/pull/2398/commits/53f3ec745c0e01c4d232a28795e99f21f15c2c96.

rimi-itk: Wait, that does not work.

rimi-itk: I've rebased my branch to branch out from https://github.com/Leantime/leantime/tree/timesheet-Refactor to be able to build on your changes.

app/helpers.php
```diff
@@ -234,6 +236,11 @@ function cast(mixed $source, string $classOrType, array $constructParams = [], a
             return Cast::castEnum($source, $classOrType);
         }
 
+        // Convert string to date if required.
+        if (is_string($source) && is_a($classOrType, CarbonInterface::class, true)) {
```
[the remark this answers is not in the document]
marcelfolaron: Correct. For our cast() class to be recognized as an object though it needs to be in the object format within Json. 
I don't think that is the issue here though. The problem is that the cast class will attempt to set each attribute it finds but some of the attributes are readonly in the Immutable class. So we have to account for that somehow. 

I'll talk to @broskees about that. He wrote the cast class.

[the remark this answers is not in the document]
rimi-itk: I'm not an expert on this and am probably be missing something, but we're successfully using the Leantime API with https://patch-diff.githubusercontent.com/raw/Leantime/leantime/pull/2398.patch applied and successfully sending requests like 

```json
{
 "jsonrpc": "2.0",
 "method": "leantime.rpc.timesheets.getAll",
 "id": "…",
 "params": {
  "dateFrom": "first day of this month today",
  "dateTo": "first day of next month today",
  …
 }
}
```

marcelfolaron:
I took your changes and added them to: #2451

Issue #2403, opened 2024-03-13
text.full_lean_canvas_helper_content missing in txt docs


vid-pas:
**What set up are you using**
* [ ] Cloud Hosted Version
* [X] Self Hosted

**Describe the bug**
On a fresh install, I loaded `<URL>/strategy/showBoards` and a full lean canvas modal opened with the text `text.full_lean_canvas_helper_content`

**To Reproduce**
For 1st time visitors in English US as default lang:
1. Go to 'Blueprints' `<URL>/strategy/showBoards`
2. A full lean canvas modal opened with the text `text.full_lean_canvas_helper_content`

**Expected behavior**
The modal would only show if `text.full_lean_canvas_helper_content` existed in the current language, or `text.full_lean_canvas_helper_content` contents does exist and the text is displayed in english. 

For ex. Using a translator from German to English on the contents of `text.full_lean_canvas_helper_content` in app/Language/de-DE_informal.ini:
```
text.full_lean_canvas_helper_content = "The full research board will guide you through the entire process of idea validation.<br/>On this page, you'll once again address your customer segment, problem, solution (but don't worry, if you've already filled this out on the simple board, you'll see it again here!).<br/>From there, you'll switch to Clear Value Proposition (What Makes Your Idea Better Than Other Offerings?),<br/>Sales Channels (How Do Customers Get Your Solution?)<br/>Revenue Streams (How does your solution make money?)<br/>Cost Structure (How Much Is Your Solution Worth?)
Key Metrics (How Do You Measure Success?)<br/>and unfair advantage (What helps you stand out against the competition?)<br/>Answering these questions is designed to drive success, mitigate failures, and allow you to track and validate your assumptions.<br/>Lean is about eliminating waste and working to build products, stores, and solutions that your customers need."
```

**Leantime Version**
Which Leantime version are you using
Master: 
`a13b28c Merge pull request #2399 from Leantime/snyk-upgrade-5ac0841a0aa8a08a67be95308cc04a84 Marcel Folaron Tue Mar 12 10:45:24 2024

**Server**
Docker

**Additional context**
I cleared my cookies to attempt to recreate this but wasn't able to.

marcelfolaron:
Fixed in 3.1.0-beta

PR #2404, opened 2024-03-13
2403 Added en-US translation

vid-pas:
#### Link to ticket
#2403 

#### Description

Dropping in the en-US translation of `text.full_lean_canvas_helper_content` in app/Language/de-DE_informal.ini into app/Language/en-US.ini

Ideally, the fullLeanCanvas modal will display that description instead of the token `text.full_lean_canvas_helper_content` when en-US is the default language.


marcelfolaron:
Good catch. Thank you!

PR #2461, opened 2024-04-09
fix incorrect symbol in the translation for zh-CN

rqi14:
#### Description

Wrong % symbol is used, causing failure of text rendering.

#### Screenshot of the result

image

#### Checklist
Translation only. 


rqi14:
ok.. I have no idea what is going on about the acceptance test. As it is just a correction of translation, I guess it is probably not going to break anything.

PR #2535, opened 2024-06-07
1621: add assignee name to todo in project view

sinejespersen:
#### Link to ticket

https://leantime.itkdev.dk/dashboard/show#/tickets/showTicket/1621

#### Description

add assignee name to todo in project view 

#### Screenshot of the result

image

#### Checklist


marcelfolaron:
Thank you for the PR. 
I think if we add the assignee we should use the same component as on other screens like the kanban board where we show the image + user name as dropdown. I'd probably suggest to just add that to the list of fields on the right to keep the UI consistent.

PR #2536, opened 2024-06-07
Update Tickets.php to include editorId in getAllBySearchCriteria

rqi14:
In the original version, authorId is included but editorId is not included. As editor first name last name are both included, it does not make sense not to have editorId included.

#### Link to ticket

None

#### Description

In the original version, authorId is included but editorId is not included. As editor first name last name are both included, it does not make sense not to have editorId included.

#### Screenshot of the result

None

#### Checklist


marcelfolaron:
editorId is included in line 401 of that file. Or am I missing something?

marcelfolaron:
Closing this for now. Let me know if there is an issue with using editorId as a field from above

rqi14:
> editorId is included in line 401 of that file. Or am I missing something?

Hi. Thanks for pointing it out. You are absolutely right.

PR #2553, opened 2024-06-20
Fixing template for new sprint creation

pvi-github:
#### Link to ticket

No ticket spotted... i rand into the bug and just fixed it.

#### Description

When using the button "create sprint" the produced html is buggy and lead to wrong project assignation if the default choice is not made.

image

#### Screenshot of the result

There is no visible difference except the list does not show projects twice any more and the projects id are correctly associated with project names.

#### Checklist

N/A

The change made consists in removing the duplicate foreach loop on select. It was probably just a copy paste made twice.


marcelfolaron:
Thank you!

marcelfolaron (approved):

PR #2569, opened 2024-07-02
Project list avatar - html based

jeppekroghitk:
#### Link to ticket

N/A

#### Description

This PR is a proposal for an alternative and simpler way to generate project avatars.

When loading the front-page, a get-request for each project is made to get the avatar for the project list in the header. In my test case, having 48 projects, the 48 simultaneous get-requests slowed each page load by over 50%.

Since the generated avatar svg is quite simple and consists of two letters, i thought that a way to go, or a simple short-term solution were to generate these avatars using HTML. 

#### Screenshot of the result

Screenshot 2024-07-02 at 16 38 06

#### Checklist

### DIFF
--- a/app/Domain/Menu/Templates/partials/projectLink.blade.php
+++ b/app/Domain/Menu/Templates/partials/projectLink.blade.php
@@ -3,11 +3,7 @@
        data-tippy-content='{{ $project["name"] }}'
     @endif >
     <span class='projectAvatar'>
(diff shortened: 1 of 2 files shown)


sinejespersen (approved):
💨🏃🚄🚅🚗🏎️🏍️

marcelfolaron:
Hey, 
Thank you but this PR would prevent uploaded avatars from showing up as well. 
The svgs/images are all served with a cache header so on subsequent calls the browser should not load the images from the server anymore. Can you check in dev tools if the images show up as "Memory Cached". If not there may be something off in the way the images are being served.

PR #2577, opened 2024-07-10
feat: zapier integration 

kylemcshea:
#### Link to ticket

**Developer Guide**
https://docs.google.com/document/d/1HWamnw-ebKEM60mPL8sLQYJPW8inJHF2a-RSXnIxG5c/edit#heading=h.7jl8tl5k1raa

#### Description

The intent of this PR is to add the following pieces for a zapier integration:

**Triggers**
- New Todo
- Todo Updated (fields in particular: status, assignee, due date, title, description, sprint, milestone)
- New Milestone (milestones are just todos with the type “milestone”)
- Milestone updated
- New Goal
- Goal Updated (objective, metrics start, current, goal) 
- New Idea
- Idea Updated
- New Comment (could be anywhere in the system)
- New Timesheet entry
- Timesheet entry updated
- New Project
- Project details updated

**Actions**
- Create Todo
- Update Todo (not sure if we want to be specific in terms of updates, see below)
```
Update Todo Status
Update Assignee
Update Milestone
```
- Create Milestone
- Create Timesheet Entry

This was done by utilizing mostly service layer methods that with the proper parameters can be utilized.

With that being said not all service layer methods needed to make these zapier interactions possible are made.

One thing to denote - when checking for an updated artifact since Zapier uses `id` to index objects in a list, what we must do is append it with its `last_updated` dateTime string so that we can detect any differences in the `id` indices.

So we would end up modifying our returned `id` to be like `{id}-{modified_date}`

Here is where I derived this decision from:
https://community.zapier.com/how-do-i-3/polling-an-api-for-updated-records-not-just-new-ones-1165

#### Screenshot of the result

#### Checklist

New Goal will be completed in a seperate PR as the canvas -> canvas item relation makes that task not exactly straight forward


app/Domain/Ideas/Repositories/Ideas.php
```diff
@@ -626,5 +626,40 @@ public function bulkUpdateIdeaStatus($params): bool
...
             return true;
         }
+
+        /**
+         * @return array|false
+         */
+        public function getAllIdeas(): array|false
+        {
+
+            // TODO: make this not select * and identify what fields we want back.
```
kylemcshea, line 636: remove this commment

app/Domain/Ideas/Services/Ideas.php
```diff
@@ -0,0 +1,32 @@
...
+    class Ideas
+    {
+        private IdeasRepository $ideasRepository;
+
+        public function __construct(IdeasRepository $ideasRepository)
+        {
+            $this->ideasRepository = $ideasRepository;
+        }
+
+        public function pollIdeas(): array
```
kylemcshea, line 16:
```suggestion
        public function pollForNewIdeas(): array
```

app/Domain/Projects/Services/Projects.php
```diff
@@ -1559,6 +1560,19 @@ public function getAll(bool $showClosedProjects = false): array
         {
             return $this->projectRepository->getAll($showClosedProjects);
         }
-    }
 
+        /**
+         * @return array
+         */
```
kylemcshea, line 1566: unneeded

```suggestion
```

app/Domain/Tickets/Services/Tickets.php
```diff
@@ -1232,14 +1232,16 @@ public function quickAddMilestone($params): array|bool|int
          */
         public function addTicket($values)
         {
+            $userId = $values['userId'] ?? session("userdata.id");
```
kylemcshea, line 1235: consider leaving a comment here to explain that zapier doesn't have user data sessions available

app/Domain/Tickets/Services/Tickets.php
```diff
@@ -2059,6 +2061,157 @@ public function prepareTicketDates(&$values)
...
+            );
+
+            foreach ($todos as $key => $todo) {
+                $todos[$key]['id'] = $todo['id'] . '-' . $todo['date'];
+            }
+
+            return $todos;
+        }
+
+        public function createTodoUsingZapier($values): array|bool
```
kylemcshea, line 2114: add a docstring entailing what the  values object fields can be.

app/Domain/Tickets/Services/Tickets.php
```diff
@@ -2059,6 +2061,157 @@ public function prepareTicketDates(&$values)
...
+            );
+
+            try {
+                $sucessfullyUpdated = $this->updateTicket($values);
+                if ($sucessfullyUpdated) {
+                    return [
+                        "message" => "Ticket updated successfully",
+                        "status" => "success",
+                    ];
+                } else {
```
kylemcshea, line 2177: nit - else block not needed

app/Domain/Timesheets/Services/Timesheets.php
```diff
@@ -452,4 +453,59 @@ public function getBookedHourTypes(): array
     {
         return $this->timesheetsRepo->kind;
     }
+
+    public function pollTimesheets(): array|false
```
kylemcshea, line 457:
```suggestion
    public function pollForNewTimesheets(): array|false
```

app/Domain/Tickets/Services/Tickets.php
```diff
@@ -2059,6 +2061,157 @@ public function prepareTicketDates(&$values)
...
+                }
+            } catch (\Exception $e) {
+                return [
+                    "message" => "Ticket could not be updated. Reason: " . $e->getMessage(),
+                    "status" => "error",
+                ];
+            }
+        }
+
+        public function createMilestoneUsingZapier($values): array|bool
```
marcelfolaron, line 2191: Is there anything special in here or could you just use the `addTicket` method from this file?

app/Domain/Tickets/Services/Tickets.php
```diff
@@ -2059,6 +2061,157 @@ public function prepareTicketDates(&$values)
...
+                'editFrom' => $values['editFrom'] ?? "",
+                'editTo' => $values['editTo'] ?? "",
+                'dependingTicketId' => $values['dependingTicketId'] ?? "",
+                'milestoneid' => $values['milestoneid'] ?? "",
+            );
+
+            return $this->addTicket($values);
+        }
+
+        public function updateTodoUsingZapier($values): array
```
marcelfolaron, line 2143: Can you use any of the existing update methods or does that create problems?

app/Domain/Tickets/Services/Tickets.php
marcelfolaron, line 2114: Why this wrapper, it seems the addTicket method would be able to handle this?

app/Domain/Tickets/Services/Tickets.php
marcelfolaron: If we need this method can we rename it and remove Zapier from the method name. I'd like to keep the service layer flexible and functionally useful for various use cases.

app/Domain/Timesheets/Services/Timesheets.php
```diff
@@ -452,4 +453,59 @@ public function getBookedHourTypes(): array
...
+     *     @type bool?   $paid         Whether the time has been paid.
+     * }
+     *
+     * @return array {
+     *     @type bool $success Whether the timesheet entry was created successfully.
+     * }
+     *
+     * @throws InvalidArgumentException if required parameters are missing.
+     */
+    public function createTimesheetViaZapierAction($values): array
```
marcelfolaron, line 493: Please rename. Is this new method needed or can we use an existing one?

marcelfolaron (changes requested):

app/Domain/Tickets/Services/Tickets.php
kylemcshea: you are correct sessions do work ✅

app/Domain/Tickets/Services/Tickets.php
kylemcshea: yep I fully agree,

I am going to hard code the `type` to be `milestone` for the request body in zapier and utilize that `addTicket` function!

app/Domain/Timesheets/Services/Timesheets.php
kylemcshea: lets use existing `logTime` I agree

app/Domain/Tickets/Services/Tickets.php
kylemcshea: yep lets use `addTicket` I agree

app/Domain/Tickets/Services/Tickets.php
kylemcshea: yep lets use `updateTicket` I agree

PR #2620, opened 2024-08-20
Refactro goalcanvas

muhtasimhafiz:
#### Description

#### Link to ticket

### Type

[] Fix
[] Feature
[x] Cleanup 

#### Screenshot of the result


app/Domain/Canvas/Repositories/Canvas.php
```diff
@@ -282,7 +282,7 @@ public function getAllCanvas($projectId, $type = null): false|array
          * @param $canvasId
          * @return array|false
          */
-        public function getSingleCanvas($canvasId): false|array
+        public function getSingleCanvas($canvasId)
```
marcelfolaron, line 285: Why did you remove that? This is the return value definition and always good to have.

app/Domain/Goalcanvas/Controllers/BigRock.php
```diff
@@ -68,7 +68,7 @@ public function get($params): Response
                 $bigrock = array("id"=>'', "title" => "", "prpojectId" => "", "author" => '');
             }
 
-
+            dd($bigrock);
```
marcelfolaron, line 71: Please remove

app/Domain/Goalcanvas/Controllers/ShowCanvas.php
```diff
@@ -265,10 +64,41 @@ public function run()
             $this->tpl->assign('allCanvas', $allCanvas);
             $this->tpl->assign('canvasItems', $this->goalService->getCanvasItemsById($currentCanvasId));
             $this->tpl->assign('users', $this->projectService->getUsersAssignedToProject(session("currentProject")));
+        
+            return $this->tpl->display(static::CANVAS_NAME . 'canvas.showCanvas');
```
marcelfolaron, line 68: Since this is the goal canvas you can change "static::CANVAS_NAME" to 'goalcanvas.showCanvas'

app/Domain/Goalcanvas/Controllers/ShowCanvas.php
```diff
@@ -265,10 +64,41 @@ public function run()
...
+                    $result = $this->goalService->mergeCanvas($_POST['canvasid'] ?? -1, $_POST['mergeCanvasId'] ?? -1);
+                    break;
+                case 'importCanvas':
+                    $result = $this->goalService->importCanvas($_FILES['canvasfile'] ?? null);
+                    break;
+            }
+        
+            if ($result['success']) {
+                $this->tpl->setNotification($result['message'], 'success');
+                return Frontcontroller::redirect(BASE_URL . '/' . static::CANVAS_NAME . 'canvas/showCanvas/');
```
marcelfolaron, line 96: See above. Canvas name can be renamed to just be canvas

app/Domain/Goalcanvas/Repositories/Goalcanvas.php
```diff
@@ -151,13 +153,15 @@ public function getSingleCanvas($canvasId): false|array
                 ORDER BY zp_canvas.title, zp_canvas.created";
 
             $stmn = $this->db->database->prepare($sql);
-            $stmn->bindValue(':canvasId', $canvasId, PDO::PARAM_STR);
+            $stmn->bindValue(':canvasName', static::CANVAS_NAME . 'canvas', PDO::PARAM_STR);
```
marcelfolaron, line 156: Can replace this with goalcanvas

app/Domain/Goalcanvas/Services/Goalcanvas.php
```diff
@@ -3,13 +3,27 @@
...
      */
     class Goalcanvas
     {
         private GoalcanvaRepository $goalRepository;
+        private ?Projects $projectService;
+
+        protected ?Language $language;
+        protected Mailer $mailer;
+
+        protected const CANVAS_NAME = 'goal';
```
marcelfolaron, line 24: You can remove that and just replace the CANVAS_NAME instances to goal

app/Domain/Goalcanvas/Services/Goalcanvas.php
```diff
@@ -202,5 +223,488 @@ public function pollForUpdatedGoals(): array|false
...
+                'title' => $params['canvastitle'],
+                'author' => session("userdata.id"),
+                'projectId' => session("currentProject"),
+            ];
+            $currentCanvasId = $this->goalRepository->addCanvas($values);
+
+            $this->sendCanvasNotification('notification.board_created', $values['title']);
+
+            session(["current" . strtoupper(static::CANVAS_NAME) . "Canvas" => $currentCanvasId]);
+            return $this->setRedirect($result, BASE_URL . '/' . static::CANVAS_NAME . 'canvas/showCanvas/');
```
marcelfolaron, line 395: The service layer should ideally not perform any redirects. If the service is called via API the redirect would break the api response. 

I think in this case the method could just return either an error message or the id of the new canvas.

app/Domain/Goalcanvas/Services/Goalcanvas.php
```diff
@@ -202,5 +223,488 @@ public function pollForUpdatedGoals(): array|false
...
+        }
+
+        private function handleNewCanvas($params, $result): array
+        {
+            if (!isset($params['canvastitle']) || empty($params['canvastitle'])) {
+                return $this->setNotification($result, 'notification.please_enter_title', 'error');
+            }
+
+            if ($this->goalRepository->existCanvas(session("currentProject"), $params['canvastitle'])) {
+                return $this->setNotification($result, 'notification.board_exists', 'error');
```
marcelfolaron, line 382: Please move template notifications into the controller. The service should just return success or failure messages. (Maybe with failure errors/exceptions) 
So in this case you could throw a "ElementExistsException()" (just create a new exception type into core/exceptions. The controller can then decide what to do with that exception. (aka show a notification)

app/Domain/Goalcanvas/Services/Goalcanvas.php
```diff
@@ -202,5 +223,488 @@ public function pollForUpdatedGoals(): array|false
...
+                $result = $this->handleImportCanvas($params, $result);
+            }
+
+            return $result;
+        }
+
+        private function handleNewCanvas($params, $result): array
+        {
+            if (!isset($params['canvastitle']) || empty($params['canvastitle'])) {
+                return $this->setNotification($result, 'notification.please_enter_title', 'error');
```
marcelfolaron, line 378: This should throw a MissingParameterException()

app/Domain/Ideas/Templates/ideaDialog.blade.php
```diff
@@ -0,0 +1,163 @@
+@extends($layout)
+
+@section('content')
+    @php
+        $id = '';
```
marcelfolaron, line 5: This should be moved into the controller

marcelfolaron (changes requested):
Overall really good update.
I am not fully done with the review but there are 2 big items we should address:

1. The service layer should not create template notifications. Since the service is also our API those notifications would not be available to API endpoints. Instead the services should throw exceptions if something goes wrong/is missing. Or it should return true or return the id of a newly created item etc. 

2. Similar to the notifications the service layers should not do any redirects. Same reason with regards to the API. API endpoints would get redirected to a UI endpoint and fail.

app/Domain/Goalcanvas/Services/Goalcanvas.php
marcelfolaron: `throw new MissingParameterException()`

app/Domain/Goalcanvas/Controllers/Dashboard.php
```diff
@@ -48,283 +49,41 @@ public function init(
...
-            $this->tpl->assign('users', $this->projectService->getUsersAssignedToProject(session("currentProject")));
-
-            if (!isset($_GET['raw'])) {
-                return $this->tpl->display(static::CANVAS_NAME . 'canvas.dashboard');
-            }
-        }
+         public function get($params): Response
+         {
+
+             $result = $this->goalService->handleDashboardGetRequest($params);
```
marcelfolaron, line 56:
```
try {
$result = $this->goalService->handleDashboardGetRequest($params);
}catch(MissingParameterException $e) {
  $this->tpl->setNotification
}catch() {
}finally {
$result = [];
}
```

app/Domain/Goalcanvas/Controllers/BigRock.php
marcelfolaron: Why did you resolve the comment but not remove the line?

muhtasimhafiz: i resolved it before pushing the code, I resolved it locally

app/Domain/Goalcanvas/Controllers/BigRock.php
```diff
@@ -66,7 +66,6 @@ public function get($params): Response
                 $bigrock = array("id"=>'', "title" => "", "prpojectId" => "", "author" => '');
             }
 
-
```
marcelfolaron, line 69: Few items missing from this file:
- No repo calls (they should all be service calls. If there is no service, just create a wrapper service method
- CANVAS_NAME is still being used

app/Domain/Goalcanvas/Controllers/EditCanvasComment.php
```diff
@@ -101,6 +101,7 @@ public function get($params)
             }
 
             $this->tpl->assign('comments', $comments);
+            $this->tpl->assign('id', $canvasItem['id']);
```
marcelfolaron, line 104: Few items missing from this file:
- No repo calls (they should all be service calls. If there is no service, just create a wrapper service method
- CANVAS_NAME is still being used

app/Domain/Goalcanvas/Controllers/EditCanvasItem.php
```diff
@@ -201,8 +201,8 @@ public function post($params): Response
                             'parent' => $params['parent'] ?? null,
                             "id" => $params['itemId'],
                             'kpi' => $params['kpi'] ?? '',
-                            'startDate' => format(value: $params['startDate'], fromFormat: FromFormat::UserDateStartOfDay)->isoDateTime(),
-                            'endDate' => format(value: $params['endDate'], fromFormat: FromFormat::UserDateEndOfDay)->isoDateTime(),
+                            'startDate' => !empty($params['startDate'])?format(value: $params['startDate'], fromFormat: FromFormat::UserDateStartOfDay)->isoDateTime():'',
```
marcelfolaron, line 204: Few items missing from this file:
- No repo calls (they should all be service calls. If there is no service, just create a wrapper service method
- CANVAS_NAME is still being used

marcelfolaron (changes requested):
Please review all the controllers and make sure that 

- No repo calls (they should all be service calls. If there is no service, just create a wrapper service method
- CANVAS_NAME is still being used and should be replaced. 
- Remove commented out code snippets
- Add doc comments to service

PR #2640, opened 2024-08-29
Fix LDAP AD login without domain (using only user ID).

hardkeo:
#### Description

The  Problem:
Even if you set a domain for the LEAN_LDAP_LDAP_DOMAIN environment variable, you will not be able to login using only the user ID, you can only login using the full form, user@domain.

**Example login entry:**

*Before changes:*
* "user@domain", result Login
* "user", result ```Error 500```

*After changes:*
* "user@domain", result OK
* "user", result OK

**Type**
[X] Fix
[] Feature
[] Cleanup

### DIFF
--- a/app/Domain/Ldap/Services/Ldap.php
+++ b/app/Domain/Ldap/Services/Ldap.php
@@ -141,14 +141,17 @@ class Ldap
             if ($this->directoryType == 'AD') {
                 $usernameDN = $username;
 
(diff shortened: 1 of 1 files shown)


marcelfolaron:
This is such a great simple way to solve this. Thank you!

marcelfolaron (approved):

PR #2646, opened 2024-09-03
Components

muhtasimhafiz:
#### Description

#### Link to ticket

### Type

[] Fix
[x] Feature
[] Cleanup 

#### Screenshot of the result


app/Views/Templates/components/actions/dropdown.blade.php
```diff
@@ -0,0 +1,36 @@
+@props([
+    'buttonClass' => 'btn btn-primary',  // Default button class
```
marcelfolaron, line 2: Please make sure to use attributes from our attribute list. The buttonClass consists of multiple attributes: 

variant [regular, header, card]
content-role [primary, secondary, tertiary]
state [info, danger, warning etc]

In this case the dropdowns would also have position to align them with the button

app/Views/Templates/components/actions/dropdown.blade.php
```diff
@@ -0,0 +1,36 @@
...
+    'menuClass' => 'dropdown-content menu bg-base-100 rounded-box z-50 w-52 p-2 shadow',  // Default menu class
```
marcelfolaron, line 3: That should not be a thing people can pass in. So just add it to the menu list

app/Views/Templates/components/actions/dropdown.blade.php
```diff
@@ -0,0 +1,36 @@
...
+])
+
+<div {{ $attributes->merge(['class' => 'dropdown']) }}>
+    <!-- Dropdown Button -->
+    <label tabindex="0" class="{{ $buttonClass }}">
+        {{ $button ?? 'Dropdown' }}
+    </label>
+
+    <!-- Dropdown Menu -->
+    <ul tabindex="0" class="{{ $menuClass }}">
```
marcelfolaron, line 13: Given that dropdowns can also be cards. This should be flexible based on the variant.

app/Views/Templates/components/actions/dropdown.blade.php
```diff
@@ -0,0 +1,36 @@
...
+        {{ $menu }}
```
marcelfolaron, line 14: Can you create a separate commponent called dropdown-item so that the links + link tags are in there?

app/Views/Templates/components/elements/accordian.blade.php
```diff
@@ -0,0 +1,14 @@
+@props([
+    'title' => 'Accordion Title',
```
marcelfolaron, line 2: Similar to my feedback above the attributes should be aligned with the ones we have on our sheet

app/Views/Templates/components/elements/avatar.blade.php
```diff
@@ -0,0 +1,37 @@
+@props([
+    'src' => '',                 // URL for the avatar image
+    'alt' => 'Avatar',           // Alt text for the avatar image
+    'size' => 'w-12 h-12',       // Size of the avatar (default is medium size)
```
marcelfolaron, line 4: Size should not just be overwritten. Instead we should pass in semantic sizing like small, medium large and the component decides which classses to choose

marcelfolaron (changes requested):

app/Views/Templates/components/actions/dropdown.blade.php
marcelfolaron: By allowing devs to rewrite the class entirely you force them to write css. So we would need to keep the head dropdown css when in actuality it is just a modifier that we should add. In addition to that you wouldn't add any of the daisyui classes and so we'd have none of the standard dropdown functionality.

app/Views/Templates/components/actions/dropdown.blade.php
muhtasimhafiz: i removed these 2 props

app/Views/Templates/components/actions/dropdown.blade.php
muhtasimhafiz: i have questions about variant 'head'

muhtasimhafiz: i get card is supposed to be
for ul with extra classes
do I allow those classes to be passed through the prop like head classes?

app/Views/Templates/components/actions/dropdown.blade.php
muhtasimhafiz: i created dropdown-list-item.blade file

app/Views/Templates/components/actions/dropdown.blade.php
```diff
@@ -0,0 +1,60 @@
...
+    'card-label' => 'Card Title!', // Text for the card title
+    'caption' => ''
+])
+
+@php
+    // Determine the button class based on the content role
+    $buttonClass = match($contentRole) {
+        'secondary' => 'btn btn-secondary',
+        'tertiary' => 'btn btn-accent',
+        default => 'btn btn-primary',
```
marcelfolaron, line 15: default in this case and based on the dasiyUI page should probably just be "btn" 
then we can have primary, secondary etc.

Lets go and allow the full list of options:
primary, secondary, accent, ghost(=tertiary), link

app/Views/Templates/components/actions/dropdown.blade.php
```diff
@@ -0,0 +1,60 @@
...
+    };
+
+    // Determine the menu class based on the variant
+    $menuClass = match($variant) {
+        'card' => 'card card-compact bg-primary text-primary-content w-64 p-2 shadow', // Card variant class
+        default => 'dropdown-content bg-base-100 rounded-box p-2 shadow w-52 z-50', // Default to regular menu
+    };
+
+    // Determine the dropdown position class
+    $positionClass = match($position) {
```
marcelfolaron, line 25: seems the dropdown definition is a little weird in that it has 
top,left,buttom,right 
and then an additional modifier "end" (or beginning as default, but that doesn't have a class) 
So 2 options:

Either we add the whole list:
let's add another attribute for that "align" with the options start and end (start doesn't do anything here and "end" will add "dropdown-end"

app/Views/Templates/components/actions/dropdown-list-item.blade.php
marcelfolaron, line 1: ORganization wise can you create a folder called "dropdown" inside of actions and then call "dropdown.blade.php" index.blade.php and dropdown-list-item.blade.php item.blade.php
Blade can parse the folder as part of the component name so that this will be dropdown and dropdown.item

marcelfolaron (changes requested):

app/Views/Templates/components/actions/dropdown/item.blade.php
```diff
@@ -0,0 +1,14 @@
+@props([
+    'label-text' => '', // Text or HTML content for the dropdown item
+    'variant' => 'link', // Variant of the dropdown item: 'link' or 'border'
+])
+
+@if ($variant === 'link')
+    <li>
+        <a {{ $attributes }}>
+            {!! $labelText !!}
```
marcelfolaron, line 9: Instead of label text let's just use $slot {{ $slot }}

app/Views/Templates/components/actions/dropdown/index.blade.php
```diff
@@ -0,0 +1,57 @@
+@props([
+    'variant' => 'regular', // Dropdown variant: regular or card
+    'contentRole' => 'primary', // Content role: primary, secondary, or tertiary
+    'position' => 'bottom', // Dropdown position: top, left, bottom, right
+    'align' => 'start', // Dropdown alignment: start or end
+    'label-text' => 'Dropdown', // Text for the dropdown button
```
marcelfolaron, line 6: Did you check if it should be labelText and cardLabel vs label-text and card-label?

marcelfolaron (changes requested):

marcelfolaron (approved):

PR #2647, opened 2024-09-03
refactor: added text-input global component and updated domain for Ca…

ctanveer:
…lendar events

#### Description

Reusable component for text-input

#### Link to ticket

### Type

[] Fix
[x] Feature
[] Cleanup 

#### Screenshot of the result

### DIFF
--- a/app/Views/Templates/components/forms/text-input.blade.php
+++ b/app/Views/Templates/components/forms/text-input.blade.php
@@ -1,0 +1,48 @@
+@props([
+    'inputType' => 'text',
+    'id' => '',
+    'name' => '',
+    'placeholder' => '',
+    'label' => '',
+    'size' => '',
+    'inputState' => '',
+    'value' => '',
+    'leadingVisual' => '',
+    'trailingVisual' => '',
+])
+
+@php
+    $sizeClass = $size ? 'input-'.$size : '';
+    $stateClass = $inputState ? 'input-'.$inputState : '';
+@endphp
+
+<div class='par relative w-full max-w-xs'>
+    @if($label)
+        <label for="{{ $id }}">
+            <span class="label-text">{{ $label }}</span>
+        </label>
+    @endif
+
+    <div class="relative">
+        @if($leadingVisual)
+            <span class="absolute inset-y-0 left-0 pl-3 flex items-center pointer-events-none">
+                <i class="{{ $leadingVisual }}"></i>
+            </span>
+        @endif
+
+        <input 
+            type="{{ $inputType }}" 
+            id="{{ $id }}" 
+            name="{{ $name }}" 
+            placeholder="{{ $placeholder }}" 
+            value="{{ $value }}" 
+            {{$attributes->merge(['class' => 'input input-bordered '.$sizeClass.' '.$stateClass.' '.'w-full max-w-xs '.($leadingVisual ? 'pl-10' : '').($trailingVisual ? 'pr-10' : '')])}}
+        />
+
+        @if($trailingVisual)
+            <span class="absolute inset-y-0 right-0 pr-3 flex items-center pointer-events-none">
+                <i class="{{ $trailingVisual }}"></i>
+            </span>
+        @endif
+    </div>
+</div>
--- a/app/Domain/Calendar/Templates/addEvent.blade.php
+++ b/app/Domain/Calendar/Templates/addEvent.blade.php
@@ -1,3 +1,7 @@
+@extends($layout)
+
+@section('content')
+
 <?php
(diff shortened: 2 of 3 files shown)


app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -0,0 +1,48 @@
...
+            </span>
+        @endif
+
+        <input 
+            type="{{ $inputType }}" 
+            id="{{ $id }}" 
+            name="{{ $name }}" 
+            placeholder="{{ $placeholder }}" 
+            value="{{ $value }}" 
+            {{$attributes->merge(['class' => 'input input-bordered '.$sizeClass.' '.$stateClass.' '.'w-full max-w-xs '.($leadingVisual ? 'pl-10' : '').($trailingVisual ? 'pr-10' : '')])}}
```
marcelfolaron, line 39: Since you are using the `$attribute` variable you don't actually need to define the other (standard) attributes (type, id, name, placeholder, value) They will all be part of the $attribute array and displayed alongside the class. (The merge function is really just a way to specifically merge 2 attributes as you do with the class).

app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -0,0 +1,48 @@
...
+        />
+
+        @if($trailingVisual)
+            <span class="absolute inset-y-0 right-0 pr-3 flex items-center pointer-events-none">
+                <i class="{{ $trailingVisual }}"></i>
+            </span>
+        @endif
+    </div>
```
marcelfolaron, line 47: In additon to the label can you also add the bottom left text and top right text as outlined here: https://daisyui.com/components/input/#with-form-control-and-labels.

Text on the bottom should be an attribute called "caption" top right probebly just label-right. 
The caption text should have a state class as well so that if we use it for validation messages and we send in "danger" as the state it becomes "text-danger"

app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -0,0 +1,48 @@
+@props([
+    'inputType' => 'text',
+    'id' => '',
+    'name' => '',
+    'placeholder' => '',
+    'label' => '',
+    'size' => '',
+    'inputState' => '',
```
marcelfolaron, line 8: can you just call it "state"

app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -0,0 +1,48 @@
+@props([
+    'inputType' => 'text',
+    'id' => '',
+    'name' => '',
+    'placeholder' => '',
+    'label' => '',
```
marcelfolaron, line 6: Lets call label labelText to avoid any collision with html attribute names. See our spreadsheet for details.

app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -0,0 +1,48 @@
...
+    @if($label)
+        <label for="{{ $id }}">
+            <span class="label-text">{{ $label }}</span>
+        </label>
+    @endif
+
+    <div class="relative">
+        @if($leadingVisual)
+            <span class="absolute inset-y-0 left-0 pl-3 flex items-center pointer-events-none">
+                <i class="{{ $leadingVisual }}"></i>
```
marcelfolaron, line 29: I like it but more than likely we will be using a different icon library moving forward (blade ui kit). Which uses blade component styles or svg directives. I think we'd want to be flexible here and just display the "leadingvisual" without being forced to be a class on an i element

app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -0,0 +1,48 @@
...
+            id="{{ $id }}" 
+            name="{{ $name }}" 
+            placeholder="{{ $placeholder }}" 
+            value="{{ $value }}" 
+            {{$attributes->merge(['class' => 'input input-bordered '.$sizeClass.' '.$stateClass.' '.'w-full max-w-xs '.($leadingVisual ? 'pl-10' : '').($trailingVisual ? 'pr-10' : '')])}}
+        />
+
+        @if($trailingVisual)
+            <span class="absolute inset-y-0 right-0 pr-3 flex items-center pointer-events-none">
+                <i class="{{ $trailingVisual }}"></i>
```
marcelfolaron, line 44: Same feedback as above on the trailing visul

marcelfolaron (changes requested):
Overall really good! 
Few items on the component but nothing too crazy.

marcelfolaron (approved):

PR #2708, opened 2024-09-24
UI login screen fix

ctanveer:
…component

### Description

Login screen ui fixed. Made minor changes to button and text-input component 

### Link to ticket

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result


app/Domain/Auth/Templates/login.blade.php
```diff
@@ -56,7 +58,7 @@
         <?php $tpl->dispatchTplEvent('beforeSubmitButton'); ?>
 
         <div class="">
-            <x-global::forms.button type="submit" content-role="primary" name="login">
+            <x-global::forms.button type="submit" content-role="primary" name="login" cornerShape="full" :takeFullWidth="true">
                 {{ $tpl->language->__('buttons.login') }}
```
marcelfolaron, line 62: for the full width attribute can we just use "scale" and add full-width as an option

app/Domain/Auth/Templates/login.blade.php
```diff
@@ -56,7 +58,7 @@
         <?php $tpl->dispatchTplEvent('beforeSubmitButton'); ?>
 
         <div class="">
-            <x-global::forms.button type="submit" content-role="primary" name="login">
+            <x-global::forms.button type="submit" content-role="primary" name="login" cornerShape="full" :takeFullWidth="true">
```
marcelfolaron, line 61: I assume the cornershape about the rounded corners? Let's take this out and keep the buttons consistent.

app/Domain/Auth/Templates/login.blade.php
```diff
@@ -72,10 +74,11 @@
     <?php if ($tpl->get('oidcEnabled')) { ?>
         <?php $tpl->dispatchTplEvent('beforeOidcButton'); ?>
         <div class="">
-            <center class="uppercase"><?php echo $tpl->language->__("label.or"); ?></center><br />
-            <a href="{{ BASE_URL }}/oidc/login" style="width:100%;" class="btn btn-primary">
-            <?php echo $tpl->language->__("buttons.oidclogin"); ?>
-            </a>
+            <br /><center class="uppercase"><?php echo $tpl->language->__("label.or"); ?></center><br />
+            
+            <x-global::forms.button onclick="window.location.href = '{{ BASE_URL }}/oidc/login'" content-role="primary" name="oidclogin" cornerShape="full" :takeFullWidth="true">
```
marcelfolaron, line 79: same here, lets take out corner-shape and make takeFullWidth a value for scale

marcelfolaron (changes requested):
Overall good but we want to avoid adding special rules for styling like border radius. The buttons should all be consistent in their design. In this case we are going to keep the new button design.

app/Domain/Auth/Templates/login.blade.php
marcelfolaron: class="w-full"

app/Domain/Auth/Templates/login.blade.php
ctanveer: removed cornerShape and takeFullWidth attribute. using class="w-full"

app/Domain/Auth/Templates/login.blade.php
ctanveer: done

app/Domain/Auth/Templates/login.blade.php
ctanveer: done

app/Views/Templates/components/forms/text-input.blade.php
```diff
@@ -38,6 +37,10 @@
         $width = 'max-w-xs';
         $sizeClass = $size ? 'input-'.$size : '';
 
+    } elseif ($variant === 'fullWidth') {
+        $style= 'input-bordered  input-sm';
+        $width = 'w-full';
+        $sizeClass = $scale ? 'input-'.$scale : '';
     } else {
```
ctanveer, line 44: for text-input component, I added a new variant called "fullWidth" as using the default did not take the full width of the input as it had both "w-full" and "max-w-xs". Would this be ok?

marcelfolaron (approved):

PR #2729, opened 2024-10-08
update canvas dialog input fields + idea save fix

muhtasimhafiz:
### Description

### Link to ticket

### Type

- [ ] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result

### DIFF
--- a/app/Domain/Projects/Services/Projects.php
+++ b/app/Domain/Projects/Services/Projects.php
@@ -244,8 +244,7 @@ class Projects
         {
 
             //Filter notifications
-            $eventCore = new EventCore();
-            $notification = $eventCore->dispatchFilter('notificationFilter', $notification);
+            $notification = self::dispatchFilter('notificationFilter', $notification);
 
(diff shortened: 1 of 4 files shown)


app/Domain/Projects/Services/Projects.php
```diff
@@ -244,7 +244,8 @@ public function notifyProjectUsers(Notification $notification): void
         {
 
             //Filter notifications
-            $notification = EventCore::dispatchFilter('notificationFilter', $notification);
+            $eventCore = new EventCore();
```
marcelfolaron, line 247: Couple things on this one:
1. The event dispatcher is a trait that can be added to a class. 
2. Objects should always be instantiated via the container calls: app()->make(EventCore::class). 
But in this case it should use the Event trait.

marcelfolaron (changes requested):

marcelfolaron (approved):

PR #2804, opened 2024-11-20
Client module refactor

ctanveer:
### Description

Refactor of Client Module

### Link to ticket

### Type

- [ ] Fix
- [ ] Feature
- [ ] Cleanup


app/Domain/Clients/Controllers/ShowClient.php
```diff
@@ -52,38 +54,95 @@ public function init(
...
-                'name' => $row['name'],
-                'street' => $row['street'],
-                'zip' => $row['zip'],
-                'city' => $row['city'],
-                'state' => $row['state'],
-                'country' => $row['country'],
-                'phone' => $row['phone'],
-                'internet' => $row['internet'],
-                'email' => $row['email'],
+            $clientValues = (object) [
```
marcelfolaron, line 135: The clientService should already return an object/model. This mapping should not be needed anymore.

app/Domain/Clients/Controllers/NewClient.php
```diff
@@ -10,35 +10,61 @@
...
+        /**
+         * get - display template and provide empty values object
+         */
+        public function get(): Response
         {
+            Auth::authOrRedirect([Roles::$owner, Roles::$admin], true);
+
+            //Only admins
+            if (Auth::userIsAtLeast(Roles::$admin)) {
+                $values = (object) [
```
marcelfolaron, line 38: Instead of creating the array/object manually you should use the newly create client model to prepare an empty object. `$client = app()->make(ClientModel::class);`

On reason to use models is so that we don't have to worry about array definitions. For example, if I want to add a new field I shouldn't have to look for arrays in controllers. I just add it to the model and it will be available everywhere

app/Domain/Clients/Controllers/NewClient.php
```diff
@@ -49,23 +75,23 @@ public function run()
...
-                        'zip' => ($_POST['zip']),
-                        'city' => ($_POST['city']),
-                        'state' => ($_POST['state']),
-                        'country' => ($_POST['country']),
-                        'phone' => ($_POST['phone']),
-                        'internet' => ($_POST['internet']),
-                        'email' => ($_POST['email']),
+                
+                if (isset($params['save']) === true) {
+                    $values = (object) [
```
marcelfolaron, line 80: Same as above. Please create a model and then map to the model directly. We don't do that everywhere but you can take at the tickets model. In there we use the constructor to map arrays to the model.

app/Domain/Clients/Controllers/ShowClient.php
```diff
@@ -52,38 +54,95 @@ public function init(
...
+
+            $row = $this->clientService->get($id);
+
+            if ($row === false) {
+                $this->tpl->display('errors.error404');
+
+                return $this->tpl->display('errors.error403');
+            }
+
+            $clientValues = (object) [
```
marcelfolaron, line 76: Not needed since the service returns an object already.

app/Domain/Clients/Controllers/ShowClient.php
```diff
@@ -52,38 +54,95 @@ public function init(
...
+                'state' => $row->state,
+                'country' => $row->country,
+                'phone' => $row->phone,
+                'internet' => $row->internet,
+                'email' => $row->email,
+            ];
+
+
+            if (empty($row) === false && Auth::userIsAtLeast(Roles::$admin)) {
+                $file = app()->make(FileRepository::class);
```
marcelfolaron, line 91: Can you pass this in via the constructor?

app/Domain/Clients/Controllers/ShowClient.php
```diff
@@ -52,38 +54,95 @@ public function init(
...
+                $project = app()->make(ProjectRepository::class);
```
marcelfolaron, line 92: We are already passing in the projectService through the constructor, can you use that instead of the repo?

app/Domain/Clients/Controllers/ShowClient.php
```diff
@@ -52,38 +54,95 @@ public function init(
...
+
+                if (session('userdata.role') == 'admin') {
+                    $this->tpl->assign('admin', true);
+                }
+
+                $this->tpl->assign('userClients', $this->clientService->getUserClients($id));
+                $this->tpl->assign('comments', $this->commentService->getComments('client', $id));
+                $this->tpl->assign('imgExtensions', ['jpg', 'jpeg', 'png', 'gif', 'psd', 'bmp', 'tif', 'thm', 'yuv']);
+                $this->tpl->assign('client', $clientValues);
+                $this->tpl->assign('users', app()->make(UserRepository::class));
```
marcelfolaron, line 102: Let's pass in the user service here and make the call in the controller. Haven't seen the template yet but I assume there is a repo method call in the template which sould probably not be in there.

app/Domain/Clients/Controllers/ShowClient.php
```diff
@@ -118,37 +177,37 @@ public function run()
...
-                        'street' => $_POST['street'],
-                        'zip' => $_POST['zip'],
-                        'city' => $_POST['city'],
-                        'state' => $_POST['state'],
-                        'country' => $_POST['country'],
-                        'phone' => $_POST['phone'],
-                        'internet' => $_POST['internet'],
-                        'email' => $_POST['email'],
+                if (isset($params['save']) === true) {
+                    $clientValues = (object) [
```
marcelfolaron, line 188: Please map using the object. In fact since you already received the client object from the db above you can just use that and override the values with your post values., better yet would be to just do that in the service so that we can do validation in there.

app/Domain/Clients/Models/Clients.php
```diff
@@ -0,0 +1,32 @@
...
+        
+        public $phone;
+        
+        public $internet;
+        
+        public $email;
+        
+        public $numberOfProjects;
+        
+        public function __construct() {}
```
marcelfolaron, line 29: You could refactor the constructor to take an array and then map the array to your client model. See the ticket model for example

app/Domain/Clients/Repositories/Clients.php
```diff
@@ -65,13 +66,15 @@ public function getClient($id): array|false
...
             $stmn->execute();
+            $stmn->setFetchMode(PDO::FETCH_CLASS, ClientsModel::class);
             $row = $stmn->fetch();
+            
             $stmn->closeCursor();
 
-            if ($row !== false && count($row) > 0) {
-                $this->name = $row['name'];
+            if ($row !== false) {
+                $this->name = $row->name;
```
marcelfolaron, line 75: The following 2 lines don't really do anything? You can just return $row.

app/Domain/Clients/Repositories/Clients.php
```diff
@@ -136,8 +140,9 @@ public function isClient($values): bool
 			name = :name AND street = :street LIMIT 1';
 
             $stmn = $this->db->database->prepare($sql);
-            $stmn->bindValue(':name', $values['name'], PDO::PARAM_STR);
-            $stmn->bindValue(':street', $values['street'], PDO::PARAM_STR);
+
+            $stmn->bindValue(':name', $values->name, PDO::PARAM_STR);
+            $stmn->bindValue(':street', $values->street, PDO::PARAM_STR);
```
marcelfolaron, line 145: I know this was there before but this is actually weird. Why are we getting client by street? I think isClient() using the name only is fine. Can you check if we are actually using this method anywhere? If not, let's just remove it.

app/Domain/Clients/Repositories/Clients.php
```diff
@@ -179,7 +185,7 @@ public function getClientsUsers($clientId): false|array
         /**
          * addClient - add a client and postback test
          */
-        public function addClient(array $values): false|string
+        public function addClient(object $values): false|string
```
marcelfolaron, line 188: lastInsertId from below should return an int instead of string. Please change from false|string to false|int

app/Domain/Clients/Templates/delClient.blade.php
```diff
@@ -12,7 +12,7 @@
     <div class="pageicon"><span class="fa fa-address-book"></span></div>
     <div class="pagetitle">
         <h5>{{ __("label.administration") }}</h5>
-        <h1><?php echo sprintf($tpl->__('headline.delete_client'), $client['name']); ?></h1>
+        <h1><?php echo sprintf($tpl->__('headline.delete_client'), $client->name); ?></h1>
```
marcelfolaron, line 15: Can you replace this with a blade syntax? `{!! sprintf( __('headline.delete_client'), $tpl->escape($client->name)) !!}`

app/Domain/Clients/Templates/delClient.blade.php
```diff
@@ -22,8 +22,8 @@
     <div class="maincontentinner">
 
         @displayNotification()
-
-        <h4 class="widget widgettitle">{{ __("subtitles.delete") }}</h4>
+        {{-- {{ __("subtitles.delete") }} --}}
+        <h4 class="widget widgettitle"><i class="fa fa-trash"></i> Delete</h4>
```
marcelfolaron, line 26: Why was this changed? This hard codes the english string rather than using the translation. If it's the html being escaped please use `{!! __("subtitles.delete") !!}`

app/Domain/Clients/Templates/delClient.blade.php
```diff
@@ -34,8 +34,7 @@
                 <x-global::forms.button type="submit" name="del" class="button">
                     {{ __('buttons.yes_delete') }}
                 </x-global::forms.button>
-
-                <x-global::forms.button tag="a" href="/clients/showClient/{{ $client['id'] }}" class="btn btn-primary">
+                <x-global::forms.button tag="a" href="/clients/showClient/{{ $client->id }}" class="btn btn-primary">
```
marcelfolaron, line 37: We should not need the class definitions here since they are already part of the component.

marcelfolaron: While we're at it, can you change the "class="button" from above as well. 
Then lastly the secondary button should have content-role="tertiary"

app/Domain/Clients/Templates/newClient.blade.php
```diff
@@ -38,7 +38,7 @@
                                     inputType="text"
                                     name="name"
                                     id="name"
-                                    value="{{ $tpl->escape($values['name']) }}"
+                                    value="{{ $tpl->escape($values->name) }}"
                                     labelText="{{ __('label.name') }}"
                                     class="form-control"
```
marcelfolaron, line 43: Let's take out the form-control class from all of these

app/Domain/Clients/Templates/newClient.blade.php
```diff
@@ -38,7 +38,7 @@
                                     inputType="text"
                                     name="name"
                                     id="name"
-                                    value="{{ $tpl->escape($values['name']) }}"
+                                    value="{{ $tpl->escape($values->name) }}"
```
marcelfolaron, line 41: {{ }} is already escaping so you can just use the value directly {{ $values->name }}

app/Domain/Clients/Templates/newClient.blade.php
```diff
@@ -49,7 +49,7 @@ class="form-control"
                                     inputType="text"
                                     name="email"
                                     id="email"
-                                    value="{{ $tpl->escape($values['email']) }}"
+                                    value="{{ $tpl->escape($values->email) }}"
```
marcelfolaron, line 52: See above, remove escape

app/Domain/Clients/Templates/newClient.blade.php
```diff
@@ -60,7 +60,7 @@ class="form-control"
                                     inputType="text"
                                     name="internet"
                                     id="internet"
-                                    value="{{ $tpl->escape($values['internet']) }}"
+                                    value="{{ $tpl->escape($values->internet) }}"
```
marcelfolaron, line 63: See above, remove escape

app/Domain/Clients/Templates/newClient.blade.php
```diff
@@ -135,15 +135,24 @@ class="form-control"
...
 
                             <div class="form-group">
-                                <div class="span4 control-label">
+                                <div class="mt-4 span4 control-label">
                                     <x-global::forms.button
                                         type="submit"
                                         name="save"
+                                        value="true"
                                         id="save"
                                         class="btn btn-primary"
```
marcelfolaron, line 144: Please remove the btn classes from here

app/Domain/Clients/Templates/newClient.blade.php
```diff
@@ -135,15 +135,24 @@ class="form-control"
...
                                     >
                                         {{ __('buttons.save') }}
                                     </x-global::forms.button>
+
+                                    <x-global::forms.button 
+                                        tag="a" 
+                                        href="/clients/showAll" 
+                                        class="btn btn-secondary"
```
marcelfolaron, line 152: Remove btn class and set content-role

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -13,7 +13,7 @@
     <div class="pageicon"><span class="fa fa-address-book"></span></div>
     <div class="pagetitle">
         <h5>{{ __("label.administration") }}</h5>
-        <h1><?php $tpl->e($values['name']); ?></h1>
+        <h1><?php $tpl->e($values->name); ?></h1>
```
marcelfolaron, line 16: Please use blade syntax {{ $values->name }}

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -45,7 +45,7 @@
                                     inputType="text" 
                                     name="id" 
                                     id="id" 
-                                    value="{{ $tpl->escape($values['id']) }}" 
+                                    value="{{ $tpl->escape($values->id) }}"
```
marcelfolaron, line 48: remove escape

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -136,14 +136,14 @@
...
-                                    value="{{ $tpl->escape($values['phone']) }}" 
+                                    value="{{ $tpl->escape($values->phone) }}" 
                                     labelText="{{ __('label.phone') }}" 
                                 />
                             </div>
                         </div>
                                                 <div class="col-md-6">
                             <h4 class="widgettitle title-light"><span class="fa fa-users"></span> {{ __("subtitles.users_assigned_to_this_client") }}</h4>
-                            <a href="#/users/newUser?preSelectedClient=<?=$values['id'] ?>" class="btn btn-primary"><i class='fa fa-plus'></i> <?=$tpl->__('buttons.add_user') ?> </a>
+                            <a href="#/users/newUser?preSelectedClient=<?=$values->id ?>" class="btn btn-primary"><i class='fa fa-plus'></i> <?=$tpl->__('buttons.add_user') ?> </a>
```
marcelfolaron, line 146: Please use blade syntax for the id and the add user strings

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -178,14 +178,22 @@
...
-                            <x-global::forms.button type="submit" name="save" id="save">
+                            <x-global::forms.button type="submit" name="save" value="true" id="save">
                                 {{ __('buttons.save') }}
-                            </x-global::forms.button>                            
+                            </x-global::forms.button>
+                            
+                            <x-global::forms.button 
+                                tag="a" 
+                                href="/clients/showAll" 
+                                class="btn btn-secondary"
```
marcelfolaron, line 190: Remove btn classes

marcelfolaron: And set content-role

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -178,14 +178,22 @@
...
+                            >
+                                {{ __('buttons.back') }}
+                            </x-global::forms.button>
                         </div>
                         <div class="col-md-6 align-right">
-                            <a href="{{ BASE_URL }}/clients/delClient/<?php $tpl->e($_GET['id']); ?>" class="delete"><i class="fa fa-trash"></i> {{ __("links.delete") }}</a>
+                            <a href="{{ BASE_URL }}/clients/delClient/<?php $tpl->e($values->id); ?>" class="delete"><i class="fa fa-trash"></i> {{ __("links.delete") }}</a>
```
marcelfolaron, line 196: blade syntax

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -194,9 +202,9 @@
 
                 <div id='comment'>
 
-                <form method="post" action="{{ BASE_URL }}/clients/showClient/<?php echo $tpl->e($_GET['id']); ?>#comment">
+                <form method="post" value="true" action="{{ BASE_URL }}/clients/showClient/<?php echo $tpl->e($values->id); ?>#comment">
```
marcelfolaron, line 205: What does value="true" do in here? I don't think form has a "value" attribute

app/Domain/Clients/Templates/showClient.blade.php
```diff
@@ -194,9 +202,9 @@
...
                     <input type="hidden" name="comment" value="1" />
-                    @include("comments::includes.generalComment", ["formUrl" => BASE_URL . "/clients/showClient/" . $tpl->escape($_GET['id'])])
+                    <x-comments::list :module="'client'" :statusUpdates="'true'" moduleId="{{ $tpl->escape($values->id) }}" />
```
marcelfolaron, line 207: Can remove the escape

marcelfolaron (changes requested):
Really good job overall. Small requests across the board

ctanveer:
requested changes made

app/Domain/Clients/Templates/showClient.blade.php
ctanveer: without the value, the save property of array is set to null. In that case, the if condition in controller does not run

marcelfolaron (approved):

PR #2823, opened 2024-11-25
Laravel auth service

marcelfolaron:
### Description

This PR implements the laravel auth contract and wraps Leantime's authentication in as a new auth provider. 

### Type

- [ ] Fix
- [x] Feature
- [ ] Cleanup 

### Screenshot of the result


marcelfolaron:
@jeppekroghitk Could you check if this PR solves the issue you mentioned? This is using laravel guards and the middleware now just iterates through the available guards to find the correct provider

Issue #2838, opened 2024-11-29
[Bug]: bin/leantime system:update does not update


bedo2991:
### What is your set up?

Self Hosted Server

### Version

3.3.0

### Describe the issue

Running sudo bin/leantime system:update does not successfully update, though it says it does

### Reproduction steps

 ```
 sudo bin/leantime system:update
 Starting the updater

Version Check
-------------

 Your current version is: v3.3.0
 Checking latest version on Github...
 The latest Leantime version is: v3.3.3

Database Backup
---------------

mysqldump: unknown variable 'column-statistics=0'
Download & Extract
------------------

 Downloading latest version...
 Extracting Archive...

Disabling Plugins
-----------------

                                                                                                                        
 [OK] Plugins disabled successfully                                                                                     
                                                                                                                        

Applying Update
---------------

cp: cannot stat '/usr/share/nginx/leantime/storage/framework/cache/leantime/*': No such file or directory
                                                                                                                        
 [OK] Files were updated                                                                                                
                                                                                                                        

Clearing Cache
--------------

                                                                                                                        
 [OK] Clearing Cache Complete                                                                                           
                                                                                                                        

Re-enabling Plugins
-------------------

                                                                                                                        
 [OK] Plugins were enabled                                                                                              
                                                                                                                        

Summary
-------

                                                                                                                        
 [OK] Update applied Successfully                                                                                       
                                                                                                                        

Again: sudo bin/leantime system:update
 Starting the updater

Version Check
-------------

 Your current version is: v3.3.0
 Checking latest version on Github...
 The latest Leantime version is: v3.3.3
 ```
 
 (And the footer also says 3.3.0)

### Error Logs (LEANTIMEFOLDER/storage/logs)

Nothing there

Er1c85:
I just fixed this on my local system, I'll issue a PR

Er1c85:
I just realized that a pull request does not help you... 
If you open the php file on your system that controls the upgrade operation, `leantime/app/Command/UpdateLeantime.php`, you will see a line where it issues an `exec(cp ...` command.  The problem was it was trying to copy the unzipped files from the wrong location.  Originally it was `exec('cp -r '.storage_path('/framework/cache/').'/* '.APP_ROOT.'/');`, change this to `exec('cp -r '.storage_path('/framework/cache/leantime').'/* '.APP_ROOT.'/');`  This line should be around line 130... hope this helps!

PR #2844, opened 2024-12-04
Update UpdateLeantime.php

Er1c85:
Added a check of the retval of exec('cp') in case, like in 3.3.1, it fails for some reason.  In 3.3.1, it was trying to cp from .../cache/leantime/* but the new files were unzipped into the folder .../cache/* Also, added a try/catch block around ZipArchive because it was not installed by default on my system.  With the try/catch block, a nice error message can be displayed for the user

### Description

### Link to ticket

[Bug]: bin/leantime system:update does not update #2838

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result


marcelfolaron (changes requested):
Thanks Eric. Can you update the Exception type in your catch to start with a \Exception?

bedo2991:
A c is missing in io->success('New update zip file  - line 130.
The echo and print_r methods should probably be an io->text( ?

PR #2857, opened 2024-12-18
In progress - chip dropdown component

ctanveer:
### Description

Chip dropdown with quickadd htmx call added

### Link to ticket

### Type

- [ ] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result


app/Domain/Tickets/Hxcontrollers/Milestones.php
```diff
@@ -48,4 +52,40 @@ public function showCard()
...
 
     }
+
+    public function post($params)
+    {
+        $result = $this->ticketService->quickAddMilestone($params);
+        if (is_numeric($result)) {
+            $params['id'] = $result;
+            $this->tpl->setNotification($this->language->__('notification.milestone_created_successfully'), 'success');
+            $subject = $this->language->__('email_notifications.milestone_created_subject');
```
marcelfolaron, line 62: The email notification part should actually just be in the quickAddMilestone Service. That way people will get notified event if the milestone is being created via api

app/Domain/Tickets/Templates/showAll.blade.php
```diff
@@ -215,20 +215,19 @@ class="ticketDropdown statusDropdown colorized show {{ $class }} f-left"
                                 </x-global::elements.table.cell>
             
                                 <x-global::elements.table.cell data-order="{{ $milestoneHeadline }}">
-                                    <x-global::forms._archive.dropdownPill
-                                        class="ticketDropdown milestoneDropdown colorized show label-default milestone f-left"
-                                        id="{{ $milestoneDropdownId }}" 
-                                        :style="'background-color:' . $milestoneColor" 
+                                    <x-global::forms.chip
+                                        class="ticketDropdown milestoneDropdown colorized milestone show f-left"
```
marcelfolaron, line 219: Do we need those classes still? Or is this something we can move into a chip component css so that it's less tied to "tickets" and more generally applicable

app/Domain/Tickets/Templates/showAll.blade.php
```diff
@@ -215,20 +215,19 @@ class="ticketDropdown statusDropdown colorized show {{ $class }} f-left"
...
                                         type="milestone"
-                                        :parentId="$row['id']">
-                                        <x-slot name="buttonText">
-                                            {{ $milestoneHeadline }} <i class="fa fa-caret-down" aria-hidden="true"></i>
-                                        </x-slot>
-            
-                                        <x-global::actions.dropdown.item class="nav-header border">
+                                        :parentId="$row['id']"
+                                        :quickaddOption="true"
+                                        quickaddPostUrl="{{ BASE_URL }}/hx/tickets/Milestones"
```
marcelfolaron, line 225: You could just pass the hx-* attributes rather than create a new attribute. 
Basically just use hx-post and create the appropriate hxPost variable in your component. Maybe add hx-swap, hx-target, hx-trigger etc as well I am pretty sure you could just use the $attributes array that is available in the component and look for entries that start with hx

app/Views/Templates/components/forms/chip/index.blade.php
```diff
@@ -0,0 +1,57 @@
...
+    'quickaddOption' => false,
+    'quickaddPostUrl' => '',
+    'labelText'=> '',
+    'selectedKey' => '',
+    'options' => [],
+    'extraClass' => '',
+])
+
+<div {{ $attributes->merge([ 'class' => '' ]) }}>
+    <div class="dropdown ticketDropdown {{ $type }}Dropdown colorized show px-2 py-0.5 ml-1.5 {{ $extraClass }}">
```
marcelfolaron, line 14: Seems this is the old dropdown code, we should use our dropdown component for this

app/Views/Templates/components/forms/chip/quick-add.blade.php
```diff
@@ -0,0 +1,52 @@
...
+    'type' => '', //status, milestone, effort, priority, user, sprint
+    'parentId' => '',
+    'postUrl' => '',
+])
+
+@php
+    $inputName = $type == 'milestone' ? 'headline' : 'name';
+@endphp
+
+<div {{ $attributes->merge([ 'class' => 'mx-2' ]) }}>
```
marcelfolaron, line 11: From above: You can take the attributes array and look for keys() that start with 
'hx'. Then take all those keys and add them below to the form tag. 

I believe $attribute is a collection so you can do:

```php 

$hxAttributes = $attributes->filter(function($field, $key){            
    if(Str::startsWith($key, 'hx')) {
         return;
    }
});
 ```

File: app/Views/Templates/components/forms/chip/quick-add.blade.php [hunk 6]
```diff
@@ -0,0 +1,52 @@
...
+    <a href="javascript:void(0);" style="padding:10px; display:block; width:100%;"
+        id="{{ $type }}_new_link_{{ $parentId }}"
+        onclick="event.stopPropagation(); jQuery('#{{ $type }}_new_link_{{ $parentId }}').toggle('fast'); jQuery('#{{ $type }}_new_{{ $parentId }}').toggle('fast', function() { jQuery(this).find('input[name={{ $inputName }}]').focus(); });">
+        <i class="fas fa-plus-circle"></i> Add {{ $type }}
+    </a>
+    <div class="hideOnLoad" id="{{ $type }}_new_{{ $parentId }}"
+        style="padding-top:5px; padding-bottom:5px;">
+
+        <form 
+            hx-post={{ $postUrl }}
```
marcelfolaron, line 21: Then here you would just do `{{ $hxAttributes }}`

app/Views/Templates/components/forms/chip/quick-add.blade.php
marcelfolaron: PS: You would want to only show the non hx tags up here in the div tag

app/Views/Templates/components/forms/chip/quick-add.blade.php
```diff
@@ -0,0 +1,52 @@
...
+@endphp
+
+<div {{ $attributes->merge([ 'class' => 'mx-2' ]) }}>
+    <a href="javascript:void(0);" style="padding:10px; display:block; width:100%;"
+        id="{{ $type }}_new_link_{{ $parentId }}"
+        onclick="event.stopPropagation(); jQuery('#{{ $type }}_new_link_{{ $parentId }}').toggle('fast'); jQuery('#{{ $type }}_new_{{ $parentId }}').toggle('fast', function() { jQuery(this).find('input[name={{ $inputName }}]').focus(); });">
+        <i class="fas fa-plus-circle"></i> Add {{ $type }}
+    </a>
+    <div class="hideOnLoad" id="{{ $type }}_new_{{ $parentId }}"
+        style="padding-top:5px; padding-bottom:5px;">
```
marcelfolaron, line 18: Please remove inline styles

marcelfolaron (changes requested):

PR #2905, opened 2025-01-20
Notifications api call fix

ctanveer:
### Description

undefined api call

### Link to ticket

image

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result


app/Domain/Menu/Templates/headMenu.blade.php
```diff
@@ -208,12 +208,14 @@ function toggleNotificationTabs(active) {
                 jQuery('.notificationHandler').on('click', function () {
                     jQuery.ajax(
                         {
-                            type: 'PATCH',
-                            url: leantime.appUrl + '/api/notifications',
+                            method: 'GET',
+                            url: '/api/notifications',
```
marcelfolaron, line 212: We still need the appUrl otherwise subfolder installations will not work. Can you change the script section to be a module and then import instance-info.module.mjs
which contains appUrl

marcelfolaron (changes requested):

Issue #2936, opened 2025-02-16
[Cleanup] Update all internal calls to /api/*


marcelfolaron:
Frontend requests should be refactored to either be htmx calls or go through the jsonRpc api. This will reduce a ton of code duplication inside of the api/ module.

ellitedom03:
Hi! I'd like to take this one. I'll refactor the internal frontend calls from the api/ module to either htmx calls or go through the jsonRpc API as described. Will open a PR shortly.

ellitedom03:
Hi @marcelfolaron — I've submitted PR #3557 which addresses this issue. Would appreciate a review when you get a chance! 🙏

marcelfolaron:
hi, thanks. I should have updated the PR. most calls were indeed updated already I do see that you found a few new ones. #3424 - #3430 dealt with that but I see we may have missed a few.

ellitedom03:
Hi @marcelfolaron — just following up on PR #3557 which addresses the API route refactoring part of this issue. The changes are straightforward URL updates from legacy /api/* shims to canonical domain-scoped routes. Happy to make any adjustments if needed. Would appreciate a review when you get a chance!

ellitedom03:
Hi @marcelfolaron — checking in on this PR #3557 that's been open for review. It addresses the internal /api/* calls to canonical domain routes issue. The PR is ready for review and has been waiting since June 20. Would really appreciate a review — this is a straightforward refactor that brings API calls in line with the domain routing conventions. The bounty on Opire is linked to this issue. Happy to make any adjustments needed. Thanks!

PR #2940, opened 2025-02-17
Fix: [Cleanup] Update all internal calls to /api/*

saksham-jain177:
Proposed fix for issue: https://github.com/Leantime/leantime/issues/2936


marcelfolaron:
Is this AI generated? It's not even PHP...

PR #3162, opened 2025-11-13
Update sample.env

ThighCandy:
The semicolon at the end of the LEAN_LDAP_DEFAULT_ROLE_KEY will cause any LDAP users permissions to fail if they do not match any assignment rules in LEAN_LDAP_GROUP_ASSIGNMENT.

### Description

### Link to ticket

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result

### DIFF
--- a/config/sample.env
+++ b/config/sample.env
@@ -100,7 +100,7 @@ LEAN_LDAP_KEYS="{
 #        \"department\":\"department\"
 #      }"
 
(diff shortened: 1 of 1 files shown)


ThighCandy:
Just for further context, I think there was a copying of this value from the configuration.sample.php as that does has a semicolon at the end and would be needed there since it's PHP (I'm assuming?).

marcelfolaron:
Thank you!

PR #3292, opened 2026-03-01
fix: accessibility, view system, and composer improvements from phase-0

marcelfolaron:
## Summary

Independent improvements extracted from the phase-0 modernization branch. Each change creates a noop diff reduction when the feature branch eventually merges. All are low-risk and independently valuable.

This is PR 4 of the phase-0 modernization series.

### Accessibility
- **Remove `maximum-scale=1.0` from viewport meta** — This restriction prevents users with low vision from zooming the page, violating WCAG 2.1 SC 1.4.4 (Resize Text). One-line change in `header.blade.php`.

### View system
- **Add `globals` alias in ViewsServiceProvider** — Registers `app/Views/Templates` under both `global` (existing) and `globals` (new) aliases. This enables `<x-globals::componentName>` syntax used by the new Blade component library in the modernization branch. Purely additive — existing `<x-global::...>` syntax continues to work.
- **Update comments** — Clarify legacy extension support and blade priority registration.

### App composer
- **Pass `module` to app layout** — Exposes `strtolower(FrontcontrollerCore::getModuleName())` as a template variable. This enables a `data-module` attribute on `<body>` for domain-specific JS lazy-loading (used by the Vite migration).
- **Pass `themeBgUrl` to app layout** — Moves background image URL resolution from inline template logic to the composer, following the composer pattern.

### Header composer
- **Add accent3/4 to accents array** — Adds `false` entries for accent3 and accent4 in the accents array. These use CSS defaults from the theme files (added in PR #3290). The `false` value means "use CSS default, don't override via inline style."

## Verification
- PHPStan: **PASS** (0 errors)
- Laravel Pint: **PASS** (1376 files)
- `getModuleName()` exists on Frontcontroller (line 384)
- `getBackgroundImage()` exists on Theme class (already used by Header composer)

### DIFF
--- a/app/Core/UI/ViewsServiceProvider.php
+++ b/app/Core/UI/ViewsServiceProvider.php
@@ -62,11 +62,10 @@ class ViewsServiceProvider extends LaravelViewServiceProvider
 
             $factory = $this->createFactory($resolver, $finder, $app['events']);
 
-            // Backwards compatible view engine resolver
+            // Legacy template extensions — still needed for plugins with .tpl.php/.sub.php/.inc.php
             array_map(fn ($ext) => $factory->addExtension($ext, 'php'), ['inc.php', 'sub.php', 'tpl.php']);
 
-            // reprioritize blade
-            // Use blade engine for all things blade
+            // Use blade engine for all things blade (registered last to take priority)
             $factory->addExtension('blade.php', 'blade');
 
             // We will also set the container instance on this view environment since the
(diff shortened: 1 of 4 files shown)


marcelfolaron:
Re: Copilot's concern about `globals` alias not taking effect due to cached `viewPaths.php`:

Not an issue in practice. Leantime's upgrade process already handles manifest cache invalidation:

- `UpdateLeantime` command explicitly deletes `storage/framework/viewPaths.php` (line 159)
- `make clear-cache` (run as part of every `make build`) deletes it
- `cache:clearAll` artisan command also deletes it

The manifest is regenerated on the next request after deletion. Every version upgrade goes through one of these paths, so the new `globals` alias will be discovered automatically.

Issue #3342, opened 2026-03-31
Plugin marketplace HTTP 500 on install: nullable API fields crash non-nullable typed properties in MarketplacePlugin model (PHP 8.x)


shaunchokshi:
### What is your set up?

Self Hosted Server

### Version

3.7.3

### Describe the issue

**What is your set up?**
Self-hosted, non-Docker (bare metal / VPS with PHP-FPM). Also reproduced by other users on Docker (see linked report).

**Version:** 3.7.3

**PHP Version:** 8.4 (also affects 8.1+ due to strict typed property enforcement introduced in PHP 8.0)

---

**Describe the issue**

Opening the plugin marketplace or attempting to install any plugin throws an HTTP 500. The root cause is that `MarketplacePlugin.php` declares several properties as non-nullable typed (`public string $excerpt`, `public int $vendorId`, etc.), but the Leantime marketplace API returns `null` for one or more of these fields at runtime. PHP 8.0+ strictly enforces typed property assignments and throws a `TypeError` on null assignment, regardless of whether a default value is set.

The raw assignment in `Build.php:131` (`$property->$key = $value`) performs no null coercion before writing API data to the model, so any null field from the API hits the type constraint and crashes the request.

This renders the entire plugin system unusable.

(also creating a PR with the fix that worked for me; only tested on Debian 13 VPS host - after applying the patch, the plugins seem to be installing fine)

### Reproduction steps

1. Install Leantime 3.7.3 on PHP 8.x (Docker or bare metal)
2. Navigate to the plugin marketplace
3. Attempt to view or install any plugin
4. HTTP 500 is returned immediately
5. Optionally: attempt to install a licensed plugin, input a purchased license key, click Install — same result

### Error Logs (LEANTIMEFOLDER/storage/logs)

Two distinct errors, both from the same root cause:

```
Cannot assign null to property
Leantime\Domain\Plugins\Models\MarketplacePlugin::$excerpt of type string
at /var/www/leantime/app/Core/Support/Build.php:131
```

```
TypeError (code: 0):
Cannot assign null to property
Leantime\Domain\Plugins\Models\MarketplacePlugin::$excerpt of type string
#0 /var/www/leantime/app/Core/Support/Build.php(29): Build->setValue()
#1 /var/www/leantime/app/Domain/Plugins/Hxcontrollers/Details.php(41): Build->set()
```

Also reported on Docker by another user with identical stacktrace:
```
json_decode(): Passing null to parameter #1 ($json) of type string
[...]

Issue #3336, opened 2026-03-26
Licensed plugin install fails with HTTP 500 on v3.7.3 in Docker: null JSON / null excerpt


joleeb4ker:
### What is your set up?

Self Hosted Docker

### Version

3.7.3

### Describe the issue

I’m seeing a reproducible HTTP 500 when trying to install any purchased/licensed plugin from the plugin marketplace on Leantime v3.7.3 running in Docker behind a reverse proxy.

**Environment**
Leantime: 3.7.3
Deployment: Docker Compose
Database: MySQL 8.4
Reverse proxy: Caddy
Browsers tested: multiple, same result

I triple checked my setup and can pretty much rule out a configuration error.

### Reproduction steps

1. Open plugin marketplace
2. Select a licensed plugin, input my purchased license key
3. Click Install
4. UI appears to do nothing
5. Browser network tab shows:
    POST /hx/plugins/details/install
    500 Internal Server Error

### Error Logs (LEANTIMEFOLDER/storage/logs)

Relevant log output:

`json_decode(): Passing null to parameter #1 ($json) of type string
... /app/Domain/Plugins/Hxcontrollers/Details.php on line 35`

`Cannot assign null to property ... excerpt of type string
... /app/Core/Support/Build.php:131`

LindezaGrey:
I tested it with v3.7.0 there it worked without a problem - 3.7.3 crashed also

marcelfolaron:
This is a duplicate of #3342, which has the same root cause and has been verified.

**Root cause:** `app/Domain/Plugins/Models/MarketplacePlugin.php` declares properties like `$excerpt`, `$vendorId`, etc. as non-nullable typed, but the marketplace API can return `null` for these fields. The setter in `app/Core/Support/Build.php:131` does a raw assignment with no null guard, causing a `TypeError` on PHP 8.0+.

**Fix:** Make all API-sourced properties nullable in `MarketplacePlugin.php`. See #3342 for the full analysis and fix approach.

A fix PR is being tracked against #3342.

PR #3343, opened 2026-03-31
[PATCH] fix: make MarketplacePlugin model properties nullable for PHP 8.x compatibility

shaunchokshi:
### Description

The Leantime marketplace API returns null for several fields (e.g. excerpt,
vendorId) that were declared as non-nullable typed properties in
MarketplacePlugin.php. PHP 8.0+ strictly enforces typed property assignments
and throws a TypeError on null assignment, causing an HTTP 500 for any
marketplace interaction.

This patch:
1. Makes all API-sourced scalar properties nullable in MarketplacePlugin.php,
   consistent with the four properties ($startingPrice, $calculatedMonthlyPrice,
   $license, $rating) that were already correctly declared as ?string.
2. Adds a defensive null coercion in Build.php::setValue() using ReflectionProperty
   so all models hydrated from external API data are protected systemically.

Confirmed broken on PHP 8.4 (bare-metal PHP-FPM) and PHP 8.x (Docker).
PHP 7.x silently accepted null assignment, masking this bug.

### Link to ticket

https://github.com/Leantime/leantime/issues/3342

also: https://github.com/Leantime/leantime/issues/3336

### Type

- [X] Fix


marcelfolaron (changes requested):
We don't accept patch files. Please create a PR that will fix the issue noted.

PR #3348, opened 2026-04-01
Fix case sensitivity in secondary color setting

Delvar:
the setting string was incorrectly capitalised,

companysettings.secondaryColor shoudl be companysettings.secondarycolor

### Description

Another small typo i noticed while initially setting up and trying (and failing) to set company colours.

### Link to ticket

No ticket

### Type

- [X] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result

image


marcelfolaron:
Thank you!

PR #3395, opened 2026-05-27
Backend API surface for Leantime Mobile (TestFlight)

gloriafolaron:
## Summary

17 commits that together provide the backend surface area Leantime Mobile depends on. Without this PR merged + deployed to cloud, mobile login fails on `projects.leantime.io` and most features error out on missing methods or malformed responses.

This is the gating PR for the **first TestFlight beta**. Mobile is built, signed, and uploaded — it just can't talk to cloud Leantime productively until these land.

## What's in this PR

Grouped by concern. Every change is additive or defensive — no existing endpoints removed or renamed.

### 1. Server-authoritative session resolution (the architectural pattern)

Mobile clients authenticate via Bearer token. The login response (`/advancedAuth/getToken`) returns an access-token row id, NOT the underlying user id. So mobile needs RPC methods that resolve "who am I" from session/token rather than requiring the client to pre-know its own id (chicken-and-egg).

| Commit | Method | Change |
|---|---|---|
| `1287a9e9` | `Users::getUser($id = null)` | Default to `session('userdata.id')` when no id passed |
| `25a00ca9` | `Users::getUsersWithProjectAccess` | Drop `\$currentUser` param; resolve server-side |
| `e0f9726b` | `Projects::getProjectsUserHasAccessTo` | Same — resolve userId server-side |

Pattern matches the existing convention at `Tickets.php:1800`, `ShowCanvas.php:51`, `Calendar.php:97`. Closes the IDOR shape where client-controlled user id could mismatch authenticated session.

### 2. New methods mobile depends on

| Commit | Method | Purpose |
|---|---|---|
| `abb69f8b` | `Auth\\AccessToken::revokeCurrentToken()` | Server-side token invalidation on logout (security pass) |
| `634611bb` | `Projects::getProjectsByUserActivity` | Recency-sorted project chips for ADHD-friendly filter UX |
| `770aa487` | `Tickets::getAllDoneUserTickets` + `Tickets::markTicketReopen` | Done filter view + re-open from list |
| `7b755503` | `Tickets::markTicketDone` | Quick-complete from task list (resolves project's DONE status server-side, no client-side guessing) |

### 3. Fixes to existing methods that mobile hits

| Commit | Area | Change |
|---|---|---|
| `e65f44ba` | Tickets | Ship `statusClass` + `statusType` on `getAllOpenUserTickets` rows so mobile can render status pills without a second lookup |
| `44e82b03` | Tickets | `quickAddTicket` respects per-project default status (was hardcoded to status 3) |
| `8d1ee88b` | Tickets | Add `int` to `quickAddTicket` return type union |
| `3fb285da` | Tickets | Recognize calendar-date `dateToFinish` input; skip unwanted TZ conversion |
| `cad5048b` | Milestones | `getAllMilestones()` actually filters to `type='milestone'` (was returning regular tickets too) |
| `32e64cc1`, `2cd43b50` | Calendar | `mapEventData` tolerates NULL description / colors / dates instead of throwing |
| `13d234ee` | Comments | Fall back to `entity.projectId` when session has no currentProject (mobile rarely sets currentProject) |
| `bf553073` | Comments | `addComment` RPC-friendly — `\$entity` optional, `father` defaults to 0 |

### 4. The scalar response fix (the subtle one — please review carefully)

| Commit | Change |
|---|---|
| `0ab8dab0` | Stop coercing JSON-RPC scalar responses into single-element arrays |

**Context:** Leantime's JSON-RPC controller was wrapping every scalar response (int, string, bool) in a single-element array (`5` → `[5]`). This was masking the actual return type — mobile callers expecting an int would receive `[5]` and have to unwrap manually, which broke type contracts and caused the cluster of bugs around `quickAddTicket` returning `array` instead of `int`.

The fix removes the unconditional wrap. Scalar returns stay scalar; arrays stay arrays. Aligns with JSON-RPC spec (Section 5: result can be any JSON value).

**Potential blast radius:** any existing client (web app, integrations) that was working around the old behavior by unwrapping `[value]` will now receive `value` directly. Worth a grep through the JS / web side to make sure consumers handle both shapes during the transition, or to confirm nothing in the web app depends on the wrap.

## How this maps to mobile

| Mobile feature | Backend commit unblocking it |
|---|---|
| Cloud login on `projects.leantime.io` | `1287a9e9` (getUser session default) |
| Server-side logout (security pass) | `abb69f8b` (revokeCurrentToken) |
| Project chip recency sort on filter sheet | `634611bb` (getProjectsByUserActivity) |
| Task status pills on list view | `e65f44ba` (statusClass/Type on getAllOpenUserTickets) |
| Quick-complete from list + Done filter | `7b755503`, `770aa447` (markTicketDone/Reopen, getAllDoneUserTickets) |
| Mentions, assignees on tasks | `25a00ca9` (getUsersWithProjectAccess session-default) |
| Task creation reliability | `44e82b03`, `8d1ee88b`, `3fb285da` (quickAddTicket fixes) |
| Comments on tasks | `13d234ee`, `bf553073` (comments RPC fixes) |
| Milestone picker on task editor | `cad5048b` (getAllMilestones filter) |
| Calendar events | `32e64cc1`, `2cd43b50` (mapEventData null tolerance) |
| Most single-value RPC returns (count, id, etc.) | `0ab8dab0` (scalar wrap removed) |

## Testing

All changes verified locally on `Herd` setup against the iOS Simulator build of Leantime Mobile. Cloud login works on local (`http://127.0.0.1`) — fails on `projects.leantime.io` (production) precisely because production lacks this branch.

After merge + deploy, verification commands from authenticated session:

```bash
# 1. getUser session default
curl -s -X POST -H \"Content-Type: application/json\" -H \"Authorization: Bearer \$TOKEN\" \\
  -d '{\"jsonrpc\":\"2.0\",\"method\":\"leantime.rpc.Users.Users.getUser\",\"params\":{},\"id\":1}' \\
  https://projects.leantime.io/api/jsonrpc | jq

# Expected: full user record, NOT \"Required Parameter Missing: id\"

# 2. revokeCurrentToken
curl -s -X POST -H \"Content-Type: application/json\" -H \"Authorization: Bearer \$TOKEN\" \\
  -d '{\"jsonrpc\":\"2.0\",\"method\":\"leantime.rpc.Auth.AccessToken.revokeCurrentToken\",\"params\":{},\"id\":1}' \\
  https://projects.leantime.io/api/jsonrpc | jq

# Expected: success result, NOT \"Method doesn't exist\"

# 3. getProjectsByUserActivity
curl -s -X POST -H \"Content-Type: application/json\" -H \"Authorization: Bearer \$TOKEN\" \\
  -d '{\"jsonrpc\":\"2.0\",\"method\":\"leantime.rpc.Projects.Projects.getProjectsByUserActivity\",\"params\":{},\"id\":1}' \\
  https://projects.leantime.io/api/jsonrpc | jq

# Expected: array of projects sorted by user's recent activity
```

End-to-end mobile verification will happen via TestFlight Build 9 (next CI build after this merges).

## Risks / things to double-check

- **Scalar wrap (`0ab8dab0`):** see callout above. Could affect web consumers if any depend on the old `[value]` wrapping. Worth a quick grep through `app/Views` and any web JS for patterns like `result[0]` on RPC responses.
- **Session-default methods:** when called from a context with no session (cron jobs, CLI), these now return false or empty rather than the previous behavior (passing 0 as a literal user id). Should be safer, but if any internal callers depend on the old behavior, they'd see different results.
- **revokeCurrentToken:** uses `ApiRequest::getBearerToken()` then looks up via `tokenRepo->findToken()`. Defense-in-depth user check is in there — verify the check matches existing patterns (probably worth a second eye).

## Deploy notes

After merge to master:

1. Deploy to staging if you have one — exercise the curl checks above
2. Deploy to `projects.leantime.io` 
3. Notify mobile side so we can fire the next CI build and the new TestFlight goes through cloud login cleanly

## Related mobile work

This PR is the unblock for `Leantime/leantime-mobile` TestFlight Build 8 (currently uploaded but unable to use cloud login). Mobile-side fixes that don't depend on this PR (workspace probe removal, dev-mode error silencing) are sitting uncommitted locally and will ship in TestFlight Build 9 once this PR merges.

### DIFF
--- a/app/Domain/Api/Controllers/Jsonrpc.php
+++ b/app/Domain/Api/Controllers/Jsonrpc.php
@@ -236,10 +236,12 @@ class Jsonrpc extends Controller
             return $this->returnServerError($e, $id);
         }
 
-        if ($method_response !== null) {
-            if (! settype($method_response, 'array')) {
-                $method_response = [$method_response];
-            }
+        // Convert objects to associative arrays for JSON serialization, but pass
+        // scalars and arrays through as-is. The previous `settype($var, 'array')`
+        // coerced scalars to `[$scalar]`, which broke RPC methods returning ints
+        // (e.g., addTicket returning a new ticket ID).
+        if ($method_response !== null && is_object($method_response)) {
+            $method_response = (array) $method_response;
         }
 
         return $this->returnResponse($method_response, $id);
@@ -368,7 +370,7 @@ class Jsonrpc extends Controller
      *
      * @see https://jsonrpc.org/specification#response_object
      */
-    private function returnResponse(?array $returnValue, int|string|null $id = null): Response
+    private function returnResponse(mixed $returnValue, int|string|null $id = null): Response
     {
         /**
          * No IDs imply notification and MUST not be responded to
--- a/app/Domain/Api/Controllers/Jsonrpc.php
+++ b/app/Domain/Api/Controllers/Jsonrpc.php
@@ -366,7 +366,17 @@ class Jsonrpc extends Controller
     }
 
     /**
-     * Echos the return response
+     * Echos the return response.
+     *
+     * @param  mixed  $returnValue  The return value from the RPC method. Widened from
+     *                              `?array` because the upstream `settype` coercion that
+     *                              wrapped scalars into single-element arrays was removed
+     *                              (it broke methods returning ints — e.g. addTicket's
+     *                              new ticket id was being delivered as [id]). Per the
+     *                              JSON-RPC 2.0 spec §5, `result` MAY be any JSON value;
+     *                              caller code in `executeRPC` already casts objects to
+     *                              associative arrays before reaching here, so in practice
+     *                              this is array|scalar|null.
      *
      * @see https://jsonrpc.org/specification#response_object
      */
--- a/app/Domain/Api/Controllers/Users.php
+++ b/app/Domain/Api/Controllers/Users.php
@@ -49,7 +49,10 @@ class Users extends Controller
                 $projectId = $params['projectUsersAccess'];
             }
 
-            $users = $this->userService->getUsersWithProjectAccess($projectId);
+            // Pass projectId by name so we don't depend on positional
+            // order and so the optional $currentUser parameter resolves
+            // to the session user inside the service per its docblock.
+            $users = $this->userService->getUsersWithProjectAccess(projectId: $projectId);
 
             if (isset($params['query'])) {
                 $query = $params['query'];
--- a/app/Domain/Api/Controllers/Users.php
+++ b/app/Domain/Api/Controllers/Users.php
@@ -49,7 +49,7 @@ class Users extends Controller
                 $projectId = $params['projectUsersAccess'];
             }
 
-            $users = $this->userService->getUsersWithProjectAccess(session('userdata.id'), $projectId);
+            $users = $this->userService->getUsersWithProjectAccess($projectId);
 
             if (isset($params['query'])) {
                 $query = $params['query'];
--- a/app/Domain/Projects/Services/Projects.php
+++ b/app/Domain/Projects/Services/Projects.php
@@ -13,6 +13,7 @@ use Leantime\Core\Language as LanguageCore;
 use Leantime\Core\Support\Avatarcreator;
 use Leantime\Core\Support\FromFormat;
 use Leantime\Domain\Auth\Models\Roles;
+use Leantime\Domain\Auth\Services\Auth;
 use Leantime\Domain\Files\Services\Files;
 use Leantime\Domain\Goalcanvas\Repositories\Goalcanvas as GoalcanvaRepository;
 use Leantime\Domain\Ideas\Repositories\Ideas as IdeaRepository;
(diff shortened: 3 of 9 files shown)


gloriafolaron:
@marcelfolaron — this is the backend gate for the mobile TestFlight build. Once this merges + deploys to projects.leantime.io, cloud login on the mobile app works end-to-end. See PR description for the full mapping of which commit unblocks which mobile feature, and curl checks for post-deploy verification.

app/Domain/Api/Controllers/Jsonrpc.php
```diff
@@ -368,7 +370,7 @@ private function prepareParameters(array $params, array $methodParams): array
      *
      * @see https://jsonrpc.org/specification#response_object
      */
-    private function returnResponse(?array $returnValue, int|string|null $id = null): Response
+    private function returnResponse(mixed $returnValue, int|string|null $id = null): Response
```
marcelfolaron, line 383: why the change to mixed?

app/Domain/Api/Controllers/Users.php
```diff
@@ -49,7 +49,7 @@ public function get(array $params): Response
                 $projectId = $params['projectUsersAccess'];
             }
 
-            $users = $this->userService->getUsersWithProjectAccess(session('userdata.id'), $projectId);
+            $users = $this->userService->getUsersWithProjectAccess($projectId);
```
marcelfolaron: The method signature for getUserrsWithProjectAccess requires the user id

app/Domain/Projects/Services/Projects.php
```diff
@@ -922,13 +922,21 @@ public function getProjectRole($userId, $projectId): mixed
...
      * @api
      */
-    public function getProjectsUserHasAccessTo($userId): false|array
+    public function getProjectsUserHasAccessTo(): false|array
     {
+        // Server-authoritative: resolve the user from the auth session.
+        // Never accept user identity from the client (IDOR risk). The project
+        // role/permission filtering inside getUserProjects() runs identically
+        // regardless of where the user ID comes from.
+        $userId = (int) session('userdata.id');
```
marcelfolaron: Except this is a service that can be used by the jsonrpc api and users can write custom code running as admins. Which is exactly why we have additional access logic on the repo itself. This is way too aggressive and changes the api footprint fundamentally. If we want to add safety mechanisms into the service let's check for roles: Admins and owners can call with user id all other roles can call but will only receive their own projects. Changing the method signatuire for our API like this is a recipe for disaster and would require a major api version change

app/Domain/Users/Services/Users.php
```diff
@@ -357,8 +372,15 @@ public function usernameExist(string $username, int|string $notUserId = ''): boo
      *
      * @api
      */
-    public function getUsersWithProjectAccess(int $currentUser, int $projectId): array
+    public function getUsersWithProjectAccess(int $projectId): array
```
marcelfolaron: Same comment as aboive. We can't just change the method signature and api set up

gloriafolaron:
@marcelfolaron — addressed all 4 review comments in commit 281b7b1c. Summary:

**Projects.php:935 (`getProjectsUserHasAccessTo`)** — your prescription implemented exactly:
- Restored `$userId` as the first param with default `null` so backwards-compatible for `ShowTicket::run`, `NewTicket::run`, and any plugin callers
- Role check inside: null/0 → session user, admin/owner → honored, non-admin passing different id → ignored + fallback to session

**Users.php:375 (`getUsersWithProjectAccess`)** — same pattern:
- Restored `$currentUser` as first param (default null), `$projectId` stays second
- Param order preserved from the original signature so positional callers keep working
- `$projectId` has a trivial `0` default (PHP requires defaults at tail) but explicit validation at top returns `[]` if missing
- Role check on `$currentUser` matches the Projects.php pattern

**Api/Controllers/Users.php:52** — fixed to use PHP named-arg syntax: `getUsersWithProjectAccess(projectId: \$projectId)`. Doesn't depend on positional order; service resolves session user internally.

**Jsonrpc.php:373 (`mixed` type)** — added thorough docblock explaining the widening. TL;DR: the upstream `settype(\$var, 'array')` was wrapping scalar returns into `[scalar]`, which broke methods returning ints (`addTicket` returning a new ticket id was being delivered as `[id]` instead of `id`). JSON-RPC 2.0 spec §5 allows `result` to be any JSON value.

All passes `./vendor/bin/pint` (no additional style fixes needed).

CI should re-run on the new commit; will let you know if anything regresses.

PR #3396, opened 2026-05-27
fix(calendar): tolerate MySQL zero-date sentinel in getCalendar

gloriafolaron:
## Summary

`Calendar\Services\Calendar::getCalendar` was throwing a generic -32000 server error on instances with historical zero-date data, killing the entire calendar feed because of one bad row.

## The bug

Inside the event-mapping loop, `CarbonImmutable::parse(\$value['dateFrom'])` and `parse(\$value['dateTo'])` were called unguarded. Rows where these columns hold the MySQL zero-date sentinel (\`0000-00-00 00:00:00\` or \`0000-00-00\` — placeholders valid in lenient MySQL mode but refused by Carbon) caused Carbon to throw:

> The string is not a valid date time string to parse as Database string : 0000-00-00 00:00:00 is not a valid value.

That exception bubbled to JSON-RPC as \`-32000 Server error\`, blanking the calendar tab on mobile (and any other consumer of \`Calendar.getCalendar\`).

## The fix

Added a tiny \`isZeroDate(?string \$value): bool\` helper that recognizes the standard sentinel forms (\`null\`, \`''\`, \`0000-00-00\`, \`0000-00-00 00:00:00\`). The event loop now skips events whose dateFrom/dateTo hits this guard before reaching the parser.

Matches the mobile-side \`parseCalendarDate\` pattern documented in our existing handling — keeps defensive parsing consistent across surfaces.

## How discovered

2026-05-27 during mobile cloud-login QA on projects.leantime.io. The Calendar tab errored on app launch. One zero-date row was enough to break the whole feed for every cloud user.

## Out of scope (memo'd as followup)

SQL audit + data cleanup to find and null-out the underlying zero-date rows. The schema should allow NULL on these columns; cleanup is a separate operations task. Tracked in the mobile project notes under \`project_calendar_zero_date_backend\`. Defensive parsing here means we don't have to wait on data cleanup to ship the fix.

## Testing

- Local: hit the affected code path via the curl chain from the mobile QA session — getCalendar now returns the rest of the feed instead of throwing.
- No new tests added (no test harness in this file's neighborhood — same pattern as the existing changes in \`fix/mobile-api-contract-and-jsonrpc-scalar-wrap\`).
- \`./vendor/bin/pint\` clean.

### DIFF
--- a/app/Domain/Calendar/Services/Calendar.php
+++ b/app/Domain/Calendar/Services/Calendar.php
@@ -361,6 +361,18 @@ class Calendar
         foreach ($dbUserEvents as $value) {
             $allDay = filter_var($value['allDay'], FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
 
+            // Skip events whose stored dates are the MySQL zero-date sentinel
+            // (`0000-00-00 00:00:00`) or empty. Carbon refuses to parse these
+            // and would throw a "not a valid date time string" exception that
+            // kills the entire getCalendar response. Treating them as invalid
+            // and filtering them out is safer than dropping the whole feed —
+            // see project_calendar_zero_date_backend memo for the data-cleanup
+            // followup.
+            if ($this->isZeroDate($value['dateFrom'] ?? null)
+                || $this->isZeroDate($value['dateTo'] ?? null)) {
+                continue;
+            }
+
             // Filter events by date range if specified
             if ($from || $until) {
                 $eventStart = CarbonImmutable::parse($value['dateFrom']);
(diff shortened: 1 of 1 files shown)


gloriafolaron:
@marcelfolaron — addressed your feedback in commit 945523f. Now uses `dtHelper()->parseDbDateTime()` which internally calls `DateTimeHelper::isValidDateString()` (covers the MySQL zero-date sentinel, the epoch sentinel `1969-12-31 00:00:00`, AND empty values — strictly better than what I had) and respects Leantime's DB timezone instead of going through `CarbonImmutable::parse` directly. Dropped the inline `isZeroDate()` helper since it duplicated existing infrastructure and missed the epoch case. Behavior is unchanged from the user's perspective — invalid-date rows are skipped, rest of the calendar feed renders. CI re-running.

Issue #3398, opened 2026-05-27
[mobile] Push notifications: backend channel + device token management


gloriafolaron:
## Background

Mobile app (TestFlight) currently has notification settings toggles (`notificationsEnabled`, `dueDateReminders`, `mentionAlerts`) but they're **local-only**. To actually deliver notifications to mobile users we need to add push as a delivery channel alongside the existing email / Slack / Discord / Mattermost / Zulip channels.

**Critically: ~80% of what's needed already exists.** Notification generation, filtering (per-project relevance + event-type categories + mention override), the `Messengers` channel-abstraction pattern, the cron queue worker pattern — all done. This issue is specifically about adding the push channel + supporting infrastructure.

## What's already in place (no work needed)

- `zp_notifications` table + in-app inbox
- `Notifications.getAllNotifications()` + `Notifications.markNotificationRead()` RPC methods
- `Projects::notifyProjectUsers()` hot path — fires on every assignment, mention, comment, status change, canvas item edit
- Three-layer preference filtering (relevance, category, mention override) in `NotifyProjectUsers` listener
- `Messengers` service pattern — push slots in alongside Slack/Discord/etc.
- `zp_queue` table + `queue:emails` / `queue:httprequests` cron worker pattern — easy to mirror for `queue:push`

## Scope of this issue

Build the push channel on top of what's there. Mobile side is a separate effort (Expo notifications integration, permission flow, token registration, payload handling).

## Architecture decisions to confirm

| Decision | Recommendation | Rationale |
|---|---|---|
| **Push transport library** | Thin wrappers around APNs HTTP/2 endpoint + FCM HTTP v1 API; no heavy framework | We've seen pain from over-abstracted notification libs; APNs/FCM HTTP APIs are stable + well-documented |
| **Token storage** | New `zp_device_tokens` table, tokens encrypted at rest, user-scoped | Mirrors existing security patterns; tokens are sensitive (account-impersonation risk if leaked) |
| **Provider abstraction** | New `PushMessenger` service alongside `Messengers` (Slack/Discord/etc.); enumerated via `Workers::PUSH_NOTIFICATIONS` | Consistent with existing channel pattern |
| **Preference source of truth** | Backend canonical, mobile mirrors | Avoids divergence; mobile already accepts that things like `usersettings.X` live server-side |
| **Badge count semantics** | Per-workspace unread total (single workspace = single badge) | Matches the workspace-scoped model |
[...]

PR #3401, opened 2026-05-28
feat(notifications): mobile push device tokens (Expo + FCM) + getUnreadCount RPC (#3398)

gloriafolaron:
**Draft for early feedback.** Per discussion on #3398, this lands the structural pieces of mobile push notification backend work that follow established conventions, then waits for your input on the Notification-class / channel-wiring approach before committing to more.

## What's in this PR

| File | What |
|---|---|
| \`Install.php\` | \`update_sql_30502\` — creates \`zp_device_tokens\` table (Schema builder, matches recent 30501 convention) |
| \`Notifications/Repositories/DeviceTokens.php\` | New repo: save (upsert), findActiveByUser, invalidate, invalidateForUser, prune |
| \`Notifications/Services/Notifications.php\` | Three new \`@api\` methods: \`registerPushToken\`, \`unregisterPushToken\`, \`getUnreadCount\` |

## Design choices to confirm

1. **userId from session, NOT client param** — same convention as the \`Users.getUser\` fix in #3395. A bearer token shouldn't be able to register push devices on someone else's account. Mobile passes (token, platform, deviceName), backend resolves userId.

2. **Soft-delete via \`invalidatedAt\`** rather than hard-delete on unregister. Audit trail + prune job sweeps stale rows past 30 days.

3. **\`VARCHAR(255)\` for token** instead of TEXT. Expo's \`ExponentPushToken[xxx]\` format is ~50-100 chars; 255 gives headroom AND lets us unique-index without prefix-key gymnastics.

4. **Light token format validation** — reject if it doesn't start with \`ExponentPushToken[\` / \`ExpoPushToken[\`. Cheap; catches obvious junk.

## What's deliberately NOT in this PR (awaiting your input)

- **ExpoChannel + Notification class refactor** — wanted your call on:
  - (a) Wrap existing \`notifyProjectUsers\` payloads in Laravel \`Notification\` classes (\`TaskAssignedNotification\` etc.) and use the full Laravel Notifications flow
  - (b) Add a parallel \`PushMessenger\` service alongside existing Slack/Discord ones in \`Messengers.php\`, leaving the rest unchanged

- **\`User::routeNotificationForExpoPushNotifications()\`** — shape depends on (a)/(b)

- **Preferences RPC** (\`getNotificationPreferences\` / \`updateNotificationPreferences\`) — preference storage is currently scattered across:
  - \`usersettings.{userId}.projectNotificationLevels\` (per-project all/my_work/muted)
  - \`usersettings.{userId}.notificationEventTypes\` (per-category)
  - legacy \`usersettings.{userId}.projectMutedNotifications\`

  Want your call on whether mobile reads/writes all three or we consolidate during this work.

- **Push queue worker registration** — depends on (a)/(b)

- **Cron job for \`prune()\`** — small, easy follow-up

## Why DRAFT

This shape is safe enough to ship to staging — mobile-side scaffolding (\`Leantime/leantime-mobile@f59a457\`) gracefully no-ops when these RPCs aren't deployed (returns -32601 → silent skip). So this PR can land alone without breaking mobile, and we iterate from here on the bigger architectural questions.

## Mobile side already prepped

Mobile (\`Leantime/leantime-mobile@f59a457\`):
- \`expo-notifications\` + \`expo-device\` installed
- \`deeplinkRouter.ts\` — pure function mapping push payloads → expo-router paths (26 tests pass)
- \`pushPermissions.ts\` — permission state machine
- \`pushToken.ts\` — register/unregister with graceful -32601 degradation

When you land this + we agree on (a) vs (b), mobile work is ~half a day.


gloriafolaron:
Heads up — we're pivoting to FCM instead of Expo Push (see #3398 comment).

**Good news for this PR:** the structural pieces in this draft are all still valid:
- `zp_device_tokens` table (token is just a string, no Expo-specific shape)
- `registerPushToken` / `unregisterPushToken` / `getUnreadCount` RPC
- Repository methods (save / findActiveByUser / invalidate / prune)

**What still needs to be added (post-FCM-decision):**
- Push channel implementation — `kreait/laravel-firebase` or `laravel-notification-channels/fcm` (your call on which package)
- Notification class refactor (the original (a) vs (b) decision from this PR's opening comment)
- `User::routeNotificationForFcm()` instead of `routeNotificationForExpoPushNotifications`
- Preferences RPC
- Push queue worker / scheduler

Want me to keep this PR open as the structural foundation, or close it and you'll spin up a fresh branch with the full FCM-based plan once Apple Dev Portal + Firebase Console are set up?

gloriafolaron:
Three things changed since this draft — flagging so the PR shape can update before more work goes in:

1. **Service: Expo Push → Firebase FCM.** Mobile re-scaffolded against `@react-native-firebase/messaging` on a `feature/firebase-push` branch (held local until Apple Dev Portal + Firebase Console setup lands). Token-format validation in `DeviceTokens.php` (`ExponentPushToken[` prefix check) doesn't match FCM tokens — drop the prefix check or swap for a length sanity check.

2. **Storage: separate table vs piggyback on `zp_access_tokens`.** Laravel doesn't ship device-token storage anywhere — checked auth/sanctum/notifications/sessions. But `zp_access_tokens` already has one row per device-session (via `device_name` on `AdvancedAuth.getToken`). Two options:
   - **(A)** Keep this PR's `zp_device_tokens` table — clean separation.
   - **(B)** Add 3 nullable columns to `zp_access_tokens`: `push_token / push_platform / push_token_updated_at`. Lifecycle tied to login/logout (revoke bearer → push registration goes with it). Smaller schema delta.

   Leaning B for the lifecycle coupling. Sessions table is wrong fit too — opaque payload, 8h lifetime, and Leantime ships `session.driver=file` so the table doesn't exist anyway.

3. **RPC names diverged.** Mobile now calls:
   ```
   Notifications.registerDevice({ pushToken, platform })
   Notifications.unregisterDevice({ pushToken })
   ```
   PR has `register/unregisterPushToken({token, platform, deviceName})`. Pick whichever; mobile will follow.

gloriafolaron:
Added `markNotificationUnread` to this PR — commit `59853e9`.

**Why:** mobile inbox now supports a swipe-to-mark-unread gesture (iOS Mail / Messages convention, right-swipe reveals a contextual pink "Unread" pill on read rows / mint "Read" pill on unread rows). Without a backend symmetric to `markNotificationRead`, the gesture couldn't actually persist the "needs review again" state.

**Shape:**
- `Notifications\Services\Notifications::markNotificationUnread(int $id, int $userId): bool` — `@api`
- `Notifications\Repositories\Notifications::markNotificationUnread(int $id): bool`

userId param follows the mobile-owns-explicit-RPC-params convention even though `id` is the primary key. Keeps audit logs consistent with the rest of the Notifications RPC surface.

**Mobile side:** `Leantime/leantime-mobile@aa63679` — soft-fails to `-32601` if backend doesn't have the method, with optimistic rollback. So the mobile UX degrades gracefully on cloud (until this lands) but the gesture only does what it visually promises once the backend ships.

Light addition — happy to split into its own PR if it muddies the push-notification scope here. Two related branches in this repo are easier to land together IMO.

marcelfolaron:
Pushed two fixes to get CI green:
- **pint:** fixed the `unary_operator_spaces`/blank-line style issue in `Notifications` service.
- **unit:** the push work added a 5th constructor dependency (`DeviceTokensRepository`) to the `Notifications` service, which broke `NotificationsServiceTest` (it constructs the service directly with the old 4-arg arity). Merged `master` in and updated the test's construction sites to inject a `DeviceTokens` stub. The `NotificationsServiceTest` suite passes locally (5 tests, 10 assertions).

Note: `license/cla` is still pending — that needs the CLA signed on your account; I can't resolve that one.

PR #3430, opened 2026-05-30
Retire the legacy Api REST controllers (binary/upload → domains, Projects sort → JSON-RPC, delete dead Tags/Timer)

marcelfolaron:
## Summary

This completes the retirement of the legacy `Api` REST controllers. The JSON anti-pattern (a second API stack parallel to JSON-RPC, with business logic leaking into controllers) is now fully removed — the `Api` domain keeps only the real API surface (`ApiKey`/`NewApiKey`/`DelAPIKey`/`Jsonrpc`) plus the every-page infra (`StaticAsset`, `I18n`).

Two things happen here:

### 1. Projects JSON → JSON-RPC (with authorization that was missing)
`/api/projects` overloaded **three** JSON operations — kanban status+sort, gantt sort, and gantt date-patch — all with **zero authorization** (any logged-in user could re-status/re-sort/patch arbitrary projects by id). These now go through JSON-RPC with proper auth:

- New authorized `@api` wrappers on `Projects\Services\Projects`: `patchProjectStatusAndSorting`, `sortProjects`, `patchProject`. Because the JSON-RPC endpoint has **no controller-level role gate**, each enforces **manager+** and **per-project access** (matching the manager+ gate on `Projects\Controllers\ShowAll`, which renders this UI). Mixed gantt payloads resolve `ticket-{id}` entries to their project before the access check.
- The 3 `projectsController.js` callers move from `jQuery.ajax('/api/projects')` to `leantime.rpc('Projects.Projects.…')`.

### 2. Relocate binary/upload endpoints into their domains
Image GET + file/photo/avatar/logo upload POST handlers move out of `Api/Controllers` into the proper domains, served by domain `routes.php` (canonical paths **and** backward-compat `/api/*` aliases):

| New controller | Canonical | Alias (kept) |
|---|---|---|
| `Users\Controllers\ProfileImage` | `/users/profileImage/{id}` | `/api/users` |
| `Projects\Controllers\ProjectImage` | `/projects/projectImage/{id}` | `/api/projects` |
| `Files\Controllers\Upload` | `/files/upload` | `/api/files` |
| `Setting\Controllers\Logo` | `/setting/logo` | `/api/setting` |

**Why the `/api/*` aliases are permanent (not redirects):** ~70 `` references across core templates *and the private plugin submodule* (Copilot, StrategyPro, PgmPro, Whiteboardscanvas, Llamadorian) hardcode `/api/users?profileImage=` and `/api/projects?projectAvatar=` and cannot be rewritten from this repo. The aliases serve the image **directly** (no 302) to avoid an avatar redirect-storm on kanban/list views. Existing `` refs are intentionally left untouched — they work via the alias, so churning 70 first-party templates would add risk for no functional gain.

### Security hardening (this is the `security/comprehensive-remediation` lineage)
- **Setting logo upload** now requires **admin+** — the legacy endpoint let *any* authenticated user overwrite the company logo.
- **Project avatar upload** now requires **manage rights** on the active project — the legacy endpoint had no role check.
- **Projects sort/status/patch** now authorized (see above).
- De-`@api`'d the file-upload setters (`setProfilePicture`, `setProjectAvatar`, `setLogo`): they trust a caller-supplied id / accept `$_FILES`-shaped arrays, so they must not be JSON-RPC callable; they now run only behind the authorized controllers.

### Supporting changes
- Moved `buildImageResponse` off the heavy `Api\Services\Api` (which injects 4 repositories) to a dependency-free `Core\Support\ImageResponse`.
- Added `Core\Routing\ControllerDispatch` to bridge the alias routes into the legacy-style controllers (generalises the existing Blueprints bridge).
- Deleted `Api/Controllers/{Users,Projects,Files,Setting}.php` and the dead `filterUsersByQuery` helper + its dead GET/PATCH branches.

## Behaviour preserved
- Avatar GET still returns SVG/binary; `profileImage=me` and `=false` (ghost) still work.
- The `/api/files` JSON response shape is unchanged (Tiptap/Uppy read `module`/`encName`/`extension`/`realName`/`fileId`).
- Upload flows (`location.reload()` on success) unchanged.

## Tests
- Authorization tests for the 3 RPC wrappers + the `userCanManageProject` helper (non-manager rejected, manager-without-access rejected, manager-with-access allowed, ticket→project resolution, control-field stripping).
- Relocated the `buildImageResponse` tests to `ImageResponseTest`; dropped the deleted `filterUsersByQuery` tests.
- `phpstan` clean, `pint` clean, full unit suite green (367 tests).

## Note
This is independent of the open Users-JSON PR #3429 (file-disjoint apart from a non-overlapping region of `Users.php`), so the two merge cleanly in either order.


marcelfolaron:
### Follow-up commit: final Api REST cleanup (Tags + Timer)

Folded the last batch of the Api→JSON-RPC initiative into this PR so it retires the legacy REST controllers as one cohesive change:

- **Deleted `Api\Controllers\Timer`** — dead; punch in/out is handled by the HTMX `Timesheets\Hxcontrollers\Stopwatch`, 0 callers remain.
- **Deleted `Api\Controllers\Tags`** — dead; the tag autocomplete already calls `leantime.rpc('Tags.Tags.getTags', …)` (`ticketsController.js`), 0 callers remain.
- **Security — closed a Tags IDOR**: the retired controller forced `session('currentProject')`, but the `getTags(int $projectId, string $term)` RPC method it was replaced by accepts an arbitrary `projectId` and had **no authorization** (cross-project tag enumeration). `getTags` now gates on `projectRepository->isUserAssignedToProject` — the full access check (admin/owner, org-wide `all`, client-level, assigned users) — returning `[]` otherwise. No regression for the legitimate current-project path. Unit tests added.

**End state of `app/Domain/Api/Controllers/`:** `ApiKey` / `NewApiKey` / `DelAPIKey` / `Jsonrpc` (API-key mgmt + the RPC endpoint) and `StaticAsset` / `I18n` (every-page infra). `Canvas` / `Goalcanvas` remain — they have **0 live callers** and appear orphaned, but are left for the Blueprints consolidation to retire rather than double-migrating here.

marcelfolaron:
### Addressed the 3 review comments (commit 71c4d70)

1. **`ControllerDispatch::dispatch` — string fragment lost / `ArgumentCountError`** ✅
   Now mirrors `Frontcontroller::executeAction` exactly: `… : $controller->getResponse($response)`. `HtmxController::getResponse($fragment)` requires the argument, so a fragment-returning controller routed through the bridge would have thrown — latent here (the four relocated controllers all return `Response`), but the bridge is reusable, so it's fixed.

2. **`Upload` returns 500 for missing request parts** ✅
   Missing `file`/`module`/`moduleId` now returns **400** with `{status:'error', message:'Missing file, module or moduleId'}`.

3. **`Upload` IDOR — arbitrary `module`/`moduleId` with no access control** ✅
   Good catch — this was a pre-existing gap (the old `Api\Controllers\Files` had it too). The endpoint now authorizes via new `Files::userCanUploadToModule()`, which mirrors the read-path model in `getFileForUser()`: admins/owners bypass; `project`/`ticket` targets require `isUserAssignedToProject` on the owning project (ticket resolved to its project via `getProjectIdForFile`); targets with no project mapping (e.g. `wiki`) fall back to that path's existing behaviour. No regression for the legitimate Tiptap/Uppy flows (they always upload to the project/ticket the user is viewing). Unit tests added in `FilesServiceTest`.

phpstan clean · pint clean · full unit suite green (373).

For the record on the other open PR: **#3429 (Users-JSON) — Copilot generated no comments** (clean review). The merged batch PRs' feedback was addressed at the time.

marcelfolaron:
### Addressed the re-review + reworked the routing/response architecture (commit 5495d84)

**1. `ControllerDispatch` getResponse() comment** — resolved by **deleting `ControllerDispatch` entirely.** It was just `Frontcontroller::executeAction` re-applied to Laravel routes, which propagates the legacy pattern rather than reducing it. The four relocated endpoints are now **plain Laravel controllers** — constructor DI, route-bound actions returning a `Response`/`Responsable`:

```php
Route::get('/users/profileImage/{id?}', [ProfileImage::class, 'show']);
```

Verified all 12 routes register and bind to `Controller@method` through the app router. No bridge, no `callAction`/`getResponse` machinery.

**2. `/api/users` GET/POST drops legacy PATCH / `projectUsersAccess`** — correct catch, and it was a real cross-PR hazard. Since #3429 (which migrates those JS callers to JSON-RPC) is now merged, I **rebased this branch onto master**, so `usersRepository.js`/`helperRepository.js`/`mention.js` no longer hit `PATCH /api/users` or `?projectUsersAccess` — the only remaining `/api/users` caller is the photo-upload `POST`, which `ProfileImage::upload` handles. The relocated routes only need GET-image + POST-upload now.

**Bonus — response types now live in the HTTP layer behind a contract** (per maintainer feedback): `ImageResponse` moved from `Core\Support` to `app/Core/Http/Responses/` and implements a new `Core\Http\Responses\Contracts\LeantimeResponse` (extends Laravel's `Responsable`). Controllers `return new ImageResponse($image)`; the router renders it. This sets up the home + contract for future response types (e.g. a `JsonRpcResponse`).

phpstan clean · pint clean · full unit suite green (376).

> Note: the `blueprintsDispatch` bridge in `Blueprints/routes.php` is the same legacy pattern, but converting those 8 canvas controllers to native binding is a separate, larger refactor (they're tied to `$this->tpl` rendering + the merged-`$params` contract + return HTMX fragments across 16 canvas variants). Tracking it as a dedicated follow-up rather than bundling that risk here.

PR #3435, opened 2026-05-30
refactor: merge Strategy domain into Blueprints

marcelfolaron:
## What

Folds the tiny **Strategy** domain into **Blueprints** and deletes it. The Strategy domain held exactly one page — the blueprints boards overview at `/strategy/showBoards` — and its service already delegated to the Blueprints service. There's no other use for a separate `strategy` module in core.

This also fixes the left-nav highlight issue: the **Blueprints** menu item now stays active no matter which blueprint board is open (it was only active on `/strategy/showBoards`).

## Changes

- **Controller:** new native `Blueprints\Controllers\ShowBoards` (plain class, constructor DI, `get(): Response`) matching the recently-converted Blueprints controllers; renders `blueprints.showBoards`.
- **Service:** the 4 methods from `Strategy\Services\Strategy` folded into `Blueprints\Services\Blueprints` (`getStrategyBoardsOverview` → `getBoardsOverview`); the former cross-domain calls become self-calls (`getLastUpdatedCanvas`/`getBoardProgress`), dropping the cross-domain dependency.
- **Template:** moved to `Blueprints/Templates/showBoards.blade.php` (no body edits — it already used `/blueprints/{slug}/showCanvas` links).
- **Route:** `Route::get('/blueprints/showBoards', …)` declared before the `{canvasSlug}` group, plus a **301 redirect** from the old `/strategy/showBoards` (for bookmarks and the not-yet-updated plugins).
- **Rewired all core references:** menu item, help modal key (`strategy.showBoards` → `blueprints.showBoards`), project checklist link, and the three `DelCanvas` redirects (Blueprints / Canvas / Goalcanvas).
- **Deleted** `app/Domain/Strategy` entirely.

## Menu highlight

The left-nav **Blueprints** item now uses `module => 'blueprints'` with **no `active` key**, so it highlights on every `/blueprints/*` URL — the same idiom the wiki/files items already use. The default-menu **Retros** quick-link was removed (it sat in the same submenu and would otherwise double-highlight on retros pages); Retros remains in the full menu and is reachable from the boards overview.

## Verification

- Route binding + DI confirmed via a boot harness: `GET /blueprints/showBoards → ShowBoards@get` (name `blueprints.showBoards`) and the `/strategy/showBoards` 301 redirect both register.
- `php -l`, phpstan, and pint clean on all touched files.
- Adversarial review (3 lenses: missed references, service/controller wiring, menu+routing) — no dangling references, wiring sound; the only finding was the retros double-highlight, resolved by dropping the quick-link.

## Out of scope (private submodule)

`StrategyPro` and `PgmPro` reference `/strategy/showBoards`; those live in the `app/Plugins` submodule and must be updated there separately. The 301 redirect keeps them working in the meantime.

> Suggest a quick manual smoke test: open `/blueprints/showBoards`, open a canvas, delete a board (confirm the redirect), and check the left-nav Blueprints item highlights across blueprint pages.


marcelfolaron:
Addressed the failing `unit` job and all three Copilot comments:

- **Failing tests / stale Strategy test:** ported the old `StrategyServiceTest` coverage into `BlueprintsServiceTest` (board metadata, `buildRecentProgressCanvas`, and `getBoardsOverview` — the latter via a partial mock of the now self-called `getLastUpdatedCanvas`/`getBoardProgress`) and deleted the `tests/.../Strategy` directory. All 10 `BlueprintsServiceTest` cases pass locally.
- **`lastlastUpdate` typo:** renamed the board-metadata placeholder to `lastUpdate` (the key the rest of the code and the template actually read) across all 15 entries.
- **Redirect query string:** the `/strategy/showBoards` legacy redirect now preserves the query string, matching the other legacy redirects in `routes.php`.

phpstan + pint clean.

PR #3436, opened 2026-05-30
feature: RPC-only service methods throw typed exceptions on denial

marcelfolaron:
## What

The end-to-end proof of the response/exception foundation (#3431). The JSON-RPC endpoint has no controller-level role gate, so these `@api` methods self-authorize. They previously signalled denial with a **sentinel return** (`false` / `[]`), which the RPC layer serialized as a **successful** `{result}` at HTTP 200 — a denied action was indistinguishable from a successful one, and a denied tag lookup was indistinguishable from "no matching tags".

Now they **throw typed exceptions**, which the JSON-RPC controller already maps to precise error codes.

## Methods converted

Only the **denial / not-found** paths throw; the success-path delegate return is left intact (smallest behavior delta, no new exception class needed):

| Method | Was | Now |
|---|---|---|
| `Projects::patchProjectStatusAndSorting` | `return false` on no-manage | `throw AuthorizationException` |
| `Projects::sortProjects` | `false` on not-found / no-manage | `NotFoundException` / `AuthorizationException` |
| `Projects::patchProject` | `false` on no-manage | `AuthorizationException` |
| `Tags::getTags` | `[]` on no-access | `AuthorizationException` (a genuine no-match still returns `[]`) |
| `Tickets::patchTicket` | `false` on role / not-found / not-assigned | `AuthorizationException` / `NotFoundException` |
| `Tickets::sortTickets` | `false` on role / not-found / not-assigned | `AuthorizationException` / `NotFoundException` |

`AuthorizationException` → 403 / `-32001`; `NotFoundException` → 404 / `-32002`.

`sortTickets` was added after review flagged it as the RPC-only twin of `sortProjects` (same shape, same `leantime.rpc(...)`-only caller) — leaving it would have been inconsistent.

## Safety

- These methods are **RPC/JS-only** — verified no PHP caller branches on the old `false`/`[]` return (the same-named *repository* methods, e.g. `ticketRepository->patchTicket`, are different methods and are untouched).
- The internal `patch()` methods in both Projects and Tickets are **NOT** converted — they have internal PHP callers (and a plugin caller) that branch on the bool return.

## JS

Every caller already had a terminal `.catch` (so the conversion introduces no false-success and no unhandled rejections — the success `.then()` is simply skipped on rejection). Most only logged to console; upgraded them to a user-facing `jQuery.growl` of the server's error message:
- `ticketsRepository.handlePatchError` (covers the milestone/hours/date edits)
- the kanban dropdown handlers (effort / priority / status / milestone / user / sprint / async field)
- the Gantt project & ticket sort + date callers
- the calendar drag/resize handlers — which now also `event.revert()` the optimistic move, matching the existing event-type branch

The tag autocomplete is intentionally left to degrade to an empty list (correct UX for a typeahead).

> JS source changes require a `npx mix` rebuild to take effect; the inline calendar blade scripts are live immediately.

## Tests

Updated the Tags / Tickets / Projects service tests (the security-remediation regression guards) to assert the **thrown exception** instead of the old `false`/`[]` return. All pass: Tags 2/2, Tickets 16/16, Projects 18/18. phpstan + pint clean.

## Review

Adversarial review (3 lenses: PHP correctness, JS correctness, design consistency). The one confirmed finding — the missed `sortTickets` twin — is included in this PR. A refuted finding about a fallback i18n key was dismissed (the error message is always present on RPC rejections).


marcelfolaron:
Addressed the review. The 14 comments came down to two root causes:

- **Missing `short_notifications.not_saved` i18n key** (flagged on 6 files): the error growls fall back to this key only when a rejection has no message. The key didn't exist (it was already referenced by the pre-existing swimlane handler). Added it to `en-US.ini` in the `short_notifications` group — that fixes all six call sites at the source. In practice the fallback is rarely hit, since JSON-RPC rejections carry the exception's message, which is what the growl shows first.
- **Inconsistent indentation** in `showMyCalendar.blade.php`: re-indented the two ticket `.catch` blocks the perl insertion left under-indented.

phpstan + pint clean; unit was already green.

PR #3440, opened 2026-05-30
ci(phpstan): raise static analysis to level 1

marcelfolaron:
## What & why

Static analysis runs at the lowest possible setting (level 0). This is the first step of climbing the levels gradually: **bump `.phpstan/phpstan.neon` from level 0 → 1** and clear everything the new level surfaces in core/domain code, leaving a clean baseline to build on.

Level 1 initially reported 1,195 findings. The vast majority were non-issues; this PR resolves them in three ways — config (template noise), a stub (false positives), and real code fixes.

## Approach

**Stub for vanilla-PHPStan blind spots** — `.phpstan/stubs/leantime-macros.stub` (new), referenced via `stubFiles`. We don't run Larastan, so PHPStan can't see runtime-registered methods. The stub declares them with `@method`:
- `Str::mixin` macros from `LoadMacrosServiceProvider`: `toMarkdown`, `sanitizeForLLM`, `sanitizeFilename`, `beautifyFilename`, `alphaNumeric`
- `Collection::countNested`
- `Event::discoverListeners` (custom dispatcher bound to the `events` singleton)

Verified the `@method` tags **merge** onto the real classes — native `Str::` calls still resolve, no regression.

**Template variable-scope ignore** — legacy `.tpl.php` and raw-`<?php` blocks in blade receive their variables from the view layer (`$tpl->assign`/extract), which PHPStan can't see. Handled with a single identifier-scoped `ignoreErrors` block (`variable.undefined` under `*/Templates/*`, `app/Views/*`), `reportUnmatched: false` since templates are mid-migration.

## Genuine bugs fixed while clearing the level

| File | Bug |
|---|---|
| `Menu/Repositories/Menu.php` | "always-open" submenus had `visual` clobbered with an **undefined** `$submenuState` |
| `Projects::duplicateProject` | Duplicating without a start date used an undefined `$startDate` (no-op `try/catch` masked it) |
| `Plugins::validLicense` / `deactivate` | On a marketplace network error the `catch` logged but didn't bail → method call on undefined `$response`. Now bails safely |
| `DelTicket` / `DelMilestone` | Deleting without an `id` param hit an undefined `$id` (500). Now defaults to `0` and degrades gracefully |
| `MigrateCommand` | Silent console install used an undefined `$adminPassword` |
| `Help/Services/Helper.php` | `hourBudget` referenced `$values` while the array was being built |

Plus: removed the Laravel-11-dropped `Composer` argument from 7 console command constructors (`CacheTableCommand`, `*TableCommand`, `SeederMakeCommand`, …), and removed redundant `??`/`isset`/`empty` checks and dead code.

## Scope note

`app/Plugins/` is a private submodule **not checked out in CI**, so plugin-only findings (~36, across Copilot/Llamadorian/etc.) are out of scope here and tracked as a separate effort in the plugins repo.

## Verification

- `make phpstan` → core/domain clean.
- Analysed with `app/Plugins` excluded (mirrors CI, which has no plugins submodule) → `[OK] No errors`.
- Laravel Pint clean on all touched files.
- Console boots; all 7 modified commands instantiate via `help <cmd>`.

### DIFF
--- a/.phpstan/phpstan.neon
+++ b/.phpstan/phpstan.neon
@@ -1,7 +1,7 @@
 parameters:
     bootstrapFiles:
         - bootstrap.php
-    level: 0
+    level: 1
     inferPrivatePropertyTypeFromConstructor: true
     paths:
         - ../public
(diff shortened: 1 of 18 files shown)


marcelfolaron:
Thanks for the review — both addressed in 5608484633:

**1. `DelMilestone::get()` missing-id path.** Good catch. Replaced the `?? 0` fallback with an early-return guard so a missing `id` no longer renders a dialog over `getTicket(0)`. While doing so I noticed the error template was the real problem: **`errors.error400` has no view file** (it's referenced in a few places but only `error403`/`error404` exist), so that "graceful" path would itself have failed. Switched to the existing `errors.error404` (the convention already used by `DelTicket::get()`, `MoveTicket`, etc.), and fixed the pre-existing `error400` reference in `DelMilestone::post()` too. Applied the same guard to **`DelTicket::post()`**, which had the identical latent issue (`delete(0)` → `$result['msg']` on a non-array).

**2. `sanitizeForLLM` stub over-promise.** Agreed the `mixed → string` declaration was dishonest. Rather than loosen the return to `mixed` (which would lose the useful string guarantee for the common case), I **tightened the parameter to `string`** — matching the macro's documented purpose ("Sanitizes string for…"). String-in → string-out is now accurate, and any caller passing a non-string gets flagged. Core callers (`ToMarkdown`) already `(string)`-cast before calling, so no breakage.

PR #3445, opened 2026-05-31
fix(tickets): enforce editor role on save + honor milestone project change

### LINKED ISSUE #3294: Bug - Changing Milestone owning Project does not do anything.
### What is your set up?

Self Hosted Docker

### Version

3.6.0

### Describe the issue

When trying to change owning project for a milestone, clicking save does nothing, it just reverts to the old owning project.

Image

### Reproduction steps

1. Create a milestone in a project.
2. Change it's project to a second project.
3. Save
4. The milestone remains in the original project.

### Error Logs (LEANTIMEFOLDER/storage/logs)

_No response_

### LINKED ISSUE #3376: Enter a title of your bug
### What is your set up?

Self Hosted Docker

### Version

:latest | 3.7.3

### Describe the issue

If i add an additional user to my project. And give him the role of "commenter".

This user (logged in in another browser) cant edit or comment existing todos, if the todos are opened from Kanban or Table view.

but if the todo is opened from List view the user can change everything in the todo (like an editor)

Both users are handled by and authentik OICD

### Reproduction steps

1. Add an additional user as role "commenter" to project
2. Try to change or comment an Todo
...

### Error Logs (LEANTIMEFOLDER/storage/logs)

_No response_

marcelfolaron:
## What
Server-side authorization + milestone fixes in the Tickets domain.

- **#3376** — `Tickets::updateTicket()` now requires editor+ before saving. Editing was only gated in the template, so the Kanban/Table modal posted straight to `updateTicket` and let **commenter/reader** roles edit tickets via a direct request. This mirrors the guards already on `patchTicket()` and `updateTicketStatusAndSorting()`, so it also protects the JSON-RPC surface.
- **#3294** — `quickUpdateMilestone()` hard-coded `projectId` to the active project and discarded the value posted by the milestone dialog's `<select>`, so changing a milestone's owning project silently reverted.

## Test
- As a **commenter**, open a ticket from Kanban and try to save → expect the "not allowed" notification; editor still saves.
- Edit a milestone, change its **project**, save → it moves to the new project.

Fixes #3376
Fixes #3294

### DIFF
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -2115,13 +2115,18 @@ class Tickets
             return ['msg' => 'notifications.ticket_save_error_no_access', 'type' => 'error'];
         }
 
+        // Authorize against the ticket's CURRENT project: getTicket() returns false
+        // unless the user is assigned to that project. This stops an editor in one
+        // project from submitting another project's ticket id (alongside their own
+        // projectId) to edit or move it — the later check only validates the
+        // caller-supplied projectId. (#3376 review)
+        $currentTicket = $this->getTicket($values['id']);
+
+        if (! $currentTicket) {
+            return ['msg' => 'notifications.ticket_save_error_no_access', 'type' => 'error'];
+        }
+
         if (! isset($values['headline'])) {
-            $currentTicket = $this->getTicket($values['id']);
-
-            if (! $currentTicket) {
-                return ['msg' => 'This ticket id does not exist within your leantime account.', 'type' => 'error'];
-            }
-
             $values['headline'] = $currentTicket->headline;
         }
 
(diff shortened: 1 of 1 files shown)


marcelfolaron:
Thanks for the review. Addressed in the latest commit:
- **Cross-project IDOR / current-project authz:** `updateTicket()` now loads the ticket via `getTicket()` (which returns false unless the user is assigned to the ticket's *current* project) and bails otherwise — so a ticket id from another project can no longer be edited/moved by supplying your own `projectId`.
- **Milestone project change:** `quickUpdateMilestone()` now validates the user is assigned to the target project before honoring the dialog-supplied `projectId`, and when the project actually changes it routes through `moveTicket()` so the milestone's child tasks move with it instead of being orphaned.
- The PR description was also corrected (it had been swapped with another PR's body).

marcelfolaron:
Good catch — the second-round review flagged that the checks were still keyed to the **session** project's role. Leantime roles are project-scoped, so that's a real gap. Addressed in the latest commit:

- **`updateTicket` (role keyed to session, not the ticket's project):** the ticket is now loaded first via `getTicket()` (the project-assignment gate), then editor+ is required **in the ticket's own project** via a new `userIsAtLeastForProject()` helper. An editor in project A who is only a commenter in project B can no longer keep the session on A, post B's ticket id, and edit B's ticket.
- **`quickUpdateMilestone` loading via the repo (no access gate):** now loads the milestone through the service `getTicket()` and fails when it returns `false`. It also requires editor+ in the milestone's **own** project to edit, and editor+ in the **target** project to move it there (previously only an assignment check).
- **Truthy error arrays read as success (`@2510`/`@2517`):** `updateMilestoneFromDialog()` now only treats `=== true` as success and returns `false` for any error/failure path, so the controller surfaces the save-error notification instead of a false success.
- Added `userIsAtLeastForProject(role, projectId)`: resolves the effective project-scoped role (manager/admin/owner keep their global role everywhere; otherwise the project role applies, falling back to the global role when none is set).

phpstan + pint clean.

PR #3450, opened 2026-05-31
fix(comments): open the edit/reply box above the replies thread

### LINKED ISSUE #3319: Comment jumps down (under its replies) when attempting to edit
### What is your set up?

Self Hosted Server

### Version

3.7.3

### Describe the issue

After upgrading to version 3.7.3 when I want to edit a comment which has replies, upon clicking "Edit" button the comment jumps to the bottom (under its replies). Also, it shows several fields in which I can type. See screenshots.

Normal view (correct order):

Image

Edit view (messed up order):
Comment jumped to the bottom. At the bottom there are 3 input fields now.

Image

Additionally, also note that I can't edit the comment replies (because there is no edit button) which I believe has been possible before.

### Reproduction steps

1. Upgrade to LeanTime version 3.7.3.
2. Open a ticket which has a comment with at least one reply.
3. Edit the comment and observe its position change.
...

### Error Logs (LEANTIMEFOLDER/storage/logs)

None.

marcelfolaron:
## What
**#3319** — Editing a comment that has replies made it "jump" below its replies. The edit/reply box lived **inside** `.replies`, after the replies loop, so showing it for an edit rendered it under all the replies. Moved the box out of `.replies` to sit directly under the parent comment, so editing opens in place.

## Test
On a ticket with a comment that has at least one reply, click Edit on the parent comment → the editor opens in place, above the replies, not at the bottom.

Fixes #3319

### DIFF
--- a/app/Domain/Comments/Js/commentsController.js
+++ b/app/Domain/Comments/Js/commentsController.js
@@ -5,11 +5,13 @@ leantime.commentsController = (function () {
         // Show the "Add new comment" toggler that makeInputReadonly may have hidden
         jQuery("[class^='mainToggler']").show();
 
-        // Show reply-level comment boxes (legacy .commentBox class only).
-        // Do NOT show commentBox-{hash} containers — those are the "new comment"
-        // forms that start hidden and open on-demand via toggleCommentBoxes().
-        jQuery(".commentBox").show();
-        jQuery(".replies .commentBox").hide();
+        // Keep per-comment reply/edit boxes (legacy .commentBox class) hidden;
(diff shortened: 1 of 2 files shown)


marcelfolaron:
Thanks — addressed in the latest commit. Moving the box out of `.replies` did mean `enableCommenterForms()` (which showed `.commentBox` and only re-hid `.replies .commentBox`) would reveal it. It now hides by the `.commentBox` class directly, so the boxes stay closed until opened via `toggleCommentBoxes()`; the new-comment form uses `commentBox-{hash}` and is unaffected. The PR description was also corrected.

PR #3451, opened 2026-05-31
fix(wiki): save notes from the All Notes grid into a default notebook

### LINKED ISSUE #3216: Notes: Cannot save notes created from All Notes grid view
## Description
Notes created from the All Notes grid view don't save because they aren't assigned to a notebook.

## Steps to Reproduce
1. Go to Notes → All Notes
2. Click to create a new note
3. Enter content and click Save
4. Note is not saved

## Expected Behavior
Notes should either:
- Be assigned to a default notebook, OR
- Prompt the user to select a notebook before saving

## Context
Discovered during Tiptap editor migration (Phase 1).

marcelfolaron:
## What
**#3216** — Notes created from the **All Notes** grid didn't save: there was no active notebook in session, so `canvasId` was empty and the note saved into nothing. Fall back to the project's default notebook (`getAllProjectWikis` auto-creates one when none exist).

## Test
From All Notes → create a new note → save → it persists and appears under the default notebook.

Fixes #3216

### DIFF
--- a/app/Domain/Notifications/Services/News.php
+++ b/app/Domain/Notifications/Services/News.php
@@ -114,6 +114,9 @@ class News
         $client = new \GuzzleHttp\Client;
         $response = $client->request('GET', 'https://leantime.io/category/leantime-updates/feature-updates/feed/', [
             'headers' => ['Accept' => 'application/xml'],
+            // Fail fast when the server (or CI runner) has no egress so the news
+            // badge/widget degrades quickly instead of stalling. (#3372/#3373)
+            'connect_timeout' => 2,
             'timeout' => 5,
         ])->getBody()->getContents();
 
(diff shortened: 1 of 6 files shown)


marcelfolaron:
Good catches — all three addressed in the latest commit:
- **News feed timeout reverted / Telemetry timeout reverted:** these were collateral from the earlier bad master merge (which resolved conflicts as *ours*). My first cleanup only restored `app.js`/`Install.php`; I've now also restored `News.php` and `Reports.php` to master, so the fail-fast `connect_timeout`/`timeout` limits are back and dashboard badge/widget rendering won't block on an unreachable endpoint.
- **Empty canvasId fallback (#3216):** when neither the session notebook nor a default project notebook resolves, it now sets an error notification and redirects back instead of creating an orphaned note with an empty `canvasId`.

PR #3454, opened 2026-05-31
feat(core): WorkStructure meta-model infrastructure

marcelfolaron:
## Summary

Adds **WorkStructure** — a Core infrastructure concern that describes how Leantime entities compose and how they map across structures. It is the orchestration "brain" that a Logic Model board (PR 2) and future orchestrators (Program, Strategy) build on.

This is **PR 1 of 2** porting [#3275](https://github.com/Leantime/leantime/pull/3275) (which targeted the abandoned `feature/phase-0-modernization` branch) onto `master`. PR 2 adds the Logic Model canvas board and builds on this.

### Why Core, not a Domain?
WorkStructure has no UI — just models, a repository, services, and events. It *governs* domains rather than being one (a Project is made of milestones/tasks/goals; a task `belongs_to` a milestone; a Logic Model output `maps_to` a Project milestone). That's the same shape as `Core/Plugins` and `Core/Auth`: infra in Core, registered via a ServiceProvider. (Its instance-level twin is the existing `Entityrelations`.)

### What's included
- `app/Core/WorkStructure/` — `Models/`, `Repositories/WorkStructureRepository` (Query Builder over 4 tables), `Services/` (`StructureRegistry`, `MappingService`, `WorkStructureService` — the `@api` surface plugins register against), and lifecycle `Events/`.
- `WorkStructureServiceProvider` — binds the services and seeds the built-in **"Project"** structure in `boot()` (idempotent, cache-guarded, degrades gracefully before tables exist). Registered in `laravelConfig.php` after `DatabaseServiceProvider`. **Replaces** the PR's `register.php` event-hook seeding, which targeted an event that no longer exists on master.
- **Schema** — `SchemaBuilder::createWorkStructureTables()` (fresh installs) + `Install::update_sql_30502()` and `dbVersion` `3.5.1 → 3.5.2` (existing installs): `zp_work_structures`, `_elements`, `_relationships`, `_mappings`.
- `EntityRelationshipEnum` gains `GeneratedFrom` and `MapsTo` for the instance-level provenance links the Logic Model wizard will write into `zp_entity_relationship`.

### Notes
- The wizard / entity-generation / health / fixtures / export remain in the private **StrategyPro** plugin, which registers its "Logic Model" structure and mappings through `StructureRegistry`. The plugin must reference the new `Leantime\Core\WorkStructure\…` namespace.

## Test plan
- [ ] Fresh install creates the 4 `zp_work_structure*` tables.
- [ ] Existing-install upgrade runs `update_sql_30502`; `db-version` advances to `30502`.
- [ ] After boot, `app(\Leantime\Core\WorkStructure\Services\StructureRegistry::class)->get('Project')` returns the seeded structure with 3 elements + 2 relationships; booting twice does not duplicate.
- [x] `php -l` clean, Laravel Pint clean, DB-free autoload/API smoke test passes.
- [ ] `make phpstan` (Docker) clean.

### DIFF
--- a/app/Core/Configuration/AppSettings.php
+++ b/app/Core/Configuration/AppSettings.php
@@ -9,5 +9,5 @@ class AppSettings
 {
     public string $appVersion = '3.8.0';
 
-    public string $dbVersion = '3.5.1';
+    public string $dbVersion = '3.5.2';
 }
--- a/app/Core/Configuration/laravelConfig.php
+++ b/app/Core/Configuration/laravelConfig.php
@@ -40,6 +40,7 @@ return [
             \Leantime\Core\Auth\AuthenticationServiceProvider::class,
             \Leantime\Core\Middleware\RateLimiter::class,
             \Leantime\Core\Database\DatabaseServiceProvider::class,
+            \Leantime\Core\WorkStructure\WorkStructureServiceProvider::class,
             \Leantime\Core\i18n\LanguageServiceProvider::class,
             // \Leantime\Core\Providers\RouteServiceProvider::class,
 
--- a/app/Core/Support/EntityRelationshipEnum.php
+++ b/app/Core/Support/EntityRelationshipEnum.php
@@ -15,6 +15,12 @@ enum EntityRelationshipEnum: string
     case Collaborator = 'collaborator';
 
     /**
-     * Add other relationship types as needed.
+     * Represents a "generated from" relationship (entity created from a canvas item).
      */
+    case GeneratedFrom = 'generated_from';
+
+    /**
+     * Represents a "maps to" relationship (cross-structure element mapping).
+     */
+    case MapsTo = 'maps_to';
 }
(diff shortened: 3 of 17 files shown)


marcelfolaron:
Thanks for the review — addressed in the latest commit:

- **Duplicate mappings / missing unique constraint** (`zp_work_structure_mappings`): added a unique index on `(source_structure_id, source_element_id, target_structure_id)` in **both** the fresh-install schema (`SchemaBuilder`) and the `update_sql_30502` path. This is exactly the idempotency key `StructureRegistry::registerMappings()` checks, so a race can no longer insert a duplicate. (`zp_work_structures.title` and `zp_work_structure_elements (structure_id, type_key)` already had unique indexes.)
- **`has()`/`get()` throwing before tables exist:** `StructureRegistry` now guards the boot-time path with a per-request `schemaReady()` check (`Schema::hasTable`, cached for the request). During a fresh install or a pending update, `register()`/`registerMappings()` no-op and `has()`/`get()` return `false`/`null` instead of crashing boot; the plugin re-registers idempotently on the next request once the migration has run.
- **`register()` null deref:** if the exists-cache drifts from the row, it now clears the cache and recreates instead of dereferencing a null `$structure->id`.
- **N+1 in `getTargetElementKey()`/`getSourceElementKey()`:** the element id→typeKey map is now built once (mirroring `getMappedElementKeys()`) instead of re-querying all elements inside the mappings loop.

PR #3456, opened 2026-06-01
fix(install): register update_sql_30503 + bump dbVersion to 3.5.3

gloriafolaron:
## Summary

Two-line follow-up to #3401 (mobile push device tokens) — without these the migration never actually fires and the new RPCs throw "Table doesn't exist".

- Append `30503` to `Install\Repositories\Install::$dbUpdates` so the migration runner picks up the `update_sql_30503` method that #3401 added.
- Bump `AppSettings::$dbVersion` from `3.5.2` → `3.5.3` so `Install\Services\Install::needsUpdate()` returns true on instances that already ran the 3.5.2 migrations (otherwise the update flow never starts).

## How I found it

Verifying #3401 end-to-end against a local Herd instance pointed at the cloud schema. After pulling master + running the dashboard, `registerPushToken` returned `-32000 Server error`. Tail of `storage/logs/leantime-*.log`:

```
PDOException: SQLSTATE[42S02]: Base table or view not found:
  1146 Table 'leantime_accounts_test.zp_device_tokens' doesn't exist
  in .../Database/Connection.php:596
```

The `update_sql_30503` method is on the Repository class (added in #3401) — it just never gets called because `$dbUpdates` stops at `30502`.

## Verification

After applying the two-line patch:
```
runUpdate() → true
registerPushToken({token, platform: 'ios', provider: 'fcm'}) → {result: true}
unregisterPushToken({token})                                   → {result: true}
getUnreadCount()                                                → {result: 0}
```

Row appears in `zp_device_tokens` with `provider='fcm'`, `lastSeenAt` set; soft-delete via `invalidatedAt` works as expected.

## cc

@marcelfolaron — this finishes wiring #3401 so the RPCs work without manual schema intervention.


gloriafolaron:
Superseded by #3457 — that PR implements the storage-on-access-tokens design Marcel preferred + the channel impl. Closing this one.

PR #3457, opened 2026-06-01
feat(notifications): mobile push tokens on access_tokens + FCM/Expo dispatcher

gloriafolaron:
## Summary

Combined follow-up to #3401 — implements the storage option you flagged ("piggyback on `zp_access_tokens`, lifecycle tied to the token and login") AND adds the channel implementation so pushes actually get delivered. Supersedes #3456 (which was an interim fix for the wrong-design storage A).

### 1. Storage refactor (option B)

Replaces the never-fired `zp_device_tokens` migration from #3401 with `update_sql_30504`, which adds push columns directly to `zp_access_tokens`:

| column | type | meaning |
|---|---|---|
| `push_token` | VARCHAR(255) NULL | opaque Expo/FCM token |
| `push_platform` | VARCHAR(16) NULL | `'ios'` or `'android'` |
| `push_provider` | VARCHAR(8) NULL | `'expo'` or `'fcm'` |
| `push_token_updated_at` | TIMESTAMP NULL | refreshed each `registerPushToken` |
| `push_invalidated_at` | TIMESTAMP NULL | set by mobile unregister OR provider DeviceNotRegistered |

Plus an index `(tokenable_id, push_invalidated_at)` for the dispatcher's "active tokens for these users" query.

**Lifecycle wins:**
- Logout / Sanctum token revoke deletes the access_tokens row → push registration dies with it. No orphan rows, no prune cron.
- Multi-device works naturally: each device has its own zp_access_tokens row (per `device_name` from login), each carrying its own push token + provider.

The original `update_sql_30503` method and `DeviceTokens` repository are removed entirely. Installs that already ran 30503 (none in production since the migration never fired without #3456's dbVersion bump) have an unused `zp_device_tokens` table sitting around — safe to drop in a future cleanup PR; not dropped here to keep this migration purely additive.

`dbVersion` bumped to `3.5.4` so the new migration fires.

### 2. Validation simplifications

- Dropped Expo prefix check (`ExponentPushToken[...]`)
- Dropped FCM length bounds
- Mobile passes `provider` explicitly; provider rejects invalid tokens at send-time and we mark `push_invalidated_at` then. Pre-validating was buying us nothing.

### 3. Channel impl — \`Notifications\Services\Push\`

Sibling to the existing `Messengers` dispatcher. Sends pushes via:
- **FCM HTTP v1** with service-account credentials. JWT signing happens inline with PHP's openssl — no new composer deps. The OAuth access token is cached ~59 minutes.
- **Expo Push** via the public exp.host endpoint (no server auth).

Per-token failure doesn't break the broadcast. UNREGISTERED / DeviceNotRegistered / INVALID_ARGUMENT → set `push_invalidated_at`.

Wired into `Projects\Services\Projects::sendNotificationToTeam()` right after the existing Messengers call. Wrapped in try/catch so push outages don't block the queued-email + messenger paths.

### 4. Configuration

Two new env vars for FCM:
- `LEAN_PUSH_FCM_CREDENTIALS_PATH` — absolute path to the Firebase service-account JSON
- `LEAN_PUSH_FCM_PROJECT_ID` — Firebase project id

Unset → FCM sends silently no-op. Expo sends always work.

## Verification

End-to-end against a local instance after running the migration:

\`\`\`
registerPushToken({token, platform: 'ios', provider: 'fcm'}) → {result: true}
unregisterPushToken({token})                                  → {result: true}
getUnreadCount()                                              → {result: 0}
\`\`\`

Row appears in `zp_access_tokens` with `push_provider='fcm'`, `push_token_updated_at` set, `push_invalidated_at` null. Unregister flips `push_invalidated_at`. The dispatcher's query for active tokens correctly skips invalidated rows.

## Mobile side

Mobile (\`Leantime/leantime-mobile@main\`) already calls these exact method names with the FCM provider, so no mobile-side changes needed for this PR to start working.

## cc

@marcelfolaron — this combines #3456 + the channel impl per your direction. Once this lands and the env vars are set on cloud, mobile push notifications deliver end-to-end. #3456 can be closed.


gloriafolaron:
Following up on the Copilot review that appeared on the (now-closed) #3458 — the push-related comments belong here. Three points to confirm:

**1. `Notifications` constructor change (Copilot @ Notifications.php:42)**

Grepped for manual construction of the service:

```
$ grep -rn "new Notifications(" app/ tests/
tests/Unit/app/Domain/Notifications/Services/NotificationsServiceTest.php:27
tests/Unit/app/Domain/Notifications/Services/NotificationsServiceTest.php:100
tests/Unit/app/Domain/Notifications/Services/NotificationsServiceTest.php:124
```

Only the test class constructs it directly, and all three call sites were updated in this PR. No production code constructs `Notifications` manually — everything goes through DI auto-resolution, which picks up the new signature transparently.

**2. Migration 30503 → 30504 gap (Copilot @ Install.php:87)**

Confirmed the runner tolerates the gap. The loop at `app/Domain/Install/Repositories/Install.php:322-326` iterates `$this->dbUpdates` and calls `update_sql_$version` only when `currentDBVersion < $updateVersion`. Because 30503 is no longer in the array, `update_sql_30503` simply never runs — no fallthrough, no error. Anyone whose `db-version` is below 30504 (which is everyone, since PR #3401's 30503 never shipped) gets `update_sql_30504` and moves on.

Orphan `zp_device_tokens` table risk: only exists for anyone who ran the PR #3401 branch locally before it was revised — that population is essentially Gloria's dev machine plus maybe Marcel's. Not in the field. Happy to add a defensive `DROP TABLE IF EXISTS zp_device_tokens` to `update_sql_30504` if you want belt-and-suspenders cleanup.

**3. `$users` shape contract (Copilot @ Projects.php:320)**

Defensive observation, not an active bug — `$users` at this point is the post-filter scalar ID array (filtered at line 269, re-keyed by `array_values()`). If a future refactor changes the shape, `Push::send()`'s `whereIn('tokenable_id', $userIds)` silently matches nothing rather than crashing. The contract is implicit; we could add a `// expects: array<int>` doc or a debug assertion if you want it made explicit, but happy to leave it as-is since the assumption matches the rest of the channel dispatchers in `Messengers`.

**Separately addressed in commit 2d08cf7:** the three real bugs Copilot caught in `Push::sendFromNotification()` (`module`/`moduleId`/`url` extraction) — fix pushed, 26/26 notification unit tests still pass.

PR #3458, opened 2026-06-01
fix(comments): stop shadowing controller $comments array in Discussion template

gloriafolaron:
## Summary

- The Discussion submodule (`generalComment.blade.php`) reassigned `$comments` to the Comments **Repository** at the top of the template so it could call `$comments->getReplies($id)`. That overwrote the array of comments the controller had already assigned to the same name (`$this->tpl->assign('comments', $comments)`), so the `@foreach ($comments as $row)` further down iterated the repository object — which yields nothing — and every Discussion section came up empty.
- Renames the local repository handle to `$commentsRepo` so the controller-provided `$comments` array is preserved for the render loop, and updates the two `getReplies()` call sites accordingly.

## How this was found

Posting a comment via the mobile RPC client (`leantime.rpc.Comments.Comments.addComment`) returned success and the row was visible in `zp_comment` (e.g. comment id 325 against ticket 1702 with `module='ticket'`, `commentParent=0`), but the web Discussion section rendered empty. Reproduced on `projects.leantime.io` directly with comments created via the web form too — same empty state — confirming it wasn't an RPC-only issue.

## Test plan

- [x] `./vendor/bin/phpunit tests/Unit/app/Domain/Comments/Services/CommentsServiceTest.php` → 6/6 passing
- [ ] Open any ticket with existing comments → Discussion section renders the list
- [ ] Reply to a comment → reply renders nested under the parent (verifies `getReplies()` calls still work via `$commentsRepo`)
- [ ] Delete a comment → existing delete flow still works


gloriafolaron:
Closing in favor of a clean PR. This branch accidentally inherited commits from the push-tokens branch (PR #3457) when I forked it — the diff shows the full push refactor as well as the one-line template fix. Re-opening with just the template change on a fresh branch off master. Thanks @copilot-pull-request-reviewer for catching the scope drift.

PR #3460, opened 2026-06-02
feat(api): plugin gate attribute + capability discovery endpoint

gloriafolaron:
## Summary

Three pieces shipped together to make JSON-RPC plugin gating clean for mobile, MCP, and web clients. Nothing here is consumed by anything yet — the surface is built first so Phase 2 (Logic Model UnifiedTools, plugins repo PR #41) can light up cleanly.

### 1. `#[RequiresPlugin('Name')]` attribute (server-side defense)
- New attribute at `app/Core/Plugins/Attributes/RequiresPlugin.php`
- Tagged on JSON-RPC service methods or classes
- Enforced inline in `Jsonrpc.php` after the existing `@api` check
- When the named plugin is disabled, returns JSON-RPC error `-32004` in the response body with HTTP 200 (per JSON-RPC 2.0 convention — clients key off the body, not HTTP status)
- Method-level usage overrides class-level
- Not `IS_REPEATABLE` by design — PHP rejects multiple stacked attributes at the language level. Documented in the attribute docblock with a forward path if a future tool needs multi-plugin gating.
- **Gate source of truth:** uses `Domain\Plugins\Services\Plugins::isEnabled()` (DB-backed, same as `config.getSystemInfo` reports), NOT `Core\Plugins\Plugins` (env-only system plugins).

### 2. JSON-RPC dispatch fallback to plugin `Tools/` namespace
- Resolution order: `Domain\{Module}\Services\` → `Plugins\{Module}\Services\` → **new: `Plugins\{Module}\Tools\`**
- Lets plugin authors place `#[UnifiedTool]`-tagged classes in the same `Tools/` directory `McpToolDiscovery` already scans, without having to duplicate them under `Services/` to make them JSON-RPC reachable
- Existing dispatch paths are unchanged; only adds a fallback when neither earlier path resolves

### 3. `leantime.rpc.api.config.getSystemInfo` (client-side prevention)
- New core service at `app/Domain/Api/Services/Config.php`
- Returns `{ version: string, enabledPlugins: string[] }`
- Always available — does NOT depend on any plugin (unlike `/advancedAuth/mobileStatus` which lives in the AdvancedAuth plugin and isn't reachable on OSS-only installs)
- Mobile/MCP/web clients query once at login and cache until next login
- Staleness contract: max acceptable staleness = session length. Admin plugin toggles propagate on next session refresh.

### Why all three together

`RequiresPlugin` alone catches the "plugin disabled" case after the call. `getSystemInfo` alone has no enforcement — a client that ignores the capability list still reaches a denied call. The `Tools/` fallback alone is just plumbing. Together: clients gate UI client-side from `getSystemInfo`, server refuses anything that slips through with a clean `-32004`, and tools live in the directory MCP discovery already understands.

## What's NOT in this PR

- **Mobile-readiness audit fixes** (M1 array shape, M2 pagination, M3 discriminated returns) — moved to Phase 2 RPC tool boundary (plugins PR #41). The service-layer shape is what internal HTMX/web callers want; reshaping at the wire happens in the UnifiedTool wrappers, not the services.
- **Retroactive `#[RequiresPlugin]` tagging on existing plugin services** — those RPC surfaces are technically exposed (`@api` methods reachable via plugin namespace dispatch) but not actually called by any JS or client today. Tag lands as Phase 2 UnifiedTools are written.

## Test plan

- [x] **Verify existing endpoints still work** (no `RequiresPlugin` tag → unchanged behavior). Any `@api` method via curl with a valid session/API key — should respond exactly as before.
- [x] **Verify Config endpoint** returns version + enabled plugins:
  ```javascript
  // From an authenticated browser session:
  fetch('/api/jsonrpc', {
    method: 'POST',
    credentials: 'include',
    headers: { 'Content-Type': 'application/json' },
    body: JSON.stringify({
      jsonrpc: '2.0',
      method: 'leantime.rpc.api.config.getSystemInfo',
      id: 1
    })
  }).then(r => r.json())
  ```
  Expected: `{ jsonrpc: '2.0', result: { version: '3.8.0', enabledPlugins: ['AdvancedAuth', 'StrategyPro', ...] }, id: 1 }`. **Verified locally — response matches.**
- [x] **Verify gate fires** for a method tagged `#[RequiresPlugin('SomePlugin')]` when SomePlugin is disabled. **Verified with `StrategyPro` UnifiedTools (plugins PR #41) — clean -32004 with the expected message.**
- [x] **Verify gate is silent** when the named plugin IS enabled — dispatch falls through normally. **Verified — `StrategyPro` tools dispatch cleanly when StrategyPro is enabled.**
- [x] **Verify `Tools/` fallback resolves** — `leantime.rpc.strategypro.strategyTools.getLogicModelTemplates` returns the 5 fixture templates. **Verified locally.**
- [x] **Pint passes**: `php vendor/bin/pint --test app/Core/Plugins/Attributes/RequiresPlugin.php app/Domain/Api/Services/Config.php app/Domain/Api/Controllers/Jsonrpc.php` — green locally.

## Files changed

- `app/Core/Plugins/Attributes/RequiresPlugin.php` (new, 25 lines)
- `app/Domain/Api/Services/Config.php` (new, 50 lines)
- `app/Domain/Api/Controllers/Jsonrpc.php` (+50 lines: imports, gate check, getRequiredPlugin helper, Tools/ namespace fallback)

## Commit history

1. `feat(api): plugin gate attribute + capability discovery endpoint` — initial three files
2. `feat(api): JSON-RPC fallback to plugin Tools/ namespace` — third dispatch path
3. `fix(api): gate uses DB-backed Domain Plugins service` — source-of-truth alignment with `getSystemInfo`
4. `fix(api): address review feedback on gate` — remove unused PluginNotEnabledException, document non-repeatable attribute design


app/Domain/Api/Controllers/Jsonrpc.php
```diff
@@ -334,6 +361,35 @@ private function isApiMethod(string $serviceName, string $methodName): bool
...
+            $method = new ReflectionMethod($serviceName, $methodName);
+            $attrs = $method->getAttributes(RequiresPlugin::class);
+            if (! empty($attrs)) {
+                return $attrs[0]->newInstance()->pluginName;
+            }
+
+            $class = new ReflectionClass($serviceName);
+            $classAttrs = $class->getAttributes(RequiresPlugin::class);
+            if (! empty($classAttrs)) {
+                return $classAttrs[0]->newInstance()->pluginName;
```
marcelfolaron, line 384: Will this instantiation work for phar as well as regular plugins?

gloriafolaron: Yes, it should work for phar plugins too — same Reflection mechanism.

By the time `getRequiredPlugin()` runs at line 384, the dispatcher has already cleared the `class_exists($pluginToolNamespace)` check at line ~216, which means the autoloader successfully resolved the class regardless of source. `ReflectionMethod`/`ReflectionClass` then operates on the loaded class metadata, which is identical whether the class came from `app/Plugins/StrategyPro/Tools/StrategyTools.php` (folder) or `phar://app/Plugins/StrategyPro.phar/Tools/StrategyTools.php` (phar).

The symmetry holds because `McpToolDiscovery` already does the same Reflection-based attribute scan over phar paths in production — `getEnabledPluginPaths()` constructs `phar://{pluginPath}/{plugin}.phar` for phar-format plugins (Core/Plugins/Plugins.php:103), then `McpToolDiscovery::getToolClassesFromDirectory()` uses the same `getAttributes()` Reflection call this PR uses to find `#[UnifiedTool]` annotations. If that works for phar today (which it does, since McpServer ships marketplace tools that way), this works too.

Honest caveat: I haven't tested with an actual phar plugin locally — every plugin in my dev install is `format = folder`. If you have a phar plugin handy and want to confirm with a quick test, the easiest path is to install one with a `#[RequiresPlugin]`-tagged method and toggle the plugin off. But the mechanism is symmetric with existing phar-aware code, so I'd be surprised if there's a difference.

PR #3482, opened 2026-06-03
feat(libs): npm-manage html2canvas + jsPDF

gloriafolaron:
## Summary
Adds two general-purpose client-side libraries to the build pipeline instead of committing minified blobs:

- \`html2canvas@^1.4.1\` — DOM-to-canvas screenshot rendering
- \`jspdf@^4.2.1\` — client-side PDF generation

Both go through npm + Mix. Lazy-loaded at runtime on demand, so users who don't trigger an export pay zero cost.

## What changed (vs. the previous shape of this PR)
This PR originally just dropped two pre-minified files into \`public/assets/js/libs/\` — committed blobs Copilot couldn't review. Pivoted to:

- **package.json** + **package-lock.json**: \`html2canvas\` and \`jspdf\` as devDependencies.
- **webpack.mix.js**: two \`.copy()\` calls that emit the UMD bundles to \`public/dist/js/html2canvas.min.js\` and \`public/dist/js/jspdf.umd.min.js\` (unversioned, stable URLs).
- **public/assets/js/libs/**: deleted both blobs.

Total diff: 6 files, ~230 insertions, ~430 deletions (the 430 is the two blobs going away).

## Why
- **Security**: \`npm audit\` now covers these. CVEs in html2canvas / jsPDF show up in the existing NPM Security Audit CI job alongside everything else.
- **Maintenance**: \`npm update\` flow for patches instead of manually re-extracting from a shelved branch.
- **Build pipeline integration**: Mix manages them like every other lib in the codebase.
- **Reviewability**: 6-file diff with intent visible vs. 580KB of opaque binary.
- **Free bonus**: jsPDF 4.x has better CSS color support than the 3.x that was in the shelved \`feature/phase-0-modernization\` branch — likely fixes the \`color(srgb …)\` parse issue the StrategyPro export currently works around with 8-digit hex (#RRGGBBAA).

## Impact on OSS users
- **Disk**: +~560KB in \`public/dist/js/\` post-build (was previously +~560KB in \`public/assets/js/libs/\` as committed blobs). Net change: roughly zero.
- **Runtime**: zero. Lazy-loaded on demand.
- **Build**: Mix already runs on every build; this just adds two \`.copy()\` calls. \~0.2s additional build time.
- **UX**: unchanged. No new buttons in core; OSS users see the same Logic Model surface.
- **Future-proofing**: any community-built export feature can \`mix.copy()\` its own libs the same way.

## Coordinated change
Plugins repo PR Leantime/plugins#47 updates the StrategyPro Logic Model export loader paths from \`/assets/js/libs/…\` → \`/dist/js/…\`. The two land together; merging either alone will 404 exports until the other lands.

## Test plan
- [ ] After \`npm install && npx mix\` (or fresh CI build): \`public/dist/js/html2canvas.min.js\` and \`public/dist/js/jspdf.umd.min.js\` exist and are served as \`application/javascript\`.
- [ ] With the StrategyPro plugin installed + the coordinated plugin PR merged: PNG and PDF exports on the Logic Model board generate downloads (no console \"Failed to load\" errors).
- [ ] Pages that don't use export: no fetch of these libs (verify in DevTools Network tab).
- [ ] \`npm audit\` includes html2canvas + jspdf in its dependency graph.


marcelfolaron:
Are these libs not available as npm packages?

PR #3490, opened 2026-06-04
security(timesheets): gate the Timesheets domain (own-time vs manage) + matrix edit

marcelfolaron:
## What & why

Timesheets are company-wide time logging, and the controllers have always gated on the user's **global** role (`Auth::authOrRedirect([...], forceGlobalRoleCheck: true)`). The `@api` service surface had critical IDORs: `getTimesheet` read any entry by id with **zero auth**, `updateInvoices` flipped invoice/paid flags on **arbitrary** ids, and `getAll` / `pollFor*` / `getUsersTickets` / `getUsersTicketHours` leaked other users' data.

This is the rollout's **one matrix-edit domain** (a non-standard `timesheets.manage` verb).

## Approach

- **Vocabulary** — `TimesheetsPermissions` (`timesheets.view/create/edit/delete/manage`), **GLOBAL-scoped** (`projectScoped = false`), matching the controllers' global-role gating. `view/create/edit/delete` = editor+ for the user's **own** time; `manage` = manager+ for acting on **others'** time, invoicing, and cross-user reports.
- **Matrix edit** — editor gets the four global keys **explicitly** (global-scoped perms are *not* matched by the project verb rule — a subtlety the design initially missed); manager gets `timesheets.manage`; admin/owner auto-grant via `scope:any`. readonly/commenter get none. **Migration `30514` re-seeds** because the matrix changed (dbVersion → **3.5.14**).
- **Service** (`extends BaseService`) — ownership is enforced in-body: an entry's `userId` vs the current user decides own (`view`/`edit`/`delete`) vs `manage`. **Reads soft-deny** (`false`/`[]`/`0` — no cross-user existence oracle); **writes throw** / pin non-managers to their own `userId`. `getAll` is **dual-use**: own userId → `view`; another user / all-users report (`null`) → `manage`.
- **Intentionally ungated** — the ticket-hours aggregate reads (`getLoggedHoursForTicketByDate`/`getSumLoggedHoursForTicket`/`getRemainingHours` + the hour-type lists) render on the **ticket view** (`ShowTicket`, reachable by readonly) and expose only a ticket's *total* hours, not per-user data; `getUsersTicketHours` is gated only for the cross-user case (ShowTicket calls it with the current user's id).
- **Controllers + Stopwatch** — all 6 controllers + the Stopwatch HxController carry `#[RequiresPermission(…, global: true)]`, replacing `Auth::authOrRedirect` (AddTime/EditTime/ShowMy/ShowMyList editor+, ShowAll manager+). `AddTime` now scopes its picker to the current user; `EditTime` drops its manual owner check, relying on the gated `getTimesheetForEdit` (read) + `updateTime` (save).

## Tests / verification
- 12 fail-closed `TimesheetsService` cases (soft-deny / no-oracle, own-vs-manage, `getAll` dual-use, non-manager pinned to own); `timesheets.*` added to the role-matrix catalog test.
- PHP lint, Pint, **PHPStan level 1**, **24 service + 7 role-matrix tests**; Tickets/Blueprints/Goalcanvas suites unaffected.
- 5-lens adversarial review verdict **SHIP** (no survivors; the ungated ticket-hours reads and the repo-layer legacy `session('userdata.role')` filtering are documented exceptions/follow-ups).

## Scope / follow-ups
Documented low-sev follow-ups (not blockers): `punchIn`/`punchOut` gate `create` but don't validate the ticket's project (a user could log their own time against a ticket they can't access); the repo's `getAllAccountTimesheets` still uses a legacy `session('userdata.role')` filter. Remaining rollout domains: **Files → Reports → Calendar → Projects (last)**, plus a separate **Connector** import-authz pass.

### DIFF
--- a/app/Core/Auth/Permissions/DefaultRolePermissions.php
+++ b/app/Core/Auth/Permissions/DefaultRolePermissions.php
@@ -60,7 +60,14 @@ final class DefaultRolePermissions
             // editor+, but a commenter is allowed to add comments — grant the key explicitly.
             ['scope' => 'project', 'keys' => ['comments.create']],
         ],
-        'editor' => [['scope' => 'project', 'verbs' => ['create', 'edit', 'delete']]],
+        'editor' => [
+            ['scope' => 'project', 'verbs' => ['create', 'edit', 'delete']],
+            // Timesheets are GLOBAL-scoped (company-wide time logging), so the project verb rule
+            // above does NOT match them — an editor's own-time capability is granted by explicit
+            // global keys. Ownership (own vs others) is enforced in the service; the cross-user
+            // `timesheets.manage` stays manager+ (below).
+            ['scope' => 'global', 'keys' => ['timesheets.view', 'timesheets.create', 'timesheets.edit', 'timesheets.delete']],
+        ],
         'manager' => [
             ['scope' => 'project', 'verbs' => ['*']],
             // Managers may INVITE users (the NewUser screen is manager+). The client-scoping
@@ -69,7 +76,9 @@ final class DefaultRolePermissions
             // those remain admin+. users.* are company-wide, so this is an explicit global
             // key grant rather than a 'create' verb rule (a verb rule would also need a global
             // scope and is fine, but the explicit key documents that ONLY create is intended).
-            ['scope' => 'global', 'keys' => ['users.create']],
+            // timesheets.manage (company-wide invoicing/reports/others' time) is manager+; the
+            // editor keys above are inherited up the hierarchy.
+            ['scope' => 'global', 'keys' => ['users.create', 'timesheets.manage']],
         ],
         'admin' => [
             ['scope' => 'any', 'verbs' => ['*'], 'exclude' => ['company.settings.*']],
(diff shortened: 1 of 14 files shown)


marcelfolaron:
Addressed in 22bb29cfe7:

- **AddTime GET gated with `timesheets.view`** → changed to **`timesheets.create`**. The GET renders the creation form, so a view-only role shouldn't load it. Grant-equivalent for every built-in role (the set holding `timesheets.create` is exactly editor+, matching the prior `authOrRedirect([owner,admin,manager,editor])`); only tightens for future custom view-only roles.
- **Docblock "writes throw" vs delete/update soft-deny** → rewrote the class docblock to describe the actual two-tier contract: the base capability is checked with `authorize()` (throws `AuthorizationException` for missing create/edit/manage), while the *ownership* check that follows soft-denies (`deleteTime()` returns false, `updateTime()` returns early). `AuthorizationException` stays referenced in the docblock, so it's not an unused import (Pint treats docblock types as usage — confirmed green).

The two stub-closure findings (potential `ArgumentCountError`) are **false positives**: PHP silently ignores extra positional args passed to a closure that declares fewer parameters (`fn () => true` invoked with `($key, $projectId, $forceGlobal)` is legal; only *too few required* args throws). The unit job passes in CI, which empirically confirms no error.

PR #3492, opened 2026-06-04
security(files): gate the Files domain + close RPC IDORs

marcelfolaron:
## Summary

Brings the **Files** domain onto the native DB-backed permission engine and closes four fail-open `@api` IDORs. Part of the service-layer hardening rollout (after Sprints, Comments, Wiki/Ideas, the canvas family, and Timesheets).

`FilesPermissions`: `files.view` (readonly+) / `files.upload` (commenter+) / `files.delete` (editor+) — all standard project verbs, so they auto-grant through the existing matrix with **no `DefaultRolePermissions` edit**.

## IDORs closed (were reachable by any authenticated JSON-RPC caller)

| Method | Before | After |
|---|---|---|
| `getFilesByModule` | no check — enumerate any project's files by guessing ids; bare call dumps the whole table | resolves the target's real project, requires `files.view` (soft-deny `[]`); owner-restricted → owner-only; empty/unknown module → `[]` |
| `upload` | no service check (only the controller pre-gated) | authorizes `files.upload` against the resolved project in-body — guards the RPC path |
| `deleteFile` | owner **OR manager+ globally** (delete any file in any project) | owner always; else `files.delete` **in the file's own project** (editor+; admin/owner bypass membership) |
| `getFileForUser` | trusted the passed `$userId` for the owner check (spoofable via RPC) | authorizes the **session** user; project files via `can(files.view, project)` |

`handleFileAction` drops `@api` (a `$_POST`/`$_FILES`-shaped controller helper, not an API surface; its delegates self-authorize). `userCanUploadToModule` migrated to the engine; dead `ProjectRepository` dependency removed. `Browse`/`ShowAll` get `#[RequiresPermission(files.view)]` (session project) as defense-in-depth.

## Grant-equivalence

The only intentional behavior changes are the two confirmed with the maintainer:
- **Non-owner delete → editor+** (was owner-OR-manager+). Aligns with the editor-gated delete button (which silently failed for non-owned files before); a non-owner editor in the project gains delete. Closes the manager-global cross-project bypass.
- **Upload → commenter+** (was any project member incl. readonly). Drops a `readonly` member's raw-endpoint upload; the UI never offered it to them.

Owner-delete, view, and admin/owner bypass are **preserved**.

## Notes / documented follow-ups (non-blocking)
- The `client` module has no project mapping; client-file listing/upload authz stays a Clients-domain concern (`ShowClient` is admin/manager-gated) — to be addressed with the Clients rollout.
- An admin replacing **another** user's avatar no longer deletes the old owner-restricted file (orphan, not a security issue).
- `getModules` (a non-`@api` UI dropdown helper) still uses `Auth::userIsAtLeast(admin)` — no clean engine equivalent; left as-is.

## Migration / tests
- `update_sql_30515`: flush registry → `syncDiscoveredPermissions` → `seedBuiltInRoles` (no matrix edit). dbVersion 3.5.14 → **3.5.15**.
- Rewrote `FilesServiceTest` (its old `userCanUploadToModule` tests asserted the pre-engine impl) with fail-closed authz coverage; added `files.*` to `DefaultRolePermissionsTest`.

Reviewed via a 5-lens adversarial pass (grant-equivalence / fail-closed-IDOR / consumer-regression / test-correctness / runtime-correctness): verdict SHIP, 0 blockers.

### DIFF
--- a/app/Core/Auth/Permissions/PermissionServiceProvider.php
+++ b/app/Core/Auth/Permissions/PermissionServiceProvider.php
@@ -74,14 +74,20 @@ class PermissionServiceProvider extends ServiceProvider
     }
 
     /**
-     * Inject PermissionService into every service that extends {@see BaseService}, without
-     * forcing subclass constructors to wire it. The afterResolving callback registered for
-     * the base type fires for any resolved instance that is `instanceof BaseService`.
+     * Wire PermissionService into every service that extends {@see BaseService}, without forcing
+     * subclass constructors to wire it. The afterResolving callback fires for any resolved instance
+     * that is `instanceof BaseService`.
+     *
+     * We wire a LAZY resolver, not the instance: a BaseService can sit inside PermissionService's
+     * own dependency graph (Files is reached via PermissionService → ChecksProjectAccess → Projects
+     * → Files), so eagerly calling `make(PermissionService)` here would re-enter PermissionService's
+     * half-built construction and recurse infinitely (stack overflow at boot). Resolving lazily on
+     * first authorize()/can() defers it until the singleton has been built.
      */
     protected function injectBaseServiceDependencies(): void
     {
         $this->app->afterResolving(BaseService::class, function (BaseService $service, $app) {
-            $service->setPermissionService($app->make(PermissionService::class));
+            $service->setPermissionServiceResolver(fn () => $app->make(PermissionService::class));
         });
     }
 }
--- a/app/Core/Configuration/AppSettings.php
+++ b/app/Core/Configuration/AppSettings.php
@@ -9,5 +9,5 @@ class AppSettings
 {
     public string $appVersion = '3.8.0';
 
-    public string $dbVersion = '3.5.14';
+    public string $dbVersion = '3.5.15';
 }
--- a/app/Core/Domains/BaseService.php
+++ b/app/Core/Domains/BaseService.php
@@ -13,21 +13,54 @@ use Leantime\Core\Exceptions\ValidationException;
  * unchanged).
  *
  * Dependency wiring is handled by {@see \Leantime\Core\Auth\Permissions\PermissionServiceProvider}:
- * an `afterResolving(BaseService::class, ...)` hook calls {@see setPermissionService()} on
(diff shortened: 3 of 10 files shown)


marcelfolaron:
Addressed in d6e449fa8d:

- **`getFilesByModule('client')` enumeration** → the lenient `client` branch now requires a **specific** client id. A `module='client'` call with no `entityId` returns `[]` instead of dumping every client's files. The legitimate caller (`Clients::getClient` via the admin-gated `ShowClient`) always passes a client id, so it's unaffected. The per-id client read stays a documented Clients-domain follow-up — gating it properly needs `clients.*` semantics that aren't synced yet (gating on an unsynced permission would 403 admins on the client file panel). Added two tests (no-id → `[]`, with-id → passthrough).

The stub-closure finding (potential `ArgumentCountError`) is a **false positive**: PHP silently ignores extra positional args passed to a closure that declares fewer parameters, so `fn () => true` invoked with `($key, $projectId, $forceGlobal)` is legal. The unit job passes green in CI, confirming no error.

marcelfolaron:
Addressed all four in 44d8810277 — good catches:

- **`upload()` / `userCanUploadToModule()`**: a project-scoped module (`project`/`ticket`) whose project can't be resolved (invalid/deleted entity) now **fails closed** (`upload()` throws, `userCanUploadToModule()` returns `false`) instead of falling through to the non-project "allow" path. Added `PROJECT_SCOPED_MODULES = ['project','ticket']` (the set `getProjectIdForFile()` resolves).
- **`getFileForUser()`**: same fail-open on the read path — an orphaned `ticket` file (deleted ticket → no project) is now denied (403) instead of served without a `files.view` check.
- **`handleFileAction()`**: now catches `AuthorizationException` (so a denied upload renders a clean "upload failed" notification rather than a 403 page) and reflects the real `upload()` result (`success = is_array($result)`) instead of reporting `true` unconditionally.

Legitimate flows always carry a valid project/ticket id, so they resolve normally and are unaffected — only orphan (invalid/deleted) ids are denied. +5 tests (32 total), full Files suite + PHPStan + Pint green locally.

Issue #3495, opened 2026-06-05
project list


SagarJadhav30:
### What is your set up?

Cloud Hosted

### Version

idk

### Describe the issue

feature

### Reproduction steps

1.
2.
3.
...

eVp3x:
@SagarJadhav30 fix: https://github.com/Leantime/leantime/pull/3499

marcelfolaron:
Thanks for the report — unfortunately there isn't enough information here to act on (no version, no description of the problem, no reproduction steps). If you were hitting the 500 error on the To-Dos list view, that's tracked in #3466 and is already fixed on master for the next release. Otherwise, could you update this issue with what you expected to happen, what happened instead, and steps to reproduce? Happy to take another look then.

SagarJadhav30:
Okk

marcelfolaron:
Closing — we still don't have enough information to act on this (version, what's wrong with the project list, expected vs. actual behavior, reproduction steps). If the problem persists, please open a new issue using the bug template with those details.

Issue #3470, opened 2026-06-02
500 Error when clicking "to-do"


mymctrashmail-sketch:
### What is your set up?

Cloud Hosted

### Version

idk

### Describe the issue

i get an error message when i try to navigate to the "to-dos" section from the goals page 

### Reproduction steps

1.click goals 
2.click to dos
3.
...

### Error Logs (LEANTIMEFOLDER/storage/logs)

<img width="2293" height="1084" alt="Image" src="https://github.com/user-attachments/assets/c8f6570f-db06-4c89-9501-df79a8352146" />

<img width="290" height="961" alt="Image" src="https://github.com/user-attachments/assets/2fe6ad73-dd1b-4a5f-8da2-ded59394d64a" />

eVp3x:
@mymctrashmail-sketch fix: https://github.com/Leantime/leantime/pull/3499

marcelfolaron:
This is fixed on master in d495e72fd — the list view 500 was caused by a leftover `$groupBy = $groupBy;` self-assignment from the Blade template migration in `showList.blade.php`. The fix will be included in the next release. Until then it only affects v3.8.0.

marcelfolaron:
This was fixed and has shipped in the v3.9.0 release line (latest **v3.9.4**). Closing now that the staged fix has been released — please upgrade and reopen if the issue persists.

Issue #3466, opened 2026-06-02
500 error when selecting the list view for the to-dos.


Kopfkirmes:
### What is your set up?

Cloud Hosted

### Version

3.8.0

### Describe the issue

When I open the To-Dos and switch the view to “List,” the page immediately returns a 500 (Internal Server Error). The List view does not load at all, while other views appear to work—so the error seems specific to the List view.

### Reproduction steps

1. Projekt Overview
2. To-Dos
3. List
...

spacenono:
Confirmed on **self-hosted 3.8.0** (PHP 8.3, MySQL).

### Root cause

In `app/Domain/Tickets/Templates/showList.blade.php`, the top `@php` block has a self-assignment referencing a variable that is never passed to the view:

```php
$groupBy = $groupBy;
```

`ShowList.php` doesn't provide `$groupBy` (grepping for `groupBy` in the controller returns nothing), so this line throws `Undefined variable $groupBy` and the List view 500s. Kanban and Table views are unaffected.

The real value lives in `$searchCriteria['groupBy']`, which the template already uses elsewhere — e.g. the hidden input:

```php
<input type="hidden" name="groupBy" value="{{ htmlspecialchars((string) ($searchCriteria['groupBy'] ?? ''), ENT_QUOTES, 'UTF-8') }}" />
```

and the JS init:

```php
leantime.ticketsController.initTicketsList("{{ $searchCriteria['groupBy'] }}");
```

### Fix

Replace the self-assignment so it reads from `$searchCriteria`:

```php
$groupBy = $searchCriteria['groupBy'] ?? '';
```

This aligns with how the rest of the template already sources the value and removes the undefined-variable error. Confirmed working on 3.8.0.

eVp3x:
@Kopfkirmes @spacenono fix: https://github.com/Leantime/leantime/pull/3499

marcelfolaron:
This is fixed on master in d495e72fd — the list view 500 was caused by a leftover `$groupBy = $groupBy;` self-assignment from the Blade template migration in `showList.blade.php`. The fix will be included in the next release. Until then it only affects v3.8.0.

marcelfolaron:
This was fixed and has shipped in the v3.9.0 release line (latest **v3.9.4**). Closing now that the staged fix has been released — please upgrade and reopen if the issue persists.

PR #3499, opened 2026-06-05
Fix for List view on To-Dos

eVp3x:
Fix undefined $groupBy in showList view. Restore the missing groupBy template variable to prevent errors when opening the To-Do List View. After this fix a user is able to press on `List` view on To-Dos.

### Description

**Suggested change**

Add the missing `groupBy` value to the template assignments:

```php
'groupBy' => $searchCriteria['groupBy'] ?? '',
```

**Reasoning**

The `showList` view expects a `$groupBy` variable, but `getTicketTemplateAssignments()` does not provide it. Passing the current `groupBy` value restores the expected contract between the service and the view, preventing the "Undefined variable $groupBy" error while preserving the user's selected grouping state.

### Link to ticket

https://github.com/Leantime/leantime/issues/3470
https://github.com/Leantime/leantime/issues/3495
https://github.com/Leantime/leantime/issues/3466

### Type

- [X] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result

image

### DIFF
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -3642,6 +3642,7 @@ class Tickets extends BaseService
             'futureSprints' => $futureSprints,
             'users' => $users,
             'milestones' => $milestones,
+            'groupBy' => $searchCriteria['groupBy'] ?? '',
             'groupByOptions' => $groupByOptions,
             'newField' => $newField,
             'sortOptions' => $sortOptions,


marcelfolaron:
Thanks for the contribution and the correct diagnosis, @eVp3x! 🙏 The root cause you identified — the `$groupBy = $groupBy;` self-assignment left over from the Blade migration — is real in v3.8.0, but it was already fixed on master in d495e72fd (May 30), which guards `$groupBy`/`$newField` directly in `showList.blade.php` and `showAll.blade.php`. On current master the list view no longer reads a top-level `$groupBy` (templates use `$searchCriteria['groupBy']` directly), so this change would add an unused variable.

The fix will ship in the next release. Closing as already fixed on master — appreciate you taking the time to debug and share your patch!

Issue #3462, opened 2026-06-02
Error- Cannot Create Milestones


mymctrashmail-sketch:
### What is your set up?

Cloud Hosted

### Version

3.8.0

### Describe the issue

Getting an error message when i try to create a milestone

### Reproduction steps

1. go to my project
2. click milestones
3. click new + add milestone
...

### Error Logs (LEANTIMEFOLDER/storage/logs)

{ "message": "Server Error" }

<img width="1936" height="727" alt="Image" src="https://github.com/user-attachments/assets/20a6cba8-6d14-4d05-a3ee-3aa780f8738f" />

r41ns4n:
Hi leantime team,

I have the same issue, but I use **another setup**:

Self-Hosted via Docker image: leantime/leantime:${LEAN_VERSION:-latest}

**Describe the issue**
Getting an error message when i try to create a milestone

**Investigations**
- permissions (set to www-data:www-data and 775)
- php.ini (memory_limit = 2G)
- rm -rf sessions, cache & views

**Logs**
[previous exception] [object] (ErrorException(code: 0): Attempt to read property \"id\" on null 
at /var/www/html/storage/framework/views/a5cffaf4ff5adf19af8e4f6a3ec0cacd.php:11)

app/Domain/Tickets/Controllers/EditMilestone.php(104)
Template->displayPartial('tickets.milesto...')

eVp3x:
@r41ns4n @mymctrashmail-sketch i fixed the problem: https://github.com/Leantime/leantime/pull/3500

marcelfolaron:
This is fixed on master in 804fea492 (#3444) — the milestone dialog read a self-referential `$currentMilestone` variable that the controller never assigns (Blade migration artifact), causing the "Attempt to read property id on null" error. The fix will be included in the next release. Until then it only affects v3.8.0.

marcelfolaron:
This was fixed and has shipped in the v3.9.0 release line (latest **v3.9.4**). Closing now that the staged fix has been released — please upgrade and reopen if the issue persists.

PR #3500, opened 2026-06-05
Fix for "Add Milestone" broken feature

eVp3x:
Refactor currentMilestone assignment logic. Make the template work whether the controller passes currentMilestone or milestone. This change fix the Issue of the current broken "Add Milestone" feature.

### Description

In a simple and not destructive way we are going to fix the current broken "Add Milestone" feature, i was happily using this patch for my Leantime implementation and so i decided to share it. Thank you for the awesome project!

### Link to ticket

Links to the GitHub issue being addressed by this change:
https://github.com/Leantime/leantime/issues/3462
https://github.com/Leantime/leantime/issues/3489

### Type

- [X] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result

image

### DIFF
--- a/app/Domain/Tickets/Templates/milestoneDialog.blade.php
+++ b/app/Domain/Tickets/Templates/milestoneDialog.blade.php
@@ -1,5 +1,5 @@
 @php
-    $currentMilestone = $milestone ?? null;
+    $currentMilestone = $currentMilestone ?? $milestone ?? null;
     $milestones = $milestones ?? [];
     $statusLabels = $statusLabels ?? [];
 @endphp


marcelfolaron:
Thanks for sharing your patch and the kind words about the project, @eVp3x! 🙏 The bug is real in v3.8.0 — the dialog had a self-referential `$currentMilestone = $currentMilestone ?? null;` while the controller assigns `milestone`, which is exactly the "Attempt to read property id on null" crash. It was already fixed on master in 804fea492 (May 31, #3444), which points the variable at `$milestone ?? null` — the same fallback your patch adds. On current master this change would be a no-op, so closing as already fixed.

The fix ships in the next release. Glad the patch kept your installation running in the meantime!

PR #3501, opened 2026-06-09
security(reports): gate the Reports domain + de-expose system/telemetry RPC surface

marcelfolaron:
## Summary

Brings the **Reports** domain onto the native permission engine. Part of the service-layer hardening rollout (after Sprints, Comments, Wiki/Ideas, the canvas family, Timesheets, and Files).

`ReportsPermissions`: a single `reports.view` — project-scoped standard verb, auto-grants readonly+ through the existing matrix, **no `DefaultRolePermissions` edit**.

## RPC IDORs closed

| Method | Before | After |
|---|---|---|
| `getSprintBurndownForReport` | any authed RPC caller could pull **any** project's burndown by id | `#[RequiresPermission(reports.view, projectIdParam: 'projectId')]` — authorized against the **requested** project at dispatch |
| `getFullReport` | same — any project's stored report history | same gate |
| `getRealtimeReport` | same — any project's live ticket report | same gate (repo additionally scopes by both `projectId` AND `sprint`; the inner `getSprint` is already entityScoped-hardened — foreign sprint ids yield nothing) |

`projectId` is a **required** param on all three, so the gate can't be dodged by omission (no session fallback possible).

## System/telemetry surface de-exposed (de-`@api`'d, kept working internally)

- **`optOutTelemetry`** — company-wide settings **mutation**: any authenticated user could opt the whole instance out of telemetry over RPC. Still invoked internally by the admin-gated company-settings save.
- **`getAnonymousTelemetry`** — instance-wide usage aggregates + company GUID (reconnaissance over RPC).
- **`getProjectStatusReport`** — ALL projects' status counts, no project scope.
- **`sendAnonymousTelemetry`** — outbound-HTTP system task (scheduler + dashboard trigger only).
- **`dailyIngestion`** — session-bound web helper (only the gated `Reports\Show` calls it).

The scheduler paths (`register.php` → `cronDailyIngestion` / `sendAnonymousTelemetry`) are direct internal calls with **no session user** and are unaffected — verified end-to-end.

## Grant-equivalence

**One maintainer-approved change** (confirmed via explicit question): the `/reports/show` page loosens from editor+ to **readonly+** — it only aggregates data readonly project members can already read item-by-item (tickets, sprints, milestones are all readonly-visible). Simultaneously the three RPC reads **harden** from no-gate to member-of-target-project. The legacy `authOrRedirect` in `init()` is replaced by action attributes, which the Frontcontroller enforces **before instantiation** — so `init()`'s `dailyIngestion()` never runs for a denied user.

## Migration / tests

- `update_sql_30516`: flush registry → sync → re-seed (no matrix edit). dbVersion 3.5.15 → **3.5.16**.
- Two reflection surface tests lock the exposure contract: the 3 reads must stay `@api` **and** carry the `projectIdParam` gate; the 7 system/internal methods must never regain a line-starting `@api`.
- `reports.view` added to the role-matrix tests (readonly exact list + commenter).
- Full local Unit suite green (532 tests / 1252 assertions); boot verified; PHPStan L1 + Pint clean.

Reviewed via a 4-lens adversarial pass (grant-equivalence / fail-open-IDOR / consumer-cron / test-runtime): verdict **SHIP, 0 blockers**.

### DIFF
--- a/app/Core/Configuration/AppSettings.php
+++ b/app/Core/Configuration/AppSettings.php
@@ -9,5 +9,5 @@ class AppSettings
 {
     public string $appVersion = '3.8.0';
 
-    public string $dbVersion = '3.5.15';
+    public string $dbVersion = '3.5.16';
 }
--- a/app/Domain/Install/Repositories/Install.php
+++ b/app/Domain/Install/Repositories/Install.php
@@ -96,6 +96,7 @@ class Install
         30513,
         30514,
         30515,
+        30516,
     ];
 
     /**
@@ -3010,6 +3011,31 @@ class Install
             Log::error('Migration 30515: '.$e->getMessage());
 
             return ['Migration 30515 failed: '.$e->getMessage()];
+        }
+
+        return true;
+    }
+
+    /**
+     * Sync the permission catalog + re-seed built-in role grants for the Reports domain. Adds
+     * reports.view (project-scoped standard verb → auto-grants readonly+; no
+     * DefaultRolePermissions matrix edit). The re-seed lands the new key in zp_role_permissions
+     * for the built-in roles.
+     *
+     * Flush the discovered-provider cache FIRST so ReportsPermissions is rediscovered.
+     */
+    public function update_sql_30516(): bool|array
+    {
+        try {
+            app(\Leantime\Core\Auth\Permissions\PermissionRegistry::class)->flush();
+
+            $seeder = app(\Leantime\Core\Auth\Permissions\PermissionSeeder::class);
+            $seeder->syncDiscoveredPermissions();
+            $seeder->seedBuiltInRoles();
+        } catch (\Exception $e) {
+            Log::error('Migration 30516: '.$e->getMessage());
+
+            return ['Migration 30516 failed: '.$e->getMessage()];
         }
 
         return true;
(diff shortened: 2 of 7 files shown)


marcelfolaron:
Addressed in 4471d00c6e — good catch, this was a real gate-binding fail-open.

`getFullReport` / `getRealtimeReport` now type `$projectId` as `int`. The enforcer binds the gate via `isset($params['projectId'])`, and `isset()` is false for an explicit `null`, so `projectId: null` would have fallen back to the session project (dodging the per-target binding). Typing it `int` makes `null` an invalid JSON-RPC param (rejected at param-prep, before the method runs), keeping the check bound to the requested project. `getSprintBurndownForReport` was already `int`; `$sprintId` stays mixed (the empty string is the meaningful "backlog" selector). All real callers already pass an int — no behavior change for legitimate flows.

Note: the underlying `isset()`-vs-explicit-null behavior in `PermissionEnforcer::resolveProjectId` affects any `projectIdParam`-gated method generally; typing the params closes it here, and I've logged the engine-level hardening (treat a declared-but-null `projectIdParam` as deny rather than session-fallback) for the upcoming engine/Projects pass.

The NPM Security Audit failure is unrelated to this PHP-only PR: a new **critical** advisory in the transitive npm dep `shell-quote` (GHSA-w7jw-789q-3m8p) was published in the last day — master's audit passed on 06-08 and would fail identically now. It needs a separate `npm audit fix` (does not touch this PR).

PR #3513, opened 2026-06-12
test(api): Bearer-auth JSON-RPC contract suite + CI gate

gloriafolaron:
## Summary

- Adds the missing test coverage for the **Bearer-token** auth path on `/api/jsonrpc` — the only one of the three auth modes (session / x-api-key / Bearer) without a CI gate.
- Mobile and AdvancedAuth integrators hit `/api/jsonrpc` via Bearer. A permission-engine deploy in master silently broke that path with `-32001 "You are not allowed to perform this action"` on every gated read — `ApiCest` stayed green because it tests x-api-key, not Bearer. This PR closes that blind spot.

## What's included

- **`app/Command/CreateTestTokenCommand.php`** — `php bin/leantime test:create-token --email=...` mints a Bearer token via `AccessTokenRepository::createToken()`. Plugin-independent — no AdvancedAuth required. `--quiet-output` for shell capture.
- **`tests/Acceptance/API/BearerApiCest.php`** — sibling to `ApiCest`. Same Codeception harness, same JSON-RPC envelope, Bearer header + the regression-class assertion: every primitive asserts `error.code !== -32001`.
  - whoami (`Users.getUser`)
  - `Projects.getProjectsUserHasAccessTo`
  - `Tickets.getAllOpenUserTickets`
  - `Tickets.quickAddTicket` → `Tickets.getTicket` pair (the projectIdParam-resolved engine check)
  - `Comments.getComments` (the entityScoped engine check)
  - `Notifications.getUnreadCount`
  - missing-bearer → 401
- **`.github/workflows/api-bearer-tests.yml`** — mirrors `acceptancetests.yml`. Runs on PR and master push.
- **`tests/Httprequests/JsonRPC-bearer.http`** — IDE-runnable companion. Same coverage, drives via a `{{ bearer-token }}` env var that `test:create-token` feeds. For local debugging without spinning up Codeception.
- **`makefile`** — adds `bearer-api-test-ci` (same docker-compose pattern as `acceptance-test-ci`, filtered to `--group bearer-api`).

## Why now

The deploy that landed on cloud broke `Tickets.getTicket` and `Comments.getComments` for every Bearer-authed caller — every mobile user, every integrator using AdvancedAuth tokens. CI stayed green because the permission engine resolves user-context correctly under x-api-key (what `ApiCest` tests) and session (what `LoginCest` tests) — only Bearer fails. This suite would have failed the relevant PRs before merge.

## Test plan

- [ ] CI on this PR runs the new `Bearer API tests` workflow and passes (or fails — if it fails, that's the live regression and a sign we need to ship the engine fix before merging this).
- [ ] Locally: `make bearer-api-test-ci` runs the suite end-to-end.
- [ ] Locally: `php bin/leantime test:create-token --email=test@leantime.io --quiet-output` prints a usable token.

### DIFF
--- a/app/Command/CreateBearerTokenCommand.php
+++ b/app/Command/CreateBearerTokenCommand.php
@@ -12,22 +12,30 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 #[AsCommand(
-    name: 'test:create-token',
-    description: 'Mint a Sanctum-style Bearer token for an existing user. Plugin-independent — does not require AdvancedAuth. Intended for CI + local testing of the JSON-RPC surface under Bearer auth.',
+    name: 'auth:create-bearer-token',
+    description: 'Mint a Sanctum-style Bearer token for an existing user. Plugin-independent — does not require AdvancedAuth. CLI-only.',
 )]
-class CreateTestTokenCommand extends Command
+class CreateBearerTokenCommand extends Command
 {
     protected function configure(): void
     {
         parent::configure();
 
         $this->addOption('email', null, InputOption::VALUE_REQUIRED, 'Email of an existing user to mint a token for')
-            ->addOption('name', null, InputOption::VALUE_OPTIONAL, 'Token name (logged on the row, useful for traceability)', 'test-bearer')
-            ->addOption('quiet-output', null, InputOption::VALUE_NONE, 'Print only the token (no decoration). Useful for shell capture in CI.');
+            ->addOption('name', null, InputOption::VALUE_OPTIONAL, 'Token name (logged on the row, useful for traceability)', 'cli-issued')
+            ->addOption('quiet-output', null, InputOption::VALUE_NONE, 'Print only the token (no decoration). Useful for shell capture.');
     }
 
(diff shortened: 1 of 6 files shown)


app/Command/CreateBearerTokenCommand.php
```diff
@@ -0,0 +1,79 @@
+<?php
```
marcelfolaron, line 1: Just call it CreateBearerTokenCommand
There is nothing special about this being a "test" token vs minting real tokens.

marcelfolaron: Just make sure it can't be run via some other means. Only CLI

gloriafolaron:
Both addressed in 645913fe6:

1. Renamed `CreateTestTokenCommand` → `CreateBearerTokenCommand`, command name `test:create-token` → `auth:create-bearer-token`, description rewritten to frame it as a general bearer-minting tool (not test-only).
2. Added a `runningInConsole()` guard at the top of `execute()` — bails before any work if invoked outside CLI. Laravel's console kernel already gates this in practice, but the guard makes it explicit.

Also updated the `.http` setup comment to reference the new command name.

PR #3529, opened 2026-06-14
feat(api): session-scoped mobile endpoints (notifications inbox + calendar)

marcelfolaron:
## Why

A post-3.9.4 mobile audit found the inbox and calendar tabs were calling RPC methods that take an arbitrary `$userId` and are **deliberately not `@api`-exposed** (IDOR guards), so they returned `-32601 Method not found`. The fix is the same pattern already used by `getUnreadCount` / `getMyExternalCalendars`: add **session-scoped siblings** that resolve the user from the token, leaving the userId-trusting originals internal.

## What

**New `@api` endpoints (mobile drops the `userId` param, calls with `{}`):**
- `Notifications.getInbox()` — inbox list for the session user. `getAllNotifications($userId)` stays internal/non-`@api`. (Mark-all-read is already available via `markRead('all')`.)
- `Calendar.getMyCalendar(from?, until?)` — session-scoped calendar feed; pins to `currentUserId()`. `getCalendar($userId)` stays internal-only (the web `ShowMyCalendar` controller keeps using it).

**Tag two reads that were *already* session-scoped and only lacked `@api`:**
- `Calendar.getICalUrl()` and `Calendar.getExternalCalendarEvents()` — both key off `session('userdata.id')`, no spoofable param. Added `@api` + `#[RequiresPermission(calendar.view)]`.

> The `calendar.view` gate is project-scoped, but on an API call there is no session project, so it resolves **capability-only** via the effective role (readonly+ holds `calendar.view`). It does **not** re-introduce `-32001`. Verified against the identical existing `getMyExternalCalendars` path + the enforcer/`currentUserCan` code.

**Security — `security(notifications)`:** `markNotificationUnread` was `@api`, ignored its `$userId` param, and the repo did an **unscoped `where('id')` update** → any authenticated user could flip any notification unread by guessing its id (IDOR; ids are a global sequence). Now session-scoped (matches `(id, session user)`); the misleading `$userId` param is dropped (RPC param-matching ignores the extra arg mobile still sends, so it's backward-compatible). A cross-user regression test is included.

**i18n — `fix(i18n)`:** the "My To-Dos" widget growls asked `Language::__()` for 11 `notifications.*` keys that only exist under `short_notifications.*` (or not at all), so the raw key leaked (e.g. `notifications.date_updated` shown literally). Repointed to `short_notifications.*` — matching the JS twins (`ticketsController.js`/`dashboardController.js`) and the success path (`short_notifications.status_updated`) already on that prefix — and added the 9 genuinely-missing keys to `en-US.ini`. (The header inbox dropdown was a red herring — it renders the stored, pre-translated message.)

**Error semantics:** `getICalUrl` now throws `MissingParameterException` (clean RPC `-32602`) instead of a plain `\Exception` (generic `-32000`) when no iCal feed is configured. Still an `\Exception`, so the web `Export` controller's `catch` is unaffected.

## Tests
`BearerApiCest` now asserts `getInbox` / `getMyCalendar` / `getExternalCalendarEvents` / `getICalUrl` exposure over Bearer for **both** the owner and a **non-manager editor** (the per-project authz path that owner-only testing masks), plus a cross-user `markNotificationUnread` IDOR regression.

Local: `php -l`, Pint, and PHPStan (level 1) all green on the changed files. Acceptance runs in CI (the Docker container mounts the main repo, not a worktree).

## Verification by an independent adversarial review
- `#[RequiresPermission]` is enforced only at dispatch entry points (Frontcontroller / CheckPermissions / Jsonrpc) — **inert on internal service-to-service calls**, so tagging the existing `getICalUrl`/`getExternalCalendarEvents` does not affect their web callers.
- The null-project `calendar.view` path resolves capability-only (no fail-closed).


marcelfolaron:
Thanks @Copilot — all three addressed in `1ddaf824e4`:

1. **`getInbox` unbounded limit / arbitrary filter columns** — dropped the repo's `filterOptions` passthrough entirely (mobile's inbox only needs `showNewOnly` + paging, and the arbitrary `where($key,$value)` allowed unknown columns → SQL errors) and clamped pagination: `limitStart >= 0`, `limitEnd` to `1..100` (default 50). An `@api` caller now gets a bounded page of its **own** rows only.
2. **`getMyCalendar` `userId = 0` fallback** — now guards `currentUserId() === 0` and returns `[]` when unauthenticated, matching `getInbox` (no query as user 0).
3. **Weak `getICalUrl` test** — strengthened: it now also fails on `-32001` (permission regression) and asserts the response is **either** a URL result **or** `-32602` (no feed configured) — never `-32601`/`-32001`.

PR #3544, opened 2026-06-18
fix(blueprints): canvas add/edit item returns 404 (canvasSlug bound to $id)

marcelfolaron:
## Problem

Opening any Blueprints canvas and clicking **"+ Add New"** (or editing an existing item) showed a **"Page not found / 404"** inside the modal — for every canvas type (insights, minempathy, swot, …). The request actually returned **HTTP 200** with the 404 *page* as its body, and nothing was logged (the env runs with `app.debug` off), which made it hard to spot.

## Root cause

The Blueprints controllers are routed under `blueprints/{canvasSlug}/{action}/{id?}` (two path params), but their action methods were declared `get(?string $id = null)` / `post(?string $id = null)` — they omitted the `{canvasSlug}` parameter.

Laravel binds route parameters to method parameters in order, so the **first route param (`canvasSlug`) was bound to the first method param (`$id`)**. `$id` therefore received the slug string (e.g. `"insights"`), and the real id was dropped.

`EditCanvasItem`/`EditCanvasComment` then treated the slug as an item id — `getCanvasItem((int)"insights" = 0, …)` returns `false` — and returned `displayPartial('errors.error404')`.

It was latent in *all* canvas controllers, but only the "edit item" actions failed visibly. `ShowCanvas`/`BoardDialog`/`DelCanvasItem` masked the same misbinding by falling back to the session board on a bad id — which also meant they were **silently ignoring the board id in the URL** (`showCanvas/27` loaded the session board, not 27).

Regression from the dispatch-bridge removal — the old `blueprintsDispatch()` helper synced the path id into `$_GET['id']`.

## Fix

Declare `?string $canvasSlug = null` as the **first parameter** on every action method routed under the `{canvasSlug}` prefix (`ShowCanvas`, `EditCanvasItem`, `EditCanvasComment`, `BoardDialog`, `DelCanvas`, `DelCanvasItem`, `Export`). This realigns `$id` to the real id. The slug is still read in each constructor via `$request->route('canvasSlug')`; the parameter only exists to keep route→method binding aligned. Method bodies are unchanged.

## Verification

- Confirmed the resolved arg for `get()` was `{"canvasSlug":"insights"}` → `$id === "insights"`; after the fix, `editCanvasItem?type=…` → `id=NULL` (add form) and `editCanvasItem/15` → `id="15"` (edit the real item).
- Live server (`leantime-oss.test`): add-item, edit-existing (real 18 KB form), `editCanvasComment`, and `showCanvas` all return 200 with real content; verified for both `insights` and the originally-reported `minempathy`.
- Browser end-to-end at the exact reported URL: the modal renders the item form (Description / Status / Relates to…) instead of the 404.
- `./vendor/bin/pint --test` passes on all changed files.

### DIFF
--- a/app/Domain/Blueprints/Controllers/BoardDialog.php
+++ b/app/Domain/Blueprints/Controllers/BoardDialog.php
@@ -58,10 +58,11 @@ class BoardDialog
     /**
      * get - display the create/edit board dialog.
      *
+     * @param  string|null  $canvasSlug  Canvas type slug from the route (resolved in the constructor)
      * @param  string|null  $id  Current board id
      */
     #[RequiresPermission(BlueprintsPermissions::VIEW, entityScoped: true)]
-    public function get(?string $id = null): Response
+    public function get(?string $canvasSlug = null, ?string $id = null): Response
     {
         if ($this->template === null) {
             return $this->tpl->displayPartial('errors.error404');
@@ -88,10 +89,11 @@ class BoardDialog
     /**
      * post - handle create/edit board submissions.
      *
+     * @param  string|null  $canvasSlug  Canvas type slug from the route (resolved in the constructor)
      * @param  string|null  $id  Current board id
      */
     #[RequiresPermission(BlueprintsPermissions::EDIT, entityScoped: true)]
-    public function post(?string $id = null): Response
+    public function post(?string $canvasSlug = null, ?string $id = null): Response
     {
         if ($this->template === null) {
             return $this->tpl->displayPartial('errors.error404');
--- a/app/Domain/Blueprints/Controllers/DelCanvas.php
+++ b/app/Domain/Blueprints/Controllers/DelCanvas.php
@@ -45,10 +45,11 @@ class DelCanvas
     /**
      * get - display the delete confirmation dialog.
      *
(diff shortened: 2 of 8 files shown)


marcelfolaron:
Thanks for the review — addressed in 63217c6:

- **DelCanvasItem `get()`/`post()` missing docblocks** — added PHPDoc documenting the `$canvasSlug` (route-prefix param, resolved in the constructor) and `$id` parameters, matching the other Blueprints controllers.
- **Regression test** — added `tests/Unit/app/Domain/Blueprints/Controllers/CanvasRouteBindingTest.php`. It loads the real `Blueprints/routes.php` and, for every action that takes an `{id?}` (showCanvas, editCanvasItem, editCanvasComment, boardDialog, delCanvas, delCanvasItem, export), asserts:
  - the real `{id}` lands in `$id` (not the slug),
  - an omitted `{id?}` leaves `$id` null (no slug leak),
  - structurally, each action declares `$canvasSlug` as its first parameter.

  It models the positional spread `ControllerDispatcher` uses (`...array_values($params)`) — the exact layer the bug lived in — and I verified it fails against the pre-fix single-param signature. No DB/session/browser needed, so it runs in the unit suite.

PR #3572, opened 2026-07-01
fix: resolve user role inheritance when adding user to project

aashir-ali:
I have updated the code for the to fix the "Inherit" issue


marcelfolaron:
Thanks, can you address the feedback from our code review agent. It's all valid

marcelfolaron:
Thanks for this — the diagnosis was spot on. Since this branch has maintainer-edits disabled we couldn't push directly, so we carried your fix forward in #3580 with your commit credited via co-author trailers. It addresses the review-agent feedback (numeric-only role storage + admin/owner guard on the write side, and treating empty/legacy-`0` as inherit on the read side without breaking the `ProjectsServiceTest` access-resolution test). We'll merge #3580 for the upcoming release and close this one in its favor. Much appreciated! 🙌

PR #3587, opened 2026-07-05
feat(tickets,sprints): cross-project (program) task views + inherited sprints

marcelfolaron:
## Summary

Adds **opt-in, additive** core support that lets a container project (a **program**, or a strategy) present the Tickets task views — **Table / Kanban / List** — across *all* of its child projects at once, and lets those child projects **inherit sprints** owned by the program. This is the core groundwork the PgmPro plugin builds on; nothing here activates unless a multi-project context is explicitly passed in.

**Every change is gated so existing per-project behavior is byte-identical when the new context is absent** — the `projects` criterion defaults empty, template branches are guarded (`isset($availableProjects)`, `$programBoard ?? false`, `$boardTabs ?? […]`, `$isKanbanView`, `empty($sprint->isInherited)`), and the new Sprints filter is a no-op with no listener registered.

## What's in it

**Tickets**
- `prepareTicketSearchArray` / `getAllBySearchCriteria` / `getAllMilestones`: a new comma-separated **`projects`** criterion that takes precedence over the single `currentProject`. A requested-but-empty set matches nothing (it never silently falls back to *all* accessible tickets).
- `getGroupByFieldOptions` + `getAllGrouped`: a **"project" group-by** dimension.
- `getProgramTicketTemplateAssignments()`: builds template assignments across a set of child projects — aggregated users/milestones, **per-project status labels**, and semantic status-type kanban columns.
- `setTicketStatusByType()` / `resolveProjectStatusKeyForType()`: a status change made from a cross-project board **always resolves to a status key that exists in the ticket's own project**, so a task can never be pushed into a status its project board doesn't recognize.
- `quickAddProgramTicket()`; optional `statusField` on `getStatusBreakdownBySwimlane()`.

**Sprints**
- `getAllSprints()` emits an `afterGettingAllSprints` filter so a plugin can append **inherited** sprints (no-op without a listener).
- `EditSprint` + `sprintdialog`: lock the project picker to the program when the sprint belongs to one. The Sprints model carries runtime `isInherited` / `ownerProgramId` flags (not persisted).

**Projects**
- `NewProject`: a project created from within a program/strategy is **nested under it by default**, and the parent is validated to be a *container* — so a project can never be nested under a regular project. The same validation is enforced in `ProjectService::addProject()` since that method is JSON-RPC reachable.

**Templates / JS**
- Guarded program-mode branches in `ticketFilter`, `ticketBoardTabs`, `showAll` / `showList` / `showKanban`, `quickadd-form`, `ticketHeader` / `timelineHeader`, `ticketDetails`; `ticketsController.js` forwards the `projects` filter. No change to per-project views.

## Compatibility
Fully backward compatible. Existing single-project task views, sprint management, and project creation are unchanged. The additions are dead code paths until a program/multi-project context opts in.

## Testing
- `php -l` clean on all changed PHP; `node --check` clean on the JS; full Blade tree compiles (`view:cache`); Laravel Pint passes on changed files.
- Manually verified: cross-project Table/Kanban/List render + filter/group by project; status drag on the program kanban writes a status valid in each ticket's own project; program sprints surface (badged) in child projects; per-project views unchanged.


marcelfolaron:
Thanks for the review — all four addressed in `9a187d944`:

1. **`ticketHeader.blade.php:56` — `$sprint` can be `false`.** Correct; `$currentSprint` can be a stale/deleted id that matches no entry, leaving `$sprint === false`. Guarded both headers (`ticketHeader` + `timelineHeader`) with `(! is_object($sprint) || empty($sprint->isInherited))`, which also preserves the original "show edit/delete" behavior in that stale state.

2. **`showKanban.blade.php:372` — unconditional plugin call.** Added feature detection — it now only calls `leantime.pgmProBoard.initProgramKanban(...)` when that function exists, and `console.warn`s otherwise instead of throwing.

3. **`Tickets.php` (annotate) — nested lookup can warn.** Switched to `isset($labels[$statusKey]['statusType']) ? … : 'NEW'` so a ticket with an undefined status key no longer emits array-offset warnings.

4. **`Tickets.php` (getProgramTicketTemplateAssignments) — N+1.** Milestones are now fetched in a **single** multi-project query via the new `projects` filter (removes the heavy per-project joined query). Users remain a per-project lookup — each is a single indexed query bounded by child count, and a correct bulk variant would have to replicate every project's `psettings`/client access rules, so I kept the per-project call (deduped) and documented the rationale inline.

`php -l`, Pint, and full Blade compilation pass.

marcelfolaron:
Round 2 addressed in `95d0864` (pushed as `e25ea782`):

- **`NewProject.php:99` — cleared parent was re-nested.** Good catch. Now only infers the container from the URL / current project when the `parent` field was **not** submitted (`isset($_POST['parent'])`); when the field is present, the user's explicit choice — including an intentional "none" — is respected, so the parent can actually be cleared.
- **`ticketFilter.blade.php:244` — placeholder quoting.** Switched to `@js(__('label.project'))` so a locale translation containing an apostrophe/quote can't break the script.
- **Quick-add project pickers (`quickadd-form`, `showAll`, `showList`) — no accessible name.** Added `aria-label="{{ __('label.project') }}"` to each required `<select>`.

On the re-flagged **N+1 in `getProgramTicketTemplateAssignments`**: milestones are already fetched in a single multi-project query (the heavy joined one). The remaining per-project loop is `getUsersAssignedToProject`, which is a single indexed lookup per child and — importantly — applies each project's own `psettings`/client access rules; a bulk version would have to faithfully replicate that per-project authorization, so I've kept the per-project call (deduped) and documented the rationale inline rather than risk an access-scoping regression. Happy to add a bulk `getUsersAssignedToProjects()` in a follow-up if you'd prefer.

`php -l`, Pint, and full Blade compilation pass locally.

PR #3592, opened 2026-07-05
fix: program board follow-ups — milestones, kanban render, status rollup clarity

marcelfolaron:
Two bugfixes surfaced while wiring up the program (cross-project) task board added in #3587. Both are small and additive.

## 1. `getAllMilestones()` honors the multi-project filter
The service-layer gate was `$searchCriteria['currentProject'] > 0` — it read `currentProject` unguarded and didn't understand the `projects` (multi-project) filter the program board uses. Called with a `projects`-only criteria array this (a) threw `Undefined array key "currentProject"` and (b) would otherwise have returned no milestones. Now proceeds when scoped to **either** a single project **or** a non-empty `projects` set, with the `currentProject` read null-coalesced.

## 2. Kanban swimlane header rendered as a component (not `Blade::render` on a string)
The swimlane header was rendered via:

```php
app('blade.compiler')::render('<x-global::kanban.swimlane-row-header .../>', $data)
```

When a `<x-component>` **string** is passed to `Blade::render` from **inside an already-compiled view**, the outer Blade pass pre-compiles the component tag into PHP that references bare variables (`$groupBy`, `$groupId`, `$label`, …). Those aren't in the inner render's scope, so it throws **`Undefined variable $label` / `$groupId`** for **any** kanban group-by. Per-project boards default `groupBy` to `all` and never hit the swimlane path, so it went unnoticed until the program board's "group by project".

Replaced with a plain `<x-...>` component tag using inline expressions — the standard idiom. Verified headlessly that the component renders identically.

Both fixes verified with `php -l`, Pint, and full Blade compilation. This also fixes kanban group-by on **regular** project boards (same latent bug).

### DIFF
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -1511,7 +1511,13 @@ class Tickets extends BaseService
     #[RequiresPermission(TicketsPermissions::VIEW)]
     public function getAllMilestones($searchCriteria, string $sortBy = 'standard'): array|false
     {
-        if (is_array($searchCriteria) && $searchCriteria['currentProject'] > 0) {
+        // Proceed when scoped to a single project (currentProject) OR to a set of projects
+        // (the multi-project `projects` filter used by program boards). The currentProject
+        // access is null-coalesced so a projects-only criteria array doesn't warn.
+        $isScoped = is_array($searchCriteria)
+            && ((($searchCriteria['currentProject'] ?? 0) > 0) || ! empty($searchCriteria['projects']));
+
+        if ($isScoped) {
             $items = $this->ticketRepository->getAllMilestones($searchCriteria, $sortBy);
 
             return $this->sortItemsHierarchically($items);
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -1502,14 +1502,16 @@ class Tickets extends BaseService
     /**
      * Retrieves all milestones based on the provided search criteria and sort option.
      *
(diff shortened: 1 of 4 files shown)


marcelfolaron:
Addressed in `22d15264`:

- **`getAllMilestones()` docs/signature drift** — PHPDoc now documents the `currentProject`-**or**-`projects` scoping, and the return type is `array` (the dead `false` is gone; the repo result is coalesced to `[]`).
- **Missing test coverage** — added two unit tests: a projects-only criteria array queries the repository (and asserts no `currentProject` key is required / no undefined-key warning), and an unscoped criteria returns `[]` without querying the repo. Full `TicketsServiceTest` green (22 tests).
- **Info icon `aria-hidden`** — added `aria-hidden="true"` on the board-note icon.

Plus one UX change from testing feedback: the per-card **status label on the program kanban is now a dropdown** of the task's own project statuses, so you can set the detailed status (e.g. "Blocked") directly from the card — it writes a key valid in that project (orphan-safe) and updates live. `initStatusDropdown()` is wired for the program board only.

marcelfolaron:
These three are already resolved on the current tip (`22d15264`) — they're re-posts of the previous round's comments:

- **`getAllMilestones()` docs/return** — the PHPDoc now documents the `currentProject`-**or**-`projects` scoping and `@return array`, and the signature is `: array` (the `false` was removed). See lines ~1502–1513.
- **Test coverage** — the scoping tests were added in `22d15264` (`test_get_all_milestones_scopes_by_projects_without_current_project` + `..._unscoped_returns_empty_and_skips_repository`); that comment is pinned to `495a625d`, the commit before the tests landed.
- **Info icon `aria-hidden`** — present at showKanban line 49: `<i class="fa fa-circle-info" aria-hidden="true"></i>`.

All CI checks are green on this commit. No code change needed.

PR #3629, opened 2026-07-09
fix(editor): delete empty lines with Backspace in the tiptap editor

jeppekroghitk:
## Problem

Pressing Backspace on an empty line in the tiptap editor did nothing. Only Option/Alt+Backspace
worked.

## Cause

`ColumnLayout`'s keyboard handlers were regular `function () { this.editor… }`.
Tiptap calls shortcut handlers bare and the webpack bundle is strict-mode, so
`this` was `undefined` — every Backspace threw before the core keymap's
`joinBackward` could run, falling back to native deletion that can't remove an
empty paragraph. Option+Backspace worked only because columns doesn't bind it.

## Fix

Switch the three handlers to arrow functions so `this` resolves to the
extension context. One-file change; verified in a real (headless Chromium,
strict-mode) harness that Backspace now deletes and holds across `<p></p>`,
`<p><br></p>`, and `text<br><br>text`.

### Link to ticket

https://github.com/Leantime/leantime/issues/3630

### Type

- [ ] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result


marcelfolaron:
Thank you!

PR #3637, opened 2026-07-12
feat(oidc): mobile SSO bridge — one-time-code → bearer token exchange (generic OIDC)

gloriafolaron:
**For review — please don't merge until you've looked it over.** This is the core backend half of mobile SSO, scoped to **generic OIDC in core**. Pairs with the gated mobile client (`leantime-mobile` #115); the assumptions being validated are noted inline below.

## What it does
Lets the mobile app sign in via the instance's existing generic OIDC, provider-agnostically. The app opens `/oidc/login` in the system auth browser → on success the callback mints a **single-use code** and redirects to the **app scheme** → the app trades the code (+ PKCE verifier) for a bearer token. The token never travels in a URL.

```
app → /oidc/login?mobile=1&redirect_uri=leantime://oidc-callback&code_challenge=…
    → provider → Oidc::callback → Oidc::login (setUserSession)
    → mobile-origin? mint OTC (bound to code_challenge), 302 leantime://oidc-callback?code=OTC
app → POST /oidc/mobile/exchange {code, code_verifier} → {token, user}
```

## Changes
- **`Oidc\Services\Oidc::buildLoginUrl(mobile, redirectUri, codeChallenge)`** — stashes the mobile flag + app-scheme redirect + PKCE challenge in the session, riding the round-trip alongside `oidc.state`.
- **`Oidc\Services\Oidc::login()`** — after `setUserSession`, a mobile-origin flow mints a one-time code and redirects to the app instead of `/dashboard/home`.
- **`Oidc\Services\Oidc::isAllowedMobileRedirect()`** — `leantime://` prefix only; also rejects whitespace and control chars (they'd otherwise break the `Location` header).
- **`Oidc\Controllers\Login`** — threads `?mobile=1&redirect_uri&code_challenge`; requires `code_challenge` on mobile flows and validates the PKCE S256 charset/length; fixes a pre-existing double `buildLoginUrl()` call.
- **`Oidc\Controllers\Mobile::exchange()`** (`POST /oidc/mobile/exchange`) — **POST-only** (405 on GET). Peeks the code non-destructively → validates PKCE (rejects `invalid_verifier` **without** burning the code, so a scheme-hijacker can't DoS the legit exchange) → confirms the user still exists → **then** consumes the code + mints via `AccessTokenRepository` directly (no session — the single-use code + verifier are the authorization). Returns `{token, user}` with **only safe identity fields** (no password hash / 2FA seed / session — cf. the `getUser` credential-dump incident).
- **`Oidc\Services\OidcMobileCode`** — sha256-hashed, ≤60s, single-use codes in the **default cache store** (self-expiring, no sweep needed, no schema change). Split into `peekCode()` (non-destructive read) + `consumeCode()` (idempotent forget) so the controller can validate-before-burn.
  > **Deployment note:** on a multi-node install the cache store must be shared across app nodes (e.g. redis) — the code is written by the callback request and read by a later exchange request that may land on a different node. The per-node file store (`installation`) is fine for a single-node install but would miss on multi-node.
- **`AuthCheck`** — allow-lists `oidc.mobile` (the exchange arrives pre-session).
- **Tests** — `MobileTest` covers POST-only, missing/invalid code, invalid-verifier-doesn't-burn, missing-user-doesn't-mint, and the valid path (consumes + mints + returns only safe fields). `AuthCheckTest` asserts `oidc.mobile.*` is public.

## Please sanity-check (my assumptions)
1. **Routing** — I'm relying on `/oidc/mobile/exchange` → `Oidc\Controllers\Mobile::exchange()` (segment-2 = method). Is that right, or do you want a different route shape?
2. **Session survival** — I assume `oidc.mobile` / `oidc.mobileRedirect` / `oidc.mobileChallenge` persist across the provider round-trip the same way `oidc.state` does, and survive `setUserSession`. Worth a look.
3. **Minting via the repo** directly (bypassing `AccessToken::createToken`'s session gate) for the cookieless exchange — intended, but flagging.

## Not in this PR (follow-ups)
- General `/status` discovery endpoint — the mobile client already calls it (best-effort), so SSO stays dormant until it ships.
- The **AdvancedAuth provider-callback twin** reusing this OTC store + exchange.


marcelfolaron:
Thanks for this, @gloriafolaron — triaged in today's daily sweep.

**Status:** needs-review · **Priority:** P1 · **Label:** Feature · **Owner:** @marcelfolaron / @broskees (human review only)

This touches the auth/token path (OTC minting, `AccessTokenRepository` direct mint bypassing the session gate, `AuthCheck` allow-listing `oidc.mobile`), so per our security-first policy it is **not** eligible for automated review or autofix — it needs a human maintainer to work through the four sanity-check questions you flagged (route shape, fresh-install `SchemaBuilder`, session survival across the round-trip, and the repo-direct mint). Migration 30521 / `dbVersion` → 3.5.21 also wants a look on the upgrade path.

**Next action:** maintainer security review of the exchange flow + migration before this advances. Flagging the token-shape guard against the prior `getUser` credential-dump as the highest-value thing to confirm. Not marking ready-for-review; the review gate stays with the maintainers.

gloriafolaron:
Thanks @Copilot — addressed the actionable items in the latest commit:

- ✅ **User validated before minting** — the exchange now fetches/validates the user first and returns `invalid_user` (401) if absent, so a deleted user can't leave an orphaned token row.
- ✅ **`code_challenge` validated** — Login rejects a present-but-malformed challenge (PKCE S256: base64url, 43–128 chars, RFC 7636).
- ✅ **Dead doc links removed** — both references to `docs/backend-mobile-auth-bridge-plan.md` are gone.

Already handled (the review looked at an earlier state):
- **Replay race** — the code store is **cache-backed** and uses `Cache::pull` (atomic read-and-delete). There is no `SELECT ... first()` + `DELETE`; a replayed code can't be exchanged twice.
- **`php://input`** — the exchange reads the framework's merged request params (`stringParam($params, …)`), not the raw stream.
- **AuthCheck allow-list** — covered by `test_oidc_mobile_exchange_is_a_public_route` in `AuthCheckTest`.
- **Nullable `expires_at` migration** — n/a; the store is cache-backed, there's no `zp_oidc_mobile_codes` table/migration. (PR description note: implementation is **cache-backed**, not DB-backed — apologies for the stale description.)

Design note on **consume-before-PKCE**: consuming atomically (`Cache::pull`) *before* PKCE validation is deliberate — it keeps the single-use guarantee with no peek-then-consume replay window. The tradeoff is that a wrong/missing verifier burns the code, but that's a minor self-inflicted DoS (the app just restarts the short-lived flow) and is the safer default vs. reopening a replay window. Happy to switch to peek→validate→consume if preferred.

marcelfolaron:
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `f4bc7c2` (new commit since the 07-13 triage note)

**1. Intent:** Adds a mobile SSO bridge over the instance's existing generic OIDC — the app opens `/oidc/login`, the callback mints a single-use code redirected to the `leantime://` app scheme, and `POST /oidc/mobile/exchange` trades that code (now PKCE-protected) for a bearer token, so the token never travels in a URL.

**2. Change requests** (the latest commit `f4bc7c2` addressed the earlier Copilot/PKCE items — validate-user-before-mint, `code_challenge` format check, dead-doc-link removal — so these are the remaining confirm-before-merge items; **this touches auth/token minting, so per our security-first policy the deep verification is a human maintainer gate, not automated approval**):
1. **PKCE consume-before-validate tradeoff — confirm the decision is recorded.** The exchange consumes the code atomically (`Cache::pull`) *before* validating the `code_verifier`. That correctly closes the replay window, but a wrong/missing verifier burns the code (minor self-inflicted DoS — the app just restarts the short-lived flow). The author flagged this explicitly and offered peek→validate→consume as the alternative. Pick one and note the rationale in the PR so it isn't re-litigated later.
2. **Token response field allow-list — the highest-value thing to confirm.** `Oidc\Controllers\Mobile::exchange()` must return **only** safe identity fields (no password hash / 2FA seed / session), per the prior `getUser` credential-dump incident. Verify the returned `{user}` shape is an explicit allow-list, not a full model/`toArray()` that could regress if the user model gains a sensitive field later.
3. **Session survival across the provider round-trip.** `oidc.mobile` / `oidc.mobileRedirect` must persist alongside `oidc.state` through the provider redirect **and** survive `setUserSession`. Confirm empirically against a live provider — the flow silently falls back to the web dashboard if these are lost, which would look like "mobile login just doesn't work" rather than an error.
4. **`redirect_uri` scheme allow-list.** Only `leantime://` is honored per the description — confirm the check rejects any other scheme (no `http(s)://`, no open-redirect) so the minted code can't be redirected to an attacker-controlled target.

- [ ] **Human security review** of the exchange flow signs off on: consume-before-validate (CR #1), the token field allow-list (CR #2), and the `redirect_uri` scheme guard (CR #4). This is the security gate — automation is explicitly *not* approving this PR.
- [ ] A test (or documented manual procedure against a live OIDC provider) covering the core behavior change: full round-trip login → code → exchange → token, **plus** the negative paths (missing/mismatched verifier rejected, non-`leantime://` redirect rejected, deleted-user → `invalid_user` 401). The `AuthCheckTest` public-route test is present; the end-to-end flow is not yet runtime-tested per the PR body.
- [ ] Back-compat confirmed: the codes are now **cache-backed** (the `zp_oidc_mobile_codes` table + migration 30521 were dropped and `dbVersion` reverted to 3.5.20), so no schema change ships — verify the PR description is updated to match (it still references the DB-backed table/migration in places).

**4. CI status:** ⚠️ **High priority — behavior change on an auth path, not yet runtime-tested end-to-end.** The PR states `php -l` + Pint clean and adds the `AuthCheck` route test, but the exchange flow itself is untested against a real provider (author acknowledges this). Per our test gate, treat the missing end-to-end coverage on the token-exchange path as a blocker until either a test or a documented manual run (with the negative cases above) is recorded here. Confirm Pint + PHPStan level 5 + full unit suite are green on `f4bc7c2` in the UI before merge — I can't read check-run status this cycle.

_Advisory review only — not an approval, and I am not marking this ready or merging. The security review + merge stay with @marcelfolaron / @broskees._

marcelfolaron:
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `7b71689` (new commit since my 07-13 review at `f4bc7c2`)

**1. Intent:** Mobile SSO bridge over the instance's generic OIDC — `/oidc/login` → callback mints a single-use, PKCE-bound code → redirect to the `leantime://` app scheme → `POST /oidc/mobile/exchange` trades code + verifier for a bearer token; the token never travels in a URL.

**Prior review — the new commit `7b71689` directly addressed my open items** ✅. From the commit + Copilot follow-up:
- **CR #1 (consume-before-validate) resolved the safer way.** `OidcMobileCode` is now split into `peekCode()` (non-destructive) + `consumeCode()` (idempotent forget); `exchange()` peeks → validates PKCE + user existence → **then** burns the code. This closes the self-inflicted-DoS where a scheme-hijacker's bad verifier would burn the legit code before the app's own exchange, **without** reopening the replay window (consume is still atomic once reached). Good — this supersedes the earlier "pick one and document" ask.
- **POST-only** — `exchange()` now returns 405 + `Allow` header on GET, keeping secrets out of query strings / access logs.
- **`code_challenge` now required on `mobile=1`** — the old flow could mint a code bound to an empty challenge whose exchange could never succeed.
- **`isAllowedMobileRedirect()` rejects whitespace/control chars** — closes the `Location`-header injection / Symfony-500 vector (CR #4 hardened).

**2. Remaining change requests** — **this is an auth/token-minting path, so per security-first policy the deep sign-off is a human-maintainer gate, not automated approval:**
1. **Token response field allow-list — still the highest-value thing to confirm** (my prior CR #2, unchanged). The commit message says the valid path "returns only safe user fields," and `MobileTest` now asserts it. Confirm the returned `{user}` is an **explicit allow-list of identity fields**, not `$user` / `toArray()` — so a future column added to the user model (2FA seed, hash) can't silently leak, per the prior `getUser` credential-dump incident.
2. **PKCE peek→validate→consume atomicity.** With the split, verify there's no TOCTOU gap: two concurrent exchanges of the same code should not both pass peek before one consumes. `Cache::pull` in `consumeCode()` is atomic, so the loser gets null at consume even if both peeked — confirm `exchange()` treats a null/failed `consumeCode()` as a hard failure (doesn't mint after a lost race).
3. **Session survival across the provider round-trip** (prior CR #3, still confirm-empirically). `oidc.mobile` / `oidc.mobileRedirect` / `oidc.mobileChallenge` must persist alongside `oidc.state` through the redirect **and** survive `setUserSession` — a live-provider run is the only real check.

- [ ] **Human security review** signs off on: the token field allow-list (CR #1), the peek/consume race (CR #2), and the `leantime://`-only redirect guard. Automation is explicitly *not* approving this.
- [ ] Behavior change **now ships tests** ✔ — `MobileTest` covers method-not-allowed (405), missing/invalid code, invalid-verifier-doesn't-burn-code, missing-user-doesn't-mint, and the valid path (consumes + mints + only safe fields); `AuthCheckTest` covers the public route. Remaining gap: a **live-provider end-to-end** run (or documented manual procedure) for the full login→code→exchange→token round-trip, since the unit tests mock the provider. Record that run here.
- [ ] Back-compat confirmed: cache-backed codes, **no schema change** (`zp_oidc_mobile_codes` table + migration 30521 dropped, `dbVersion` back to 3.5.20). Confirm the PR description is fully reconciled to the cache-backed implementation (earlier revisions still referenced the DB table in places).

**4. CI status:** ⚠️ **High priority — auth-path behavior change.** The unit-test gap from my prior review is now **cleared** (MobileTest covers the exchange incl. negative paths), so this is confirm-green + human-security-sign-off rather than a coverage blocker. The one residual is the live-provider round-trip (checklist item 2). Confirm Pint + PHPStan level 5 + full unit suite green on `7b71689` in the UI — I can't read check-run status this cycle; treat any red/skipped job as blocking.

_Advisory review only — not an approval, and I am not marking this ready or merging. The security review + merge stay with @marcelfolaron / @broskees._

marcelfolaron:

**Status:** needs-changes (security-sensitive) · **Priority:** P1 · **Next action:** address token-scope + rate-limit + multi-node items below · **Owner:** @marcelfolaron / @broskees

**Intent:** Adds a mobile SSO bridge over generic OIDC — the app opens `/oidc/login?mobile=1` with a PKCE `code_challenge`, the callback mints a single-use one-time code bound to that challenge, and `POST /oidc/mobile/exchange` trades `{code, code_verifier}` for a bearer token, so the token never travels in a URL.

The security architecture is careful and I want to acknowledge it: peek-before-burn so a scheme-hijacker can't DoS the legit exchange, `hash_equals` PKCE S256 compare, POST-only with 405 on GET, user-still-exists check before mint, and a `safeUser()` allow-list that avoids the `getUser` credential-dump class of bug. Comments below are the hardening gaps, not objections to the shape.

### Change requests
1. **Token scope / lifetime — the biggest open question** (`Oidc/Controllers/Mobile.php`, `exchange()`: `$this->tokens->createToken($userId, 'mobile-sso')`). What scope and TTL does an `AccessTokenRepository::createToken()` bearer carry? A mobile SSO token is long-lived on a device; confirm it is **not** a full-privilege, non-expiring API token, and that it can be **revoked** (device lost). If `createToken` has no expiry/scope column, that's a blocker for a mobile auth token — please document the lifecycle or add one.
2. **No rate limit on `/oidc/mobile/exchange`** (`Mobile::exchange()`). The endpoint is public (allow-listed `oidc.mobile` in `AuthCheck`) and returns distinct 400/401 codes. Even with ≤60s single-use codes, an unauthenticated code-guessing / verifier-brute endpoint should sit behind the existing `RateLimiter` middleware. Confirm it applies to allow-listed actions, or add an explicit throttle.
3. **Multi-node cache correctness is a deploy-time footgun** (`OidcMobileCode`, default cache store). The PR body flags it, but since the per-node file store is the default on self-hosted (`installation`), a multi-node install will *silently* fail exchange (code written on node A, read on node B → `invalid_code`). Please (a) document this in the OIDC/mobile setup docs, and ideally (b) log a warning when the resolved cache store is a per-node file driver.
4. **Open-redirect surface on `isAllowedMobileRedirect()`** (`Oidc/Services/Oidc.php`). Diff shows `leantime://` prefix-only + whitespace/control-char rejection — good. Confirm it's an exact scheme match (`leantime://`) and not a substring/`str_contains`, so `https://evil/leantime://` can't pass. A unit test pinning the reject cases would lock this down.
5. **`Login::get()` PKCE validation is duplicated with the service** (`Oidc/Controllers/Login.php` regex `^[A-Za-z0-9\-_]{43,128}$` vs. `pkceMatches()` in the controller). Fine to keep the front-door guard, but consider centralizing the S256 charset/length check so the two can't drift.

- [ ] CI green: **Pint + PHPStan level 5 + full unit suite** (incl. the new `MobileTest` + `AuthCheckTest`).
- [ ] Documented token **scope, TTL, and revocation** for the `mobile-sso` bearer, plus a rate-limit covering `/oidc/mobile/exchange`.
- [ ] Multi-node cache requirement documented (and/or a warning when a per-node store is active); `isAllowedMobileRedirect` reject-cases covered by a test.

### CI status
Tests are present and target the security-critical branches (POST-only, missing/invalid code, invalid-verifier-doesn't-burn, missing-user-doesn't-mint, safe-fields-only) — good, not a missing-test blocker. Confirm they're **green** at head, and confirm PHPStan level 5 passes on the new `Mobile` controller. This is auth code; I'd treat any skipped/flaky check here as a hard blocker.

_Advisory only — not an approval. Given this is authentication code, a second human reviewer (@broskees) before merge is warranted. Merge authority stays with @marcelfolaron / @broskees._

marcelfolaron:
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `7b71689` (advanced past `f4bc7c2` from my 07-13 19:11 review)

**1. Intent:** Mobile SSO bridge over the instance's existing generic OIDC — `/oidc/login?mobile=1&redirect_uri=leantime://…&code_challenge=…` mints a single-use, PKCE-bound one-time code redirected to the app scheme, and `POST /oidc/mobile/exchange` trades `{code, code_verifier}` for a bearer token, so the token never travels in a URL.

**What changed since my last review** ✅ — the new commits reworked the exchange from *consume-before-validate* to an explicit **peek → PKCE → user-exists → consume-then-mint** flow: `OidcMobileCode::peekCode()` (non-destructive) + `consumeCode()` (idempotent forget), `Mobile::exchange()` now returns `invalid_verifier` **without burning the code** (closes the scheme-hijacker DoS), enforces **POST-only** (405 on GET), and confirms the user still exists before minting. The token response is an explicit allow-list (`id/firstname/lastname/username`) via `safeUser()`.

**2. Change requests** (this touches auth/token minting → per security-first policy the deep verification is a **human maintainer gate**, not automated approval; these are the confirm-before-merge items):
1. **⛔ Migration-number collision with #3643 — confirm #3637 truly dropped its migration.** The `[PATCH 1/4]` history and PR description both reference `update_sql_30521` + `dbVersion` → 3.5.21 + a `zp_oidc_mobile_codes` table, but the current HEAD diff shows the store is **cache-backed** (`Cache::pull`, no table, no migration) and commit `d0ee571` says "dbVersion reverted to 3.5.20." Meanwhile **#3643 also defines `update_sql_30521` (3.5.21)** for `zp_goal_history`. Verify this PR's HEAD carries **no** `30521`/`dbVersion` bump — if any stale migration remains, it collides with #3643 and one silently won't run on upgrade. Also update the PR body (it still describes the DB-backed table) so a reviewer isn't verifying a store that no longer exists.
2. **Peek→consume replay window — confirm the store's `peekCode`+`consumeCode` is race-safe.** Moving from atomic `Cache::pull` (read-and-delete) to `peekCode()` (read) then a later `consumeCode()` reopens a theoretical window where two concurrent exchanges both peek the same valid code before either consumes it. With PKCE this is low-severity (both need the verifier), but confirm `consumeCode()` is atomic/idempotent (e.g. `Cache::pull` inside it, or a compare-and-delete) so a double-exchange can't mint two tokens. This is the one thing the peek/consume split must get right.
3. **Multi-node cache requirement — surface it as an install constraint, not just a PR note.** The code is written by the callback request and read by a later exchange that may hit a different node; the per-node file store (`installation` default) will miss on multi-node. The PR flags this, but a wrong verifier on a cache miss returns `invalid_code` with no diagnostic — consider a one-line note in the OIDC/mobile docs (or a startup warning when a non-shared cache store is configured) so a multi-node operator isn't left debugging silent SSO failures.
4. **`isAllowedMobileRedirect()` scheme check — confirm prefix-match can't be bypassed.** It honors `leantime://` prefix only and rejects whitespace/control chars. Confirm the check is a true scheme match (not `str_contains`) so `https://evil/?x=leantime://` can't slip through, and that the value is validated **before** it reaches the `Location` header on the callback redirect.

1. [ ] **Token response is an explicit allow-list** (no password hash / 2FA seed / session / reset token), asserted by a test — the highest-value guard against the prior `getUser` credential-dump class. `safeUser()` reads correct; lock it with the test so a future User-model field can't regress it.
2. [ ] **`MobileTest` covers the security-critical paths** and is green in CI: POST-only (405 on GET), missing/invalid code, invalid-verifier-does-not-burn-the-code, missing-user-does-not-mint, and the happy path (consumes + mints + returns only safe fields). `AuthCheckTest` asserts `oidc.mobile.*` is public. This is an auth behavior change — the test suite is the gate, not optional.
3. [ ] **Migration ownership + backward-compat resolved** (CR #1): exactly one open PR owns `30521`; PR body matches the cache-backed reality; no schema change means no upgrade migration to verify here — confirm that's actually the case at HEAD.

**4. CI status:** ⚠️ Can't read check-run status this cycle — confirm Pint + PHPStan level 5 + full unit suite (incl. `MobileTest`/`AuthCheckTest`) are green at `7b71689` in the UI; treat any red/skipped Pint, PHPStan, or auth-test job as **blocking**. End-to-end SSO isn't runtime-verified (needs a live OIDC provider) — the flow logic + the isolation tests are the review target, and the security items above stay a human gate.

PR #3638, opened 2026-07-12
fix(projects): getProjectsByUserActivity excludes archived projects (projectStatus 'open')

gloriafolaron:
**For review.** One-liner Marcel identified.

`getProjectsByUserActivity` calls `getUserProjects` with `accessStatus: 'all'` but leaves `projectStatus` at its default `'all'` — so it returns **active AND archived** projects. "Projects I'm actively working in" shouldn't include archived ones, and they were padding the **mobile project picker** (a user saw ~60 extra entries under "+N more").

```php
// Domain/Projects/Services/Projects.php  (getProjectsByUserActivity)
- getUserProjects(userId: $userId, accessStatus: 'all');
+ getUserProjects(userId: $userId, projectStatus: 'open', accessStatus: 'all');
```

`projectStatus: 'open'` applies `project.state <> -1` at the query level (verified in `getUserProjects`), excluding archived — the same default every other project-list method (`getProjectsAssignedToUser`, the hierarchy methods) already uses. Also benefits web callers of this method.

**Scope:** only `getProjectsByUserActivity` (the mobile-oriented, activity-sorted method). Left `getProjectsUserHasAccessTo` untouched (more general, potentially web-relied-upon) — the mobile client keeps a defensive `state === -1` filter that covers the fallback path and older self-hosted backends that predate this fix.

`php -l` + Pint clean.

### DIFF
--- a/app/Domain/Projects/Services/Projects.php
+++ b/app/Domain/Projects/Services/Projects.php
@@ -1207,7 +1207,11 @@ class Projects extends BaseService implements ChecksProjectAccess
             return false;
         }
 
-        $projects = $this->projectRepository->getUserProjects(userId: $userId, accessStatus: 'all');
+        // projectStatus 'open' excludes archived projects (state === -1) at the
+        // query level — "projects I'm actively working in" shouldn't surface
+        // archived ones. Without it this defaulted to 'all' and returned
+        // archived projects too (they were padding the mobile project picker).
+        $projects = $this->projectRepository->getUserProjects(userId: $userId, projectStatus: 'open', accessStatus: 'all');
         if (! $projects) {
             return false;
         }


marcelfolaron:
Thanks @gloriafolaron — triaged in today's daily sweep.

**Status:** needs-review · **Priority:** P2 · **Label:** Bug · **Owner:** @marcelfolaron / @broskees

Clean, localized fix: `getProjectsByUserActivity` was leaving `projectStatus` at its `'all'` default, so archived projects leaked into the "actively working in" set and padded the mobile project picker. Scope looks contained to the one method.

**Next action:** maintainer review. One acceptance item worth confirming before merge — a regression test (or documented manual check) that the picker/query returns only non-archived projects, since this is a behavior change. Pairs conceptually with #3639 (same archived-exclusion class, for tickets).

marcelfolaron:
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees) · reviewed at head `dd2dd03`

**1. Intent:** Makes `getProjectsByUserActivity()` pass `projectStatus: 'open'` so archived projects (`state === -1`) are excluded at the query level, stopping archived projects from padding the "projects I'm actively working in" set (the mobile project picker was showing ~60 extra entries under "+N more").

**2. Change requests** (diff is a clean +5/−1 one-liner at HEAD `dd2dd03`; these are confirm-before-merge, not rewrites):
1. `app/Domain/Projects/Services/Projects.php:1210` — the call now reads `getUserProjects(userId: $userId, projectStatus: 'open', accessStatus: 'all')`. Confirm `projectStatus: 'open'` maps to the `state <> -1` filter in `getUserProjects` (the PR body says it's verified) and that `'open'` is the exact literal the repository switches on — a typo'd status string would silently fall through to no filtering rather than error.
2. **Consistency with #3639.** This is the project-side twin of #3639 (archived-ticket exclusion). #3639 landed its exclusion as an explicit `$excludeClosedProjects` SQL flag with the `state <> -1 OR state IS NULL` pattern; this one rides the existing `projectStatus` param. Both are fine, but confirm `projectStatus: 'open'` also handles the **NULL-state** case the way #3639 does (a project row with `state IS NULL` shouldn't be dropped as if archived) — verify against the `getUserProjects` WHERE construction.
3. **Behavior-change surface.** The PR notes this "also benefits web callers of this method." Confirm no web screen relies on `getProjectsByUserActivity()` *including* archived projects (e.g. an "all my projects incl. archived" view) — the scope note says only the mobile-oriented activity method changes, but a quick grep of the method's callers before merge closes that off.

1. [ ] A user with both active and archived projects sees only active ones in the activity-sorted picker; the previously-inflated "+N more" count drops accordingly.
2. [ ] **Test or documented manual check covering the behavior change** — this is a read-path behavior change and the diff ships **no test** (+5/−1, code only). Per our test gate that is the merge blocker: either a regression test asserting `getProjectsByUserActivity()` excludes `state === -1` projects, or a documented manual verification recorded here. (Its twin #3639 added exactly such a test in its latest commit — mirror that here.)
3. [ ] No schema/API change (query-param-only) confirmed — no migration note needed.

**4. CI status:** ⚠️ **Missing test on a behavior change — treat as blocker per repo policy** (see checklist item 2). I can't read check-run status through my current tooling this cycle, so confirm Pint + PHPStan level 5 + full unit suite are green on `dd2dd03` in the UI before merge and treat any red/skipped Pint or PHPStan job as blocking.

marcelfolaron:
**Maintainer review** (advisory — not an approval; merge stays with @marcelfolaron / @broskees) · head `4868900` (@ 2026-07-13)

**Intent:** Excludes archived projects (`state === -1`) from `getProjectsByUserActivity()` by passing `projectStatus: 'open'` to `getUserProjects()`, so the mobile project picker stops listing archived projects.

**Change requests:**
1. None blocking — the one-line change is correct and scoped. `Projects.php:1210` now matches the archived-exclusion default every other project-list method already uses. ✅
2. *(minor, non-blocking)* No automated test accompanies this behavior change. Given the sibling PR #3639 added a service-level test capturing the flag it passes, consider mirroring that here — a test asserting `getProjectsByUserActivity()` calls `getUserProjects` with `projectStatus: 'open'` guards against a future default-arg regression. Not a merge blocker for a one-liner, but cheap insurance.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite.
- [ ] Manual: as a user with ≥1 archived project, confirm the mobile project picker (+N more) no longer lists archived projects, and the web callers of this method are unaffected.
- [ ] No API/schema change → no migration/back-compat note required (confirmed: query-param only).

**CI status:** Not yet verified from here (checks tab). Since this is a behavior change with no test in the diff, treat "unit suite green" as necessary but not sufficient — the manual check above is the real gate. Not a blocker.

marcelfolaron:

**Status:** needs-test · **Priority:** P2 · **Next action:** add a regression test (or documented manual proc), confirm CI green · **Owner:** @marcelfolaron / @broskees

**Intent:** `getProjectsByUserActivity()` now passes `projectStatus: 'open'` to `getUserProjects()` so archived projects (`state = -1`) stop padding the mobile project picker.

### Change requests
1. **Missing test on a behavior change — blocker per our review policy** (`app/Domain/Projects/Services/Projects.php:1210`). The sibling ticket fix #3639 added `test_get_all_open_user_tickets_excludes_closed_projects_at_query_level`; this one changes returned rows with no test. Add an equivalent that asserts `getUserProjects` is called with `projectStatus: 'open'` (or that archived projects are excluded from the result).
2. **Confirm `'open'` semantics match intent** (`getUserProjects`). The comment says `'open'` applies `state <> -1`. Verify `'open'` doesn't *also* filter on some other status dimension (e.g. a separate "closed/completed" flag) that would drop projects the user still considers active. If `'open'` means only "not archived," good; if it's narrower, this could hide legitimately-active projects.
3. **Web-caller blast radius** (`getProjectsByUserActivity` other consumers). PR says web callers also benefit — confirm no dashboard/widget relies on this method *including* archived projects. A quick grep of callers in the review notes would close this out.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite.
- [ ] A unit test (or documented manual procedure) proving archived projects are excluded from `getProjectsByUserActivity`.
- [ ] No schema/API change (confirmed — pure query-scope change, backward-compatible for the method's contract).

### CI status
No CI signal read at head. **Missing test is the gating item** — this is a data-visibility behavior change and should not merge without the coverage in item 1, especially since its sibling #3639 set the precedent.

PR #3639, opened 2026-07-12
fix(tickets): exclude archived-project tickets from getAllOpenUserTickets

gloriafolaron:
**For review.** Same class as the archived-projects picker fix — but for tickets.

`getAllOpenUserTickets` returned open tickets from **every** project the user can access, including **archived** ones. "My open tickets" shouldn't surface work from projects that are no longer active, and they were inflating the mobile task list (a user saw 111 "Later" tasks; archived-project tickets were part of that).

- `simpleTicketQuery` now selects `zp_projects.state as projectState` (a harmless extra column for its other two callers).
- `getAllOpenUserTickets` skips tickets whose `projectState === -1` (archived).

**Scoped:** the exclusion lives in `getAllOpenUserTickets` only; `simpleTicketQuery`'s filtering is unchanged for its other callers, so nothing else shifts.

`php -l` + Pint clean.

### DIFF
--- a/app/Domain/Tickets/Repositories/Tickets.php
+++ b/app/Domain/Tickets/Repositories/Tickets.php
@@ -738,6 +738,10 @@ class Tickets
                 'milestones.headline as milestoneHeadline',
                 'zp_projects.name as projectName',
                 'zp_projects.details as projectDescription',
+                // Lets callers drop tickets that belong to archived projects
+                // (state === -1) without a second query. getAllOpenUserTickets
+                // uses it; other callers just get a harmless extra column.
+                'zp_projects.state as projectState',
             ])
             ->selectRaw("CASE WHEN zp_tickets.type <> '' THEN zp_tickets.type ELSE 'task' END AS type")
             ->leftJoin('zp_projects', 'zp_tickets.projectId', '=', 'zp_projects.id')
--- a/app/Domain/Tickets/Repositories/Tickets.php
+++ b/app/Domain/Tickets/Repositories/Tickets.php
@@ -708,7 +708,7 @@ class Tickets
         return array_map(fn ($item) => (array) $item, $results->toArray());
     }
 
-    public function simpleTicketQuery(?int $userId, ?int $projectId, array $types = []): array|false
+    public function simpleTicketQuery(?int $userId, ?int $projectId, array $types = [], bool $excludeClosedProjects = false): array|false
     {
         $requestorId = session()->exists('userdata') ? session('userdata.id') : -1;
         $clientId = session('userdata.clientId') ?? '-1';
@@ -738,10 +738,6 @@ class Tickets
                 'milestones.headline as milestoneHeadline',
                 'zp_projects.name as projectName',
                 'zp_projects.details as projectDescription',
-                // Lets callers drop tickets that belong to archived projects
-                // (state === -1) without a second query. getAllOpenUserTickets
-                // uses it; other callers just get a harmless extra column.
-                'zp_projects.state as projectState',
             ])
             ->selectRaw("CASE WHEN zp_tickets.type <> '' THEN zp_tickets.type ELSE 'task' END AS type")
             ->leftJoin('zp_projects', 'zp_tickets.projectId', '=', 'zp_projects.id')
@@ -784,6 +780,17 @@ class Tickets
 
         if (count($types) > 0) {
             $query->whereIn('zp_tickets.type', $types);
+        }
+
+        // Closed projects (state === -1) are inactive. Callers wanting a "my
+        // active work" view drop their tickets at the SQL level so closed-
+        // project rows are never fetched or returned — matches the existing
+        // `state <> -1 OR state IS NULL` pattern used elsewhere in this repo.
+        if ($excludeClosedProjects) {
+            $query->where(function ($q) {
+                $q->where('zp_projects.state', '<>', -1)
+                    ->orWhereNull('zp_projects.state');
+            });
         }
 
         $results = $query->orderByDesc('zp_tickets.dateToFinish')
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -555,7 +555,7 @@ class Tickets extends BaseService
         // Exclude closed projects (state === -1) at the SQL level — "My open
         // tickets" shouldn't surface work from projects that are no longer
         // active (they were padding the mobile task list).
-        $tickets = $this->ticketRepository->simpleTicketQuery($userId, $project, [], true);
+        $tickets = $this->ticketRepository->simpleTicketQuery($userId, $project, excludeClosedProjects: true);
 
         $ticketArray = [];
 
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -552,7 +552,10 @@ class Tickets extends BaseService
     public function getAllOpenUserTickets(?int $userId = null, ?int $project = null): array
     {
 
-        $tickets = $this->ticketRepository->simpleTicketQuery($userId, $project);
+        // Exclude closed projects (state === -1) at the SQL level — "My open
+        // tickets" shouldn't surface work from projects that are no longer
+        // active (they were padding the mobile task list).
+        $tickets = $this->ticketRepository->simpleTicketQuery($userId, $project, [], true);
 
         $ticketArray = [];
 
@@ -561,13 +564,6 @@ class Tickets extends BaseService
             $projectStatusLabels = [];
 
             foreach ($tickets as $ticket) {
-
-                // Skip tickets that belong to archived projects (state === -1).
-                // "My open tickets" shouldn't surface work from projects that
-                // are no longer active — they were padding the mobile task list.
-                if ((int) ($ticket['projectState'] ?? 0) === -1) {
-                    continue;
-                }
 
                 if ($ticket['type'] !== 'milestone') {
                     if (! isset($projectStatusLabels[$ticket['projectId']])) {
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -562,6 +562,13 @@ class Tickets extends BaseService
 
             foreach ($tickets as $ticket) {
 
+                // Skip tickets that belong to archived projects (state === -1).
+                // "My open tickets" shouldn't surface work from projects that
+                // are no longer active — they were padding the mobile task list.
+                if ((int) ($ticket['projectState'] ?? 0) === -1) {
+                    continue;
+                }
+
                 if ($ticket['type'] !== 'milestone') {
                     if (! isset($projectStatusLabels[$ticket['projectId']])) {
                         $projectStatusLabels[$ticket['projectId']] = $this->ticketRepository->getStateLabels($ticket['projectId']);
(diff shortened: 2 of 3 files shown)


marcelfolaron:
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees)

**1. Intent:** Excludes tickets belonging to archived projects (`state === -1`) from `getAllOpenUserTickets`, so "My open tickets" / the mobile task list no longer surfaces work from inactive projects — the ticket-side twin of the archived-projects picker fix (#3638).

**2. Change requests** (diff is clean; these are confirm-before-merge, not rewrites — verified against the +11/−0 diff at HEAD `f343633`):
1. `app/Domain/Tickets/Repositories/Tickets.php:741` — `zp_projects.state as projectState` is added to `simpleTicketQuery`'s select. Confirm none of the **other two callers** of `simpleTicketQuery` already select a `projectState`/`state` alias or `SELECT *`, so this can't collide or duplicate a column. The comment says it's a harmless extra column — just verify that empirically for the other two call sites.
2. `app/Domain/Tickets/Services/Tickets.php:565-571` — the guard `((int) ($ticket['projectState'] ?? 0) === -1)` filters in PHP **after** the query returns. Functionally correct, but confirm this doesn't throw off any count/pagination that assumes the pre-filter row count (e.g. a "N open tickets" badge computed from the same result set upstream). If a count is derived elsewhere from the unfiltered query, it'll now over-report.
3. Minor consistency: #3638 pushed the archived exclusion **down to the query** (`projectStatus: 'open'` → `state <> -1` in SQL) while this PR filters **in PHP**. Both are valid, but a query-level `where('zp_projects.state', '<>', -1)` in `getAllOpenUserTickets` would be consistent with #3638 and avoid transferring archived rows only to drop them. Not a blocker — flagging for symmetry; the PHP guard is safe as written.

- [ ] A user with open tickets in both active and archived projects sees only the active-project tickets in "My open tickets" / mobile list; the previously-inflated "Later" count drops accordingly.
- [ ] The other two `simpleTicketQuery` callers are unaffected (no changed row shape or duplicate column) — this is a behavior change, so a regression test (or documented manual check) covering the archived-exclusion in `getAllOpenUserTickets` is the gate.
- [ ] CI green: Pint + PHPStan level 5 + `php -l` + full unit suite. No schema/API change (read-path only, extra select column) — no migration note needed.

**4. CI status:** I can't read check-run status through my current tooling this cycle — please confirm CI is green in the UI before merge and treat any red/skipped Pint or PHPStan job as blocking. This is a **behavior change with no test in the diff (+11/−0, no test file)**, so per our test gate the regression test or a documented manual verification in checklist item 2 is a **merge blocker**, not optional.

gloriafolaron:
Thanks @Copilot — addressed in the latest commit:

- ✅ **SQL-level exclusion** — moved from a PHP post-filter to an optional `$excludeClosedProjects` flag on `simpleTicketQuery` that applies `state <> -1 OR state IS NULL` (matches the existing pattern in this repo). Closed-project rows are no longer fetched, and the extra `projectState` column was dropped from the select.
- ✅ **Terminology** — "archived" → "closed" to match how `state === -1` is described elsewhere in the Tickets domain.
- ✅ **Test** — added `test_get_all_open_user_tickets_excludes_closed_projects_at_query_level` asserting the service requests the query-level exclusion.

marcelfolaron:
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `f0e0fbb` (new commit since my 07-13 14:12 review at `f343633`)

**1. Intent:** Excludes closed-project (`state === -1`) tickets from `getAllOpenUserTickets` so "My open tickets" / the mobile task list no longer surfaces work from inactive projects — the ticket-side twin of the archived-projects picker fix (#3638).

**Prior review resolved** ✅ — the new commit `f0e0fbb` addressed all three points from my last pass: the exclusion moved from a PHP post-filter to a **query-level** `$excludeClosedProjects` flag on `simpleTicketQuery` (`state <> -1 OR state IS NULL`), the extra `projectState` select column was dropped, terminology changed "archived" → "closed", and a regression test was added. Nicely done. Re-scoped change requests against the new +38/−4 diff:

**2. Change requests:**
1. `app/Domain/Tickets/Repositories/Tickets.php:785-795` — the new guard is `state <> -1 OR state IS NULL`, correctly preserving NULL-state rows (genuine general-work / no-project tickets aren't dropped). ✔ Confirm the closure is `AND`-combined with the surrounding WHERE and not accidentally widening the result set via a stray top-level `OR` — a quick read of the generated SQL for one call confirms it.
2. `app/Domain/Tickets/Services/Tickets.php:558` — `getAllOpenUserTickets` now calls `simpleTicketQuery($userId, $project, [], true)`. The flag defaults to `false`, so the **other two callers** of `simpleTicketQuery` are unaffected. ✔ No further action — the earlier "harmless extra column" concern is moot now that the column was removed.
3. `tests/.../TicketsServiceTest.php:328` — the test asserts the service *passes* `$excludeClosedProjects = true` (captures the flag). Good contract test. Consider (non-blocking) also asserting at the **repository** level that a `state = -1` row is actually filtered from results, since the current test guards the service→repo call but not the SQL behavior itself; if the repo query regresses, this test would still pass.

1. [ ] A user with open tickets in both active and closed projects sees only active-project tickets in "My open tickets" / mobile list; the previously-inflated "Later" count drops.
2. [ ] Test present for the behavior change ✔ (`test_get_all_open_user_tickets_excludes_closed_projects_at_query_level`) — confirm it runs green in the suite, and ideally add the repo-level assertion in CR #3.
3. [ ] CI green: Pint + PHPStan level 5 + full unit suite. No schema/API change (read-path, query-only) — no migration note needed.

**4. CI status:** ⚠️ Confirm CI is green on `f0e0fbb` in the UI before merge — I can't read check-run status this cycle. The behavior change now **ships with a test** (blocker from my prior review cleared), so this is confirm-green rather than a coverage gap. Treat any red/skipped Pint or PHPStan job as blocking.

marcelfolaron:
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `0b2a3b0` (new commit since my 07-13 19:12 review at `f0e0fbb`)

**1. Intent:** Excludes closed-project (`state === -1`) tickets from `getAllOpenUserTickets` at the query level, so "My open tickets" / the mobile task list no longer surface work from inactive projects — the ticket-side twin of #3638.

**What changed since my last review:** the new commit `0b2a3b0` is a **readability-only** cleanup — the bare positional boolean `simpleTicketQuery($userId, $project, [], true)` became a **named argument** (`excludeClosedProjects: true`), dropping the empty types array and making intent self-documenting at the call site. No behavior change, no change to the SQL guard or the test. ✔ Good hygiene; all substantive items from my `f0e0fbb` review remain resolved (SQL-level `state <> -1 OR state IS NULL`, `projectState` column dropped, "archived"→"closed" terminology, regression test present).

**2. Change requests:** None new — the diff since `f0e0fbb` is cosmetic. Carrying forward the one **non-blocking** suggestion from last pass:
1. `tests/.../TicketsServiceTest.php` — the test asserts the service *passes* `excludeClosedProjects: true` (a good contract test on the service→repo call). Consider additionally asserting at the **repository** level that a `state = -1` row is actually filtered from the result set, so a future regression in the SQL guard itself would be caught (the current test would still pass if the WHERE clause broke).

1. [ ] A user with open tickets in both active and closed projects sees only active-project tickets; the previously-inflated "Later" count drops.
2. [ ] Test present for the behavior change ✔ (`test_get_all_open_user_tickets_excludes_closed_projects_at_query_level`) — confirm green in the suite; optionally add the repo-level assertion (CR #1).
3. [ ] CI green: Pint + PHPStan level 5 + full unit suite. Read-path/query-only — no schema/API change, no migration note.

**4. CI status:** ⚠️ Confirm green on `0b2a3b0` in the UI — I can't read check-run status this cycle; treat any red/skipped Pint or PHPStan job as blocking. Test gate **met** (behavior change ships with a test). This is confirm-green-and-merge from my side; the named-arg cleanup is a safe, welcome polish.

marcelfolaron:
**Maintainer review** (advisory — not an approval; merge stays with @marcelfolaron / @broskees) · head `4869235` (@ 2026-07-14)

**Intent:** Excludes closed-project (`state === -1`) tickets from `getAllOpenUserTickets()` at the SQL level via a new `excludeClosedProjects` flag on `simpleTicketQuery()`, so "My open tickets" / the mobile task list no longer surfaces work from inactive projects.

**Change requests:**
1. **Verify the `zp_projects` join is guaranteed present in `simpleTicketQuery`.** `Tickets.php` (repo) `:785` adds `where('zp_projects.state', ...)`. If any caller path reaches this query without the projects table joined, the new predicate throws an "unknown column" SQL error. Please confirm the join is unconditional in this method (the `orWhereNull('zp_projects.state')` suggests a LEFT join — good — but confirm the table is always in the FROM/JOIN list).
2. **Confirm `dateToFinish`/sort ordering is unaffected for the general-work path.** This method is the same read touched by the mobile Time/task work — cross-check that a ticket on a NULL-state project (general work, cf. #3642) still appears; the `orWhereNull` branch should cover it, but worth a manual pass so #3639 and #3642 don't interact badly.
3. *(minor)* The new bool param is 4th positional with a default — fine for BC, but the service-side call uses named-arg `excludeClosedProjects: true` (good, keeps call sites readable). No change needed.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite (the added `test_get_all_open_user_tickets_excludes_closed_projects_at_query_level` must pass).
- [ ] Manual: user with tickets in a closed project confirms they drop from "My open tickets" AND general-work (no-project) tickets still appear.
- [ ] No schema change → no migration note. Confirm no other caller of `simpleTicketQuery` regressed by the new optional param (default `false` preserves prior behavior). ✅ by inspection.

**CI status:** Not verified from here. Behavior change **is** covered by a new unit test in the diff — good. Confirm it's green before merge.

marcelfolaron:

**Status:** ready pending CI · **Priority:** P2 · **Next action:** confirm CI green, then merge · **Owner:** @marcelfolaron / @broskees

**Intent:** Adds an `excludeClosedProjects` flag to `simpleTicketQuery()` and sets it in `getAllOpenUserTickets()`, so "My open tickets" (and the mobile task list) no longer surface tickets from archived projects (`state = -1`).

### Change requests
1. **Opt-in default is correct — verify the other two callers** (`app/Domain/Tickets/Repositories/Tickets.php:711`). New param defaults to `false`, so existing callers are unchanged — good, backward-compatible. Confirm the PR body's claim that `simpleTicketQuery`'s other two callers intentionally keep archived-project tickets (i.e. excluding them there would be wrong), so the asymmetry is deliberate.
2. **`orWhereNull('zp_projects.state')` correctness** (`Tickets.php:786-789`). The `state <> -1 OR state IS NULL` pattern keeps general-work/no-project tickets visible via the null branch — good, and consistent with #3642's no-project handling. Confirm the join to `zp_projects` is a LEFT join here so no-project rows survive to hit the `IS NULL` branch (an INNER join would silently drop them regardless of this clause).
3. **Nice: test included** (`TicketsServiceTest.php`) — `test_get_all_open_user_tickets_excludes_closed_projects_at_query_level` captures the flag. Consider one repository-level test asserting the actual SQL excludes a `state=-1` row, since the service test only checks the flag is passed, not that the query honors it.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite (incl. the new test).
- [ ] Confirmed LEFT join to `zp_projects` so no-project tickets aren't dropped by the new clause.
- [ ] No schema/API change (confirmed — additive optional param, backward-compatible).

### CI status
Test present for the behavior change — not a missing-test blocker. Confirm the suite is **green** at head. The only residual gap is repo-level SQL coverage (item 3), which is a nice-to-have, not a blocker.

PR #3642, opened 2026-07-13
fix(timesheets): show a user's own general-work (no-project) time in account reads

gloriafolaron:
## Problem
A user's **general-work time** (logged with no task → the virtual `ticketId = -1`, `GENERAL_WORK_TICKET_ID`) is saved fine but **disappears from `getAllAccountTimesheets()`**. Every branch of that method's access `WHERE` keys off the entry's **project**, but a general entry has no ticket → after the `leftJoin`s, `zp_tickets.projectId` / `zp_projects.*` are NULL → it matches **none** of the branches and is dropped.

## Why it matters (mobile)
Leantime Mobile's Time tab can only call **`pollForNewTimesheets()`** over JSON-RPC — the user-scoped reads (`getWeeklyTimesheets()`, etc.) take a `CarbonInterface $fromDate`, which can't travel over JSON-RPC. So mobile has no server-side workaround: a user's no-task time is **invisible** in the app. Concretely, a 20h general entry on a day showed the weekly total as **0h**, and the user couldn't view / edit / delete it from mobile.

## Fix
One `orWhere` branch in the same access closure: a user always sees their **own** no-project entries.

```php
->orWhere(function ($q4) use ($userId) {
    $q4->whereNull('zp_tickets.projectId')
       ->where('zp_timesheets.userId', $userId);
});
```

## Scope / safety
- Non-managers are already `AND`-scoped to their own `userId` immediately below this closure, so they only ever see their own entries regardless.
- Managers already match everything via the `1=1` branch, so this is a no-op for them.
- Net effect: **only** the regular-user "see my own general time" case is rescued. No new cross-user exposure.

## Test
- `php -l` clean.
- Manual: as a non-admin, log general time (no task) → it now appears in account timesheet reads (and thus mobile's Time tab) for that user; other users' general time is unaffected.

> ⚠️ Please review before merging (do not auto-merge) — cc @marcelfolaron.

### DIFF
--- a/app/Domain/Timesheets/Repositories/Timesheets.php
+++ b/app/Domain/Timesheets/Repositories/Timesheets.php
@@ -215,6 +215,20 @@ class Timesheets extends Repository
                         if ($requesterRole === 'admin' || $requesterRole === 'manager') {
                             $q3->whereRaw('1=1');
                         }
+                    })
+                    ->orWhere(function ($q4) use ($userId) {
+                        // General-work time (virtual ticketId -1 → no matching
+                        // ticket/project after the left joins) has no project to
+                        // gate on, so it fails every project-access branch above
+                        // and silently disappears from this read — e.g. mobile's
+                        // Time tab (which can only call pollForNewTimesheets over
+                        // JSON-RPC) shows 0h even on a day with real general
+                        // hours. Always let a user see their OWN no-project
+                        // entries. Non-managers are AND-scoped to their userId
+                        // just below; managers already match via 1=1, so this
+                        // only rescues the regular-user case.
+                        $q4->whereNull('zp_tickets.projectId')
+                            ->where('zp_timesheets.userId', $userId);
                     });
             });
 


marcelfolaron:
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees) · reviewed at head `d620e1d`

**Intent:** Rescues a user's own general-work time (virtual `ticketId = -1`, no project) from being silently dropped by `getAllAccountTimesheets()`, whose access branches all key off the entry's project, by adding an `orWhere` that always lets a user see their own no-project entries.

**Change requests:**
1. **Access-boundary claim needs a second reader — this is the one thing to verify.** `Timesheets.php:219-228` adds the `orWhere($q4)` *as a sibling* of the role branches inside the same closure. The safety argument ("non-managers are AND-scoped to their own userId just below") depends on that outer `AND userId` clause actually sitting **outside** this closure and applying to the whole thing. Please confirm the structure is `WHERE ( <project-branches> OR <own-no-project> ) AND (non-manager ⇒ userId = self)` and not `WHERE <project-branches> OR ( <own-no-project> )` — an `OR` that escapes the outer userId scope would leak other users' general time to non-managers. The `$q4` closure itself is correctly self-scoped (`->where('zp_timesheets.userId', $userId)`), so even if the outer clause moved this branch stays safe; item is to confirm the *other* branches don't regress.
2. **Confirm `$userId` is the requester, not a filter argument.** The closure captures `use ($userId)`; verify `$userId` here is bound to the authenticated requester (session user) and not a caller-supplied project/user filter that an API caller could set to someone else's id.
3. **`whereNull('zp_tickets.projectId')` vs the `-1` sentinel.** The branch matches on the post-leftJoin NULL rather than on `GENERAL_WORK_TICKET_ID = -1`. That's fine for genuine general-work rows, but it will *also* match any timesheet whose ticket/project row is missing for other reasons (orphaned `ticketId` pointing at a deleted ticket). Confirm that's intended (surfacing a user's own orphaned entries is arguably correct) or tighten to the `-1` sentinel if you want to scope strictly to general work.

1. [ ] **Access test proving no cross-user leak:** as a non-manager, `getAllAccountTimesheets()` returns *my* general time but *not* another user's general time; as manager/admin the `1=1` path is unchanged. A repository/feature test asserting both directions is the guard here (see CI note — this is a behavior change to an access query with no test in the diff).
2. [ ] Manual: non-admin logs general time (no task) → appears in account reads / mobile Time tab; the reported 20h-shows-as-0h case now shows correct weekly total.
3. [ ] CI green: Pint, PHPStan level 5, `php -l`, full unit suite.

**CI status:** ⚠️ **High priority — the diff changes an access-control `WHERE` and ships no test.** Per repo policy a behavior change on a security-relevant read is a blocker until covered: add a test that asserts a non-manager cannot see another user's no-project entries (item 1 above). The PR body lists only `php -l` + a manual check; that does not exercise the cross-user boundary. Not runtime-verified in this review — flagging the access-scope structure (change request #1) as the gate before merge.

marcelfolaron:
**Maintainer review** (advisory — not an approval; merge stays with @marcelfolaron / @broskees) · head `4875605` (@ 2026-07-13)

**Intent:** Rescues a user's own general-work time (virtual `ticketId = -1`, no project) from being silently dropped by `getAllAccountTimesheets()` — it failed every project-access `WHERE` branch because it has no project after the left joins — so mobile's Time tab stops showing 0h on days with real general hours.

**Change requests:**
1. **Confirm the new `orWhere` sits at the correct nesting level.** `Timesheets.php:218` adds `->orWhere(fn($q4) => $q4->whereNull('zp_tickets.projectId')->where('zp_timesheets.userId', $userId))` as a sibling of the access closure. Because it's an `orWhere` at the closure level, please verify it's grouped **inside** the same outer `where(function…)` that wraps the access branches — otherwise the `OR (projectId IS NULL AND userId = X)` could escape the outer AND and widen results for the manager/`1=1` path. The nesting looks right from the diff, but this is the one thing that would turn a scoped fix into a cross-user leak, so verify against the full method.
2. **Guard against the admin/manager double-count.** For a manager the `1=1` branch already matches everything, so a general entry is now matched by both branches. `OR` de-dupes at the row level so this shouldn't duplicate rows — confirm the result set isn't distinct-sensitive downstream (aggregation of hours). Likely fine; flagging because it's a report read.
3. **Missing test — blocker for a data-visibility behavior change.** The PR ships `php -l` + manual steps only. A repository/integration test asserting (a) a non-admin sees their **own** general time and (b) does **not** see another user's general time is the exact guard this change needs — it touches an access `WHERE` clause. Per our test-gate rule, treat this as **required before merge**.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite.
- [ ] **A test (or documented, executed manual procedure) proving no cross-user exposure**: user A cannot see user B's general-work entries via this read. This is the security-relevant assertion.
- [ ] No schema change → no migration. Backward-compat confirmed (additive `orWhere` only).

**CI status:** Not verified from here. **Missing automated test on an access-control `WHERE` change = high-priority gap.** Please add the cross-user isolation test before merge.

marcelfolaron:

**Status:** needs-test · **Priority:** P2 · **Next action:** add a test proving no cross-user leak, confirm CI green · **Owner:** @marcelfolaron / @broskees

**Intent:** Adds an `orWhere` branch to `getAllAccountTimesheets()`'s access closure so a user always sees their **own** general-work (no-project, virtual `ticketId -1`) time, which previously vanished because every access branch keyed off a project the entry doesn't have.

### Change requests
1. **This is an access-control closure — a test is a blocker, not optional** (`app/Domain/Timesheets/Repositories/Timesheets.php:219-228`). The change adds an `orWhere` to a `WHERE` that governs *who sees which timesheets*. It must have a test proving: (a) a regular user now sees their own no-project time, AND (b) user A does **not** see user B's no-project time. `php -l` clean is not sufficient for an authorization-boundary change.
2. **Verify the `orWhere` nesting can't widen manager/other-user visibility** (`Timesheets.php`, the `->orWhere($q4)` sits as a sibling to the role branches). Because it's an `orWhere` at the closure's top level, confirm it's correctly parenthesized so it ANDs with any outer user-scope, and doesn't accidentally OR *around* the non-manager `userId` scope the PR says sits "just below." A raw SQL dump of the built query for a non-admin user in the test would settle this definitively.
3. **`whereNull('zp_tickets.projectId')` vs. `ticketId = -1`** (`Timesheets.php:227`). The branch keys on the post-left-join null projectId rather than the sentinel `ticketId = -1`. Confirm no *other* path produces a null `zp_tickets.projectId` for a row that ISN'T the current user's general-work (e.g. an orphaned timesheet whose ticket was deleted) — if it can, pairing the null check with `userId` (already present) is what saves it, so this is likely fine, but worth a sentence in the test.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite.
- [ ] A test proving the access boundary: own no-project time visible **and** another user's no-project time NOT visible to a non-manager.
- [ ] No schema/API change (confirmed — query-scope only, backward-compatible).

### CI status
⚠️ **Missing test on an authorization-boundary change = blocker.** The PR lists only `php -l` + a manual check. For a `WHERE` clause that controls timesheet visibility across users, an automated cross-user isolation test is required before merge.

PR #3644, opened 2026-07-13
fix(content-templates): polish batch — Phase 4 slug bug, status label rendering, typos

gloriafolaron:
## Summary

Cleanup for six items flagged during the pre-merge review of [#3493](https://github.com/Leantime/leantime/pull/3493) that shipped as-is. All latent (no user reports), but two are functionally significant:

- **Phase 4 startContent has been a silent no-op since it merged** — the feature the PR ostensibly added never actually runs on any board creation.
- **The Green / Yellow / Red / Gray Status wiki templates render as literal HTML in the editor** — the fix was written during the June session but never committed. Same bug that motivated the original session.

### 1. `Blueprints::applyStartContent` slug bug

`createBoard()` is called from the Blueprints controllers with the **database type** (`"swotcanvas"`, `"logicmodelcanvas"`). But both the Blueprints `TemplateRegistry` and the core `ContentTemplateRegistry` key by the **slug** (`"swot"`, `"logicmodel"`). The line

```php
$blueprint = $this->templateRegistry->get($canvasType);
```

silently returned `null` for every real board type. `TemplateRegistry` has a separate `getByDatabaseType()` that strips the `canvas` suffix — this PR wires that in and passes the resolved slug to the ContentTemplates lookups so both sides agree.

Regression test added under `BlueprintsServiceTest::test_create_board_applies_start_content_against_the_slug_not_the_db_type`. Uses an anonymous subclass of `ContentTemplateRegistry` to record what slug reaches `get()`.

### 2. `CanvasItemsApplier` missing `modified` timestamp

The insert only set `created`. Board "last updated" ordering across the app uses `COALESCE(MAX(zp_canvas_items.modified), zp_canvas.created)` — and `MAX()` ignores `NULL`s, so freshly-templated boards silently sort as stale in recent-activity lists. Set `modified = created` on the insert.

### 3-6. Status label wiki templates → TipTap `<mark>` highlights

Green / Yellow / Red / Gray shipped with `<span class="label label-warning">Yellow</span>`-style payloads. TipTap's configured extension set has **no Span mark** for arbitrary classes (StarterKit + Highlight + Link + Underline + Typography + Superscript), so at insert time the raw HTML falls through as escaped literal text in the editor. Rewritten to `<p><mark data-color="…" style="background-color: …">…</mark></p>` which the already-loaded Highlight extension renders as a colored badge.

*This was the specific bug that motivated the June review session — the fix was written at the time but never `git add`'d before the PR merged. Restoring here.*

### 7. `wiki/prd.yaml` missing `)` on "Goals" heading

`<h2>Goals (What are we working towards?</h2>` → `<h2>Goals (What are we working towards?)</h2>`

### 8. `wiki/bug-report.yaml` "your are" → "you are"

User-visible template text in the Summary row.

### 9. `Library/leancanvas/` → `Library/lean/`

The directory-per-appliesTo convention keys by the Blueprints slug (`lean`, `swot`, `logicmodel`, `goal`). `leancanvas` was the only outlier — plugin authors following the naming convention would have dropped templates into a bucket the registry never looks at (`leancanvas` vs the correct `lean`).

## Test plan

- [x] `vendor/bin/codecept run Unit` — 671 tests, 1678 assertions, all passing (+1 new regression test)
- [x] `vendor/bin/pint --test` on touched files — clean
- [x] `vendor/bin/phpstan analyse` on touched files — no errors
- [ ] Manual: create a SWOT board from the Blueprints hub — with a `startContent` set on the SWOT YAML (not currently the case in shipped core, but plugins can wire one). Confirm items appear.
- [ ] Manual: in wiki editor, `/` → Yellow Status template → confirm colored highlight rendering instead of raw HTML text.
- [ ] Manual: same for Green / Red / Gray Status templates.

### DIFF
--- a/app/Domain/Blueprints/Services/Blueprints.php
+++ b/app/Domain/Blueprints/Services/Blueprints.php
@@ -244,23 +244,31 @@ class Blueprints extends BaseService
      */
     private function applyStartContent(int $canvasId, string $canvasType): void
     {
-        $blueprint = $this->templateRegistry->get($canvasType);
+        // createBoard() is invoked with the DATABASE type (e.g. "swotcanvas",
+        // "logicmodelcanvas"), but both registries key by the SLUG form
+        // ("swot", "logicmodel"). Use getByDatabaseType() to bridge, then
+        // pass the resolved slug to the ContentTemplates lookups so both
+        // sides agree on the identifier. Prior to this the initial registry
+        // read silently returned null and the whole feature was a no-op.
+        $blueprint = $this->templateRegistry->getByDatabaseType($canvasType);
         if ($blueprint === null || $blueprint->startContent === null) {
             return;
         }
 
-        $contentTpl = $this->contentTemplates->get($canvasType, $blueprint->startContent);
+        $slug = $blueprint->slug;
+
+        $contentTpl = $this->contentTemplates->get($slug, $blueprint->startContent);
         if ($contentTpl === null) {
             Log::debug(sprintf(
                 'Blueprints::createBoard: blueprint "%s" references startContent "%s" but the template was not found.',
-                $canvasType,
+                $slug,
                 $blueprint->startContent
             ));
 
             return;
         }
 
-        $applier = $this->contentTemplates->applierFor($canvasType);
+        $applier = $this->contentTemplates->applierFor($slug);
         if ($applier === null) {
             return;
         }
--- a/app/Domain/Blueprints/Services/TemplateRegistry.php
+++ b/app/Domain/Blueprints/Services/TemplateRegistry.php
@@ -65,7 +65,12 @@ class TemplateRegistry
      */
     public function getByDatabaseType(string $dbType): ?CanvasTemplate
     {
-        $slug = str_replace('canvas', '', $dbType);
+        // Strip the trailing "canvas" suffix only — a naive str_replace would
+        // corrupt any type whose name embeds the word (e.g., "canvassing") or
+        // return an empty slug for the bare "canvas" type.
+        $slug = str_ends_with($dbType, 'canvas')
+            ? substr($dbType, 0, -6)
+            : $dbType;
 
         return $this->get($slug);
     }
--- a/app/Domain/Blueprints/Services/TemplateRegistry.php
+++ b/app/Domain/Blueprints/Services/TemplateRegistry.php
@@ -65,11 +65,13 @@ class TemplateRegistry
      */
     public function getByDatabaseType(string $dbType): ?CanvasTemplate
     {
-        // Strip the trailing "canvas" suffix only — a naive str_replace would
-        // corrupt any type whose name embeds the word (e.g., "canvassing") or
-        // return an empty slug for the bare "canvas" type.
-        $slug = str_ends_with($dbType, 'canvas')
-            ? substr($dbType, 0, -6)
+        // Strip a trailing "canvas" suffix only — a naive str_replace would
+        // corrupt any type whose name embeds the word (e.g., "canvassing").
+        // Require something *before* the suffix so the bare "canvas" type
+        // resolves to itself, not an empty slug; -strlen() avoids a brittle -6.
+        $suffix = 'canvas';
+        $slug = str_ends_with($dbType, $suffix) && strlen($dbType) > strlen($suffix)
+            ? substr($dbType, 0, -strlen($suffix))
             : $dbType;
 
         return $this->get($slug);
--- a/app/Domain/ContentTemplates/Library/wiki/bug-report.yaml
+++ b/app/Domain/ContentTemplates/Library/wiki/bug-report.yaml
@@ -16,7 +16,7 @@ wiki:
          <tbody>
          <tr>
          <td style="width: 17.2438%; background-color: var(--accent1);"><span style="color: #ffffff;"><strong>Summary</strong></span></td>
-         <td style="width: 82.7562%;">summarize the issue your are experiencing</td>
+         <td style="width: 82.7562%;">summarize the issue you are experiencing</td>
          </tr>
          <tr>
          <td style="width: 17.2438%; background-color: var(--accent1);"><span style="color: #ffffff;"><strong>Environment</strong></span></td>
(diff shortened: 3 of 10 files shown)


marcelfolaron:
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees) · reviewed at head `91be44f`

**1. Intent:** Polish batch cleaning up six items flagged during the #3493 pre-merge review that shipped as-is — the Phase-4 `applyStartContent` slug bug (a silent no-op since merge), the Green/Yellow/Red/Gray status wiki templates rendering as literal HTML in the editor, a shared per-apply timestamp on templated canvas items, and three small YAML/dir-naming fixes.

**2. Change requests** (diff read in full at `91be44f`, +96/−12 across 11 files; these are confirm-before-merge, the changes themselves are correct):
1. `app/Domain/Blueprints/Services/TemplateRegistry.php:68-73` — the suffix strip was correctly hardened from `str_replace('canvas', '')` to `str_ends_with(...) ? substr($dbType, 0, -6) : $dbType`. ✔ Good catch on the Copilot follow-up — the old form would have corrupted an embedded "canvas" and emptied the bare `canvas` type. Confirm no db-type in the wild is *exactly* `canvas` (which would now resolve to an empty slug and `get('')` → null) — from the registry that's a safe null, but worth a mental check that no board type is literally named `canvas`.
2. `app/Domain/Blueprints/Services/Blueprints.php:247-271` — `applyStartContent` now resolves via `getByDatabaseType($canvasType)` then threads `$blueprint->slug` into both `contentTemplates->get($slug, ...)` and `applierFor($slug)`. ✔ Correct — both sides now agree on the slug. One thing to verify: the regression test (`BlueprintsServiceTest:510`) stubs `getByDatabaseType()` with the **old** `str_replace('canvas','')` logic inside the anonymous class, not the new `substr` form. The test still passes because `'swotcanvas'`→`'swot'` is identical under both, but the test no longer exercises the hardened path it's guarding. Consider updating the stub to mirror the shipped `str_ends_with/substr` so the test and prod code can't drift.
3. `app/Domain/ContentTemplates/Services/Appliers/CanvasItemsApplier.php:70-92` — hoisting `$now = now()` out of the loop and setting `modified = $now` (not null) is the right fix for the `COALESCE(MAX(modified), created)` staleness bug. ✔ No change needed.
4. Status templates (`status-{green,yellow,red,gray}.yaml`) — the `<span class="label label-*">` → `<p><mark data-color=… style="background-color:…">` rewrite matches TipTap's registered Highlight extension. ✔ Confirm the hardcoded hex colors (`#75BB1B`, `#F1C40F` w/ dark text, `#E74C3C`, `#95A5A6`) are the intended palette and render acceptably in both light and dark editor themes — the yellow correctly uses `#1a1a1a` text while the others use `#ffffff`, so contrast was considered.

1. [ ] Create a SWOT/logic-model board from the Blueprints hub with a `startContent` set → items actually apply (the Phase-4 no-op is fixed). The shipped core doesn't wire a `startContent`, so this needs a manual test with one set (or a plugin that does) — the unit test locks the *slug-resolution* contract but not the end-to-end apply.
2. [ ] In the wiki editor, insert each of Green/Yellow/Red/Gray Status templates → colored highlight badge renders, not raw HTML text (the original user-reported bug).
3. [ ] CI green: Pint + PHPStan level 5 + full unit suite (PR body reports 671 tests +1 new regression, Pint/PHPStan clean on touched files — confirm the full suite, not just touched files, is green in CI).

**4. CI status:** ⚠️ Confirm green on `91be44f` in the UI — I can't read check-run status this cycle; treat any red/skipped Pint or PHPStan job as blocking. Test-gate note: the Phase-4 slug fix is a **behavior change** (a feature that never ran now runs) and it **ships with a regression test** (`test_create_board_applies_start_content_against_the_slug_not_the_db_type`) ✔, so the gate is met — with the caveat in CR #2 that the test's stub should track the hardened suffix logic. The status-template and timestamp fixes are low-risk and covered by manual verification (items 1–2). No schema/API change — no migration note needed.

marcelfolaron:
**Maintainer review** (advisory — not an approval; merge stays with @marcelfolaron / @broskees) · head `4876818` (@ 2026-07-13)

**Intent:** Cleanup batch for six items flagged during the #3493 pre-merge review — two functionally significant (the Phase-4 `startContent` slug bug that made the feature a silent no-op; the Green/Yellow/Red/Gray status wiki templates rendering as literal HTML) plus typos and a directory rename.

**Change requests:**
1. **`getByDatabaseType()` suffix strip — good fix, verify the `str_ends_with` edge.** `TemplateRegistry.php:68` replaces the naive `str_replace('canvas', '', $dbType)` with a trailing-only strip. ✅ This correctly fixes `leancanvas`→`lean` and won't corrupt an embedded "canvas". Confirm no caller passes an already-slugged value (`"swot"`) expecting it to pass through — it does (the `else` branch returns `$dbType` unchanged), so that's safe.
2. **Directory rename `Library/leancanvas/` → `Library/lean/` — confirm nothing references the old path.** The rename is correct per the slug convention, but grep for `leancanvas` across the plugins repo too (PgmPro/StrategyPro template authors may have hardcoded the old bucket). A stale reference would silently drop templates into an unread directory again.
3. **Status templates → `<mark>` — verify against the actual TipTap extension config.** `status-{green,yellow,red,gray}.yaml` now emit `<p><mark data-color=… style=…>`. The comment claims the Highlight extension is `multicolor`; please confirm the editor's Highlight is configured with `multicolor: true` (otherwise `data-color`/inline bg is stripped on load). One manual insert of each of the four templates is the real proof.
4. **`CanvasItemsApplier` `modified = $now` — correct.** `:86` setting `modified` alongside `created` fixes the `COALESCE(MAX(modified), created)` staleness. ✅ No concerns.
5. *(nit)* Typos (`your are`→`you are`, missing `)` on the PRD Goals heading) are correct and user-visible. ✅

- [ ] CI green: Pint + PHPStan level 5 + full unit suite (incl. the new `test_create_board_applies_start_content_against_the_slug_not_the_db_type` regression test).
- [ ] Manual: create a SWOT board with a `startContent`-bearing blueprint → items appear (proves the no-op is fixed); insert each of the 4 status templates in the wiki editor → colored badge, not raw HTML.
- [ ] No schema change / no migration. Confirm the `leancanvas`→`lean` rename has no dangling references in **both** repos.

**CI status:** Not verified from here. The functionally-significant Phase-4 fix **is** covered by a new regression test — good. The status-template rendering is the one behavior with no automated coverage (view-layer); the manual insert check is the gate there.

Issue #72, opened 2019-12-09
Milestone page shows empty after adding a new milestone


GhaziTriki:
I see an empty page after adding a new milestone. I am using nginx.

![image](https://user-images.githubusercontent.com/4991088/70471327-36e7cc00-1acd-11ea-941b-2e3bfd0c38ea.png)

marcelfolaron:
Do you have steps to reproduce the issue? Where did you create the milestone (on the milestone page or from the To-Do page)? Did you see any javascript issues? I don't think this is a nginx problem since it routed correctly. It seems the javascript didn't fire.

GhaziTriki:
json extension was disabled.

PR #3645, opened 2026-07-14
feat(core): Resources contract surface — gateway, registry, value objects, shared blade primitives

gloriafolaron:
## Summary

Introduces `app/Core/Resources` — a small contract layer that lets a plugin provide Resources data (people allocations, budget lines, dependencies at the program level) to the rest of the app without core knowing which plugin is providing.

Follows the WorkStructure precedent verbatim: **core owns the contract + registry + typed value objects; the plugin owns the concrete implementation, storage, and UI.** Companion plugin PR: [Leantime/plugins#72](https://github.com/Leantime/plugins/pull/72).

## Architectural rationale

**Why not put the whole thing in the plugin.** Resources are a concept the stakeholder report (core) needs to render a Resources section. Baking a hardcoded reference to PgmPro into core would be wrong. The gateway/registry pattern lets core define the concept and any plugin implement it, matching how `Core\WorkStructure` does the same for structures and how `Domain\ContentTemplates` does the same for template libraries.

**Why not events/filters instead of a formal contract.** Contract + typed value objects give consumers a real IDE-navigable, PHPStan-checkable surface. Filters would work but push the shape into runtime string keys that break silently when the plugin changes.

**Why "first plugin wins" instead of a chain.** There is exactly one Resources provider per install in v1 (PgmPro). If a second plugin ever wants to provide, that's a real product decision (which numbers do reports believe?) that shouldn't be silently resolved by registration order — the second plugin's registration is logged and refused.

### `app/Core/Resources/`

| File | Purpose |
|---|---|
| `Contracts/ResourcesGateway` | Interface. Two methods: `getForProjects(int[] $ids): ResourceSummary`, `getForProgram(int $id): ResourceSummary`. |
| `Models/ResourceSummary` | The aggregation shape every consumer sees. `::empty()` distinguishes "plugin installed but nothing authored" from "no plugin registered" (which is null from the registry). |
| `Models/PersonAllocation` | Person + weekly capacity + per-project allocations map. `totalAllocated()`, `available()`. |
| `Models/BudgetLine` | Budget assigned to one project. `spent` is a committed non-null `float` — providers that don't track actuals pass `0.0`. We deliberately do not encode an "unknown actuals" state at model level; that ambiguity would leak into every consumer (report tile, UI, RPC caller). Providers that need to surface "actuals unavailable" do so via their own domain, not through a nullable here. |
| `Models/Dependency` | External partnership. `confirmed` bool distinguishes "we have this locked in" from "we're counting on it." |
| `Services/ResourcesRegistry` | Singleton. Same-class re-registration replaces the stored instance (last-write-wins; safe because two instances of the same class are functionally interchangeable). Cross-class registration is refused and logged. |
| `ResourcesServiceProvider` | Binds the registry as a singleton. Nothing seeded — an install without a Resources plugin gets a registry that resolves to null. |

Wired into `laravelConfig.php` next to `WorkStructureServiceProvider`.

## Reporting integration

**Not in this PR.** The seam is a one-liner in the future report engine (see `feature/report-screens`):

```php
$resources = $this->resourcesRegistry->resolve()?->getForProjects($projectIds);
return [..., 'resources' => $resources, ...];
```

Lands as a follow-up after the report screens branch merges. Keeps this PR reviewable independently of the reporting work.

## Test plan

- [x] `vendor/bin/pint --test` — clean
- [x] `vendor/bin/phpstan analyse` — no errors on touched files
- [x] `vendor/bin/codecept run Unit` — full suite passing
- [x] 12 new tests under `tests/Unit/app/Core/Resources/`:
  - `ResourcesRegistryTest` (4 tests) — null-when-empty, resolve-returns-registered, same-class-replaces-instance, different-class-refused-with-warning-logged
  - `ResourceSummaryTest` (8 tests) — utilization edge cases (zero-divide, over-allocation), isEmpty in the "only people" and "only dependencies" cases
- [ ] Manual: verify `ResourcesRegistry` resolves to null on an install without the companion plugin — no crashes anywhere in the app.
- [ ] Manual (with companion plugin installed): verify `app()->make(ResourcesRegistry::class)->resolve()` returns the PgmPro implementation.

## What this PR does NOT change

- No new DB tables. Existing `zp_canvas` / `zp_canvas_items` handle storage (plugin side).
- No new routes.
- No migration.
- Zero behavior change for installs without the companion plugin — the registry resolves to null and every planned consumer treats that as "no Resources provider."

Pairs with plugin PR [Leantime/plugins#72](https://github.com/Leantime/plugins/pull/72).


marcelfolaron:
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees) · reviewed at head `80e1e05`

**1. Intent:** Adds `app/Core/Resources` — a contract layer (gateway interface + first-wins registry + immutable value objects `ResourceSummary`/`PersonAllocation`/`BudgetLine`/`Dependency`) that lets a plugin provide program-level Resources data (people/budget/dependencies) to core consumers without core knowing the provider, following the `Core\WorkStructure` precedent. Companion plugin PR: #72.

**2. Change requests** (full diff read at `80e1e05`, +399 across 10 files; the design is clean and well-documented — these are confirm-before-merge):
1. `app/Core/Resources/Services/ResourcesRegistry.php:39-52` — the register() semantics are now **same-class-replaces, cross-class-refused-and-logged**, and the docblock (fixed per the Copilot pass) matches the code and the test. ✔ One thing to confirm: `registeredBy` stores `$gateway::class`, so two *different anonymous* classes count as distinct and the second is refused — correct. But note that if PgmPro's provider class is ever reloaded under a different fully-qualified name (e.g. a namespaced alias), it'd be treated as cross-class and refused. Not a real risk for a single-provider install; flagging only so the "first wins" behavior isn't surprising during plugin dev/hot-reload.
2. `app/Core/Configuration/laravelConfig.php:45` — `ResourcesServiceProvider` is registered right after `WorkStructureServiceProvider` and **before** `ContentTemplatesServiceProvider`. ✔ Since the plugin registers *into* the registry from its own boot (post-plugins-load), confirm the provider-ordering guarantees the `ResourcesRegistry` singleton is bound before any plugin `register.php` runs — it should be (core providers boot before plugins), but it's the one ordering assumption the whole seam rests on.
3. **Submodule pointer moved in this PR.** `app/Plugins` bumps `289d476…` → `4b55637…` (the companion #72 head). That couples this "core-only, reviewable independently" PR to the plugin submodule state — if #3645 merges before #72, master's submodule pointer references an unmerged plugin commit. Recommend **dropping the submodule bump from this PR** and doing the pointer bump as the explicit last step after both #3645 and #72 merge (matches the merge-order discipline already used for #3643/plugins#71). This is the one thing I'd change before merge.
4. `app/Core/Resources/Models/BudgetLine.php:22` — `spent` was correctly made **non-nullable** (0.0 when no actuals) per the Copilot pass, removing the unknown-vs-zero ambiguity from `totalSpent`/`budgetUtilization()`. ✔ Consistent with the value-object philosophy; no change.
5. Note the earlier speculative Blade primitives (avatar-stack, collapsible, metric-cell, proportion-bar) were **deleted** in `cfe5dc6` (no consumers; collapsible needed Alpine which isn't loaded) — good call removing dead surface rather than shipping it "for later." The PR description still lists them under "4 shared Blade primitives"; update the description so it matches the shipped diff.

1. [ ] Fresh install with **no** Resources plugin: `app()->make(ResourcesRegistry::class)->resolve()` returns null and nothing in the app crashes (the honest "not installed" state). Zero-behavior-change claim verified.
2. [ ] Tests green and present for the behavior ✔ — `ResourcesRegistryTest` (null-when-empty, resolve, same-class idempotent, cross-class refused + `Log::warning` asserted) and `ResourceSummaryTest` (utilization zero-divide, over-allocation, isEmpty incl. deps-only) cover the core contract. Confirm they run green in the full suite (PR body: 681 tests / 1700 assertions).
3. [ ] Merge order + submodule: **#3645 (core, without the submodule bump) → #72 (plugin) → single submodule-pointer bump** as the last step (CR #3). No DB tables, no routes, no migration — confirmed from the diff.

**4. CI status:** ⚠️ Confirm Pint + PHPStan level 5 + full unit suite green on `80e1e05` in the UI — I can't read check-run status this cycle; treat any red/skipped job as blocking. Test gate is **met** (the new contract ships with 11 tests). The primary merge-blocking item is not a test gap but the submodule-pointer coupling in CR #3 — a core "reviewable independently" PR shouldn't carry a bump to an unmerged plugin commit. No schema/API change ⇒ no migration note.

marcelfolaron:
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `bf4f33f` (new commit since my earlier review at `80e1e05`)

**1. Intent:** Adds `app/Core/Resources` — the contract layer (gateway interface + first-wins registry + immutable value objects `ResourceSummary`/`PersonAllocation`/`BudgetLine`/`Dependency`) that lets a plugin provide program-level Resources data to core consumers without core knowing the provider, following the `Core\WorkStructure` precedent. Companion plugin PR: #72.

**What changed since `80e1e05`:**
- ✅ **CR#5 resolved** — the speculative Blade primitives (avatar-stack, collapsible, metric-cell, proportion-bar) are gone from the diff; the PR is now purely the contract surface + tests. Good.
- ⚠️ **CR#3 is NOT resolved — and was freshly re-touched.** `app/Plugins` still moves in this PR: the submodule pointer is now bumped `289d476…` → `d45c03454…`, and `d45c03454b5fa745c17b786e2a625a2a644a4eab` is exactly the **current head of the unmerged plugin PR #72**. So this "core, reviewable independently" PR carries a submodule bump to an unmerged plugin commit — if #3645 merges before #72, master's submodule pointer references a commit that isn't on plugins `main`.

**2. Change requests:**
1. **Drop the `app/Plugins` submodule bump from this PR** (carried-forward CR#3, still the one merge-blocker). Do the pointer bump as the explicit **last** step after both #3645 and #72 merge — same merge-order discipline already used for #3643 / plugins#71. Right now the bump targets an unmerged commit.
2. `app/Core/Configuration/laravelConfig.php:45` — `ResourcesServiceProvider` is registered right after `WorkStructureServiceProvider` and before `ContentTemplatesServiceProvider`. ✔ The whole seam rests on the `ResourcesRegistry` singleton being bound before any plugin `register.php` runs (the plugin registers *into* it on the post-plugins-load event, priority 25). Core providers boot before plugins, so this holds — just confirming it's the one ordering assumption.
3. `app/Core/Resources/Services/ResourcesRegistry.php:39-52` — first-wins with same-class-replaces / cross-class-refused-and-logged is clean and matches the test. ✔ No change. (Note for plugin-dev/hot-reload only: a provider reloaded under a different FQN counts as cross-class and is refused — not a real risk for a single-provider install.)

1. [ ] Fresh install with **no** Resources plugin: `app()->make(ResourcesRegistry::class)->resolve()` returns null and nothing crashes (the honest "not installed" state) — zero-behavior-change claim verified.
2. [ ] Tests green ✔ present — `ResourcesRegistryTest` (null-when-empty, resolve, same-class idempotent, cross-class refused + `Log::warning` asserted) and `ResourceSummaryTest` (utilization zero-divide, over-allocation, isEmpty incl. deps-only). Confirm green in the full suite.
3. [ ] **Merge order + submodule:** #3645 (core, *without* the submodule bump) → #72 (plugin) → single submodule-pointer bump as the last step (CR#1). No DB tables, no routes, no migration — confirmed from the diff.

**4. CI status:** ⚠️ Confirm Pint + PHPStan level 5 + full unit suite green on `bf4f33f` in the UI — I can't read check-run status this cycle; treat any red/skipped job as blocking. Test gate is **met** (contract ships with 11 tests). The primary merge-blocking item is not a test gap but the submodule-pointer coupling in CR#1.

marcelfolaron:
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `18c18e39` (advanced past `bf4f33f` from my last pass) · full diff read

**1. Intent:** Adds `app/Core/Resources` — the contract layer (gateway interface + first-wins registry + immutable value objects `ResourceSummary`/`PersonAllocation`/`BudgetLine`/`Dependency`) letting a plugin provide program-level Resources data to core consumers, following the `Core\WorkStructure` precedent. Companion plugin PR #72.

**What I verified at `18c18e39`:**
- ✅ Contract + registry + 4 models + `ResourcesServiceProvider` are clean; **11 tests present** (`ResourcesRegistryTest` ×4: null-when-empty, resolve, same-class-replaces, cross-class-refused-with-`Log::warning`; `ResourceSummaryTest` ×7: utilization zero-divide, over-allocation, isEmpty incl. deps-only). Test gate **met**.
- ✅ `laravelConfig.php:45` registers `ResourcesServiceProvider` right after `WorkStructureServiceProvider`, before `ContentTemplatesServiceProvider` — core boots before plugins, so the registry singleton is bound before any plugin `register.php` runs. Correct.

**2. Change requests:**
1. **⛔ Carried-forward, STILL unresolved — drop the `app/Plugins` submodule bump.** The diff still moves the pointer `289d476…` → `d45c03454…`, and `d45c03454` is the current head of the **unmerged** plugin PR #72. This is the same coupling I flagged at `80e1e05` and `bf4f33f`. If #3645 merges before #72, master's submodule references a commit that isn't on plugins `main`. Remove the bump from this PR and do the pointer bump as the explicit **last** step after both #3645 and #72 merge. **This remains the single merge-blocker.**
2. `Models/BudgetLine.php` — `spent` is non-nullable (`0.0` default) with the unknown-vs-zero rationale documented. ✔ No change.
3. `Services/ResourcesRegistry.php` — first-wins / same-class-replaces / cross-class-refused-and-logged matches the tests exactly. ✔ No change. (PHPStan-5 note: `hydrateBudget` on the plugin side uses `?? 0.0` on the now-non-nullable `spent` — harmless, may draw an always-false null-coalesce flag there, not here.)

1. [ ] Fresh install with **no** Resources plugin: `app()->make(ResourcesRegistry::class)->resolve()` returns null, nothing crashes (zero-behavior-change verified).
2. [ ] Tests green ✔ present (11) — confirm in the full suite.
3. [ ] **Merge order + submodule:** #3645 (core, *without* the bump) → #72 (plugin) → single submodule-pointer bump as the last step (CR #1).

**4. CI status:** ⚠️ Can't read check-run status this cycle — confirm Pint + PHPStan level 5 + full unit suite green on `18c18e39` in the UI; treat any red/skipped job as blocking. Test gate is **met**; the merge-blocker is the submodule-pointer coupling (CR #1), not coverage. No DB tables / routes / migration — confirmed from the diff.

PR #3653, opened 2026-07-16
feat(users): weekly_hours + employment_type on user (Slice 1)

gloriafolaron:
Adds two nullable capacity attributes to zp_user. Ships the model in isolation — nothing reads these values yet.

**Migration** update_sql_30523 adds weekly_hours + employment_type (nullable, no defaults, no backfill).

**Enum** EmploymentType (FTE / PT / Contractor / Volunteer) with countsAgainstCapacity() + overCapTone() helpers.

**Repository** editUser + addUser accept + persist the fields using array_key_exists (partial updates preserve existing values).

**UI** editUser.blade.php Capacity section gated behind Roles::admin.

**Design decisions**
- No backfill.
- Admin-only edit.
- Nullable — downstream code treats NULL as unconfigured and skips warnings.

**Test plan**
- [x] Migration ran: db-version 3.5.22 → 3.5.23.
- [x] Round-trip save works.
- [x] Fresh-install schema includes both columns.

🤖 Generated with Claude Code


marcelfolaron:
**Maintainer triage** (advisory — merge authority stays with @marcelfolaron / @broskees) · first pass · reviewed at head `28ace124` · full diff read · status: **needs-changes** · priority: **P2** · next action: add persistence tests + confirm base branch · owner: @gloriafolaron

**1. Intent:** Slice 1 of user-level capacity modelling — adds two nullable columns to `zp_user` (`weekly_hours`, `employment_type`), an `EmploymentType` enum with capacity-semantics helpers, repo persistence, and an admin-only Capacity section on the edit-user form. Model ships in isolation; nothing reads it yet.

**2. Change requests:**
1. **⛔ CR #1 — behavior change with zero tests (merge blocker).** This PR adds real persistence logic — `Users::editUser`/`addUser` now parse and coerce two fields (`Repositories/Users.php:301-317`, `:411-412`) and a migration mutates schema (`Repositories/Install.php:2936` `update_sql_30523`). None of it is covered. Per the standing test-gate, a persistence behavior change needs a test or a documented manual procedure for the core behavior. Add unit coverage for `editUser`: (a) `weekly_hours: ''` / `null` persists `NULL`, not `0`; (b) `weekly_hours: '40'` persists int `40`; (c) omitting the keys entirely leaves an existing value untouched (the `array_key_exists` partial-update contract — the whole reason for the guard); (d) `employment_type` accepts only enum values / an unknown string round-trips or is rejected per intent.
2. **⚠ CR #2 — base branch is `feat/logic-model-reverse`, not `master`.** This PR targets `feat/logic-model-reverse` (`base.ref`), so its diff and eventual merge ride that feature branch rather than landing a standalone schema slice on `master`. A `zp_user` migration + enum is core infrastructure several downstream PRs will want — confirm the target is intentional; if it should land independently on `master`, retarget before review so the migration sequences cleanly against db-version `3.5.23`.
3. **⚠ CR #3 — no server-side validation of `employment_type`.** `editUser`/`addUser` cast the posted value to `(string)` and store it verbatim (`Repositories/Users.php:314`, `:412`); the enum is used only to render the `<select>`. A crafted POST can store an arbitrary 20-char string that later `EmploymentType::from()` will throw on. Validate against `EmploymentType::tryFrom()` at the persistence (or controller) boundary and drop/normalize invalid input to `NULL`.
4. **✓ Nits (non-blocking):** `EmploymentType` is a value enum but lives under `Domain\Users\Contracts\` — fine, but `Contracts` usually implies interfaces; consider `Domain\Users\Enums\` for discoverability. Migration correctly guards with `Schema::hasColumn` (idempotent ✔) and fresh-install schema is updated in `sqlPrep()` ✔.

1. [ ] Persistence unit tests for `editUser`/`addUser` covering NULL-vs-int coercion + the `array_key_exists` partial-update guard (CR #1) — blocker.
2. [ ] Base branch confirmed intentional (`feat/logic-model-reverse` vs `master`, CR #2); `employment_type` validated against the enum at the write boundary (CR #3).
3. [ ] CI green: Pint + PHPStan level 5 + full unit suite. **Schema change** — migration `update_sql_30523` runs clean on upgrade *and* fresh install (db-version 3.5.22 → 3.5.23); both new columns nullable, no backfill, so no data-migration/rollback risk beyond a column drop.

**4. CI status:** ⚠️ Can't read check-runs this cycle — confirm green on `28ace124` in the UI; treat any red/skipped Pint or PHPStan job as blocking. The **missing persistence tests (checklist #1) are the merge blocker** — this is the layer downstream capacity math will trust, so lock the NULL/int/partial-update contract with tests before it ships.

marcelfolaron:
**Maintainer review** (advisory — not an approval/merge) · head `28ace124` · reviewed from full diff

**Intent:** Ships Slice 1 of user capacity attributes — adds nullable `weekly_hours` + `employment_type` columns to `zp_user` (migration 30523), an `EmploymentType` enum, and admin-only edit UI/persistence, with nothing reading the values yet.

**Change requests:**

1. **BLOCKER — migration run-array skips 30522.** In `app/Domain/Install/Repositories/Install.php` the `$dbUpdates`/run array is patched to `…30519, 30520, 30521, 30523,` — `30522` is omitted even though `public function update_sql_30522()` exists in the same file. If 30522 is not registered in the run array, it never executes on upgrade, and jumping the recorded db-version straight to 30523 will permanently skip it. Confirm whether 30522 is registered elsewhere; if not, add it to the array (ascending) before this merges.
   ```php
   30519,
   30520,
   30521,
   30523,   // ← 30522 missing here
   ```

2. **BLOCKER — no automated test for the behavior change.** The diff adds real persistence logic (`Users::editUser` / `addUser` `array_key_exists` guards + int/string casts) and enum decision logic (`countsAgainstCapacity()`, `overCapTone()`) but contains no `*Test.php`. Per our test-gate, a behavior change needs a unit test (or a documented manual procedure in the PR). At minimum: a repo round-trip test (set → null-preserve on omitted key → clear on empty string) and an enum test asserting Volunteer is excluded from capacity and the tone mapping.

3. **`employment_type` accepts any string — no enum validation on write.** `Users::editUser`/`addUser` cast the posted value with `(string)` and persist it directly; nothing constrains it to the four `EmploymentType` cases. A malformed POST persists garbage that downstream `EmploymentType::from()` will later throw on. Suggest validating against `EmploymentType::tryFrom()` (reject/normalize) before persist.

4. **Confirm `weekly_hours` bounds.** Column is `int(11)` nullable with no range guard; a negative or absurd value persists. If downstream capacity math divides by it, add a `>= 0` (and sane upper-bound) guard at the repo boundary.

- [ ] CI green: Pint + PHPStan level 5 + full unit suite on `28ace124`.
- [ ] Migration verified: fresh install includes both columns AND upgrade path runs 30522 then 30523 (db-version 3.5.22 → 3.5.23) — resolve item #1 first.
- [ ] A unit test (or documented manual procedure) covering repo round-trip + enum capacity/tone logic (item #2).

**CI status:** I can't read check-runs from here — please confirm Pint/PHPStan/unit are green in the PR checks tab. Migration ordering (#1) and missing tests (#2) are treated as blockers regardless of CI. Note this bumps `dbVersion` into the 3.5.2x migration band Marcel is sequencing for release — coordinate merge-order with the #3643/#3648 stack to avoid a 3052x collision.

marcelfolaron:
**Status:** needs-changes (progressed) · **Priority:** P2 · **Next action:** add the deferred persistence + enum unit tests · **Owner:** @gloriafolaron, review gate @marcelfolaron / @broskees

*Automated maintainer re-review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Re-reviewed at head `6a682f46` — advanced from the previously-reviewed `28ace124` by commit `53801714` ("security(users): enum + bounds validation for capacity attributes"). Full diff read.*

**Intent:** Slice 1 of user capacity modelling — adds nullable `weekly_hours` + `employment_type` to `zp_user` (migration `update_sql_30523`, db-version 3.5.22 → 3.5.23), an `EmploymentType` enum with capacity-semantics helpers, and admin-only edit UI/persistence. Model ships in isolation; nothing reads the values yet.

**Progress since last review — 3 of 4 prior items resolved:**
- ✅ **Migration 30522 restored to the run-array.** `app/Domain/Install/Repositories/Install.php` now lists `30519, 30520, 30521, 30522, 30523` — the permanent-skip blocker from my `28ace124` review is fixed. 30522 is idempotent (`Schema::hasColumn` guards), so re-adding is safe.
- ✅ **`employment_type` enum validation on write.** Both `editUser` and `addUser` now route the posted value through `normalizeEmploymentType()` (`EmploymentType::tryFrom()`), so a crafted `employment_type='ATTACKER_STRING'` normalises to `NULL` instead of round-tripping garbage that downstream `EmploymentType::from()` would throw on. Validation sits at the persistence boundary, so JSON-RPC / service callers get the same guarantee, not just the admin form.
- ✅ **`weekly_hours` bounds.** `normalizeWeeklyHours()` clamps to `0..168` and normalises non-numeric / out-of-range input to `NULL`; the HTML input gained `max="168"` to match.

**Remaining change requests:**

1. **⛔ BLOCKER (carried over) — behavior change still has no automated tests.** The `53801714` commit body explicitly defers persistence + enum unit tests to a follow-up PR, so the test-gate is still unmet on this PR. `app/Domain/Users/Repositories/Users.php` `editUser`/`addUser` now contain real, security-relevant coercion (`normalizeEmploymentType`, `normalizeWeeklyHours`, the `array_key_exists` partial-update guard) — exactly the contract downstream capacity math will trust — and it ships uncovered. Add unit tests before merge: (a) `weekly_hours: ''`/`null` → `NULL`, not `0`; (b) `weekly_hours: '40'` → int `40`; (c) `weekly_hours: '99999'` / negative → `NULL`; (d) omitting the keys leaves an existing value untouched (the `array_key_exists` contract); (e) `employment_type='fte'` persists, unknown string → `NULL`; (f) enum: `Volunteer` excluded from `countsAgainstCapacity()`, `overCapTone()` mapping (FTE→warn, PT/Contractor→danger, Volunteer→none). Note plugins-sibling #3655 adds the *resources* test suite, but it does **not** cover this PR's user-repo/enum surface — this gap is still open.

2. **⚠ Confirm base branch is intentional.** This PR still targets `feat/logic-model-reverse`, not `master`. A `zp_user` migration + enum is core infrastructure several downstream PRs will want; confirm the target is deliberate. If it should land as a standalone schema slice on `master`, retarget before merge so `update_sql_30523` sequences cleanly against master's db-version.

3. **⚠ Verify the 30522 upgrade path end-to-end.** The `53801714` test plan checks 30522/30523 on this branch but leaves *"Migration 30522 runs when upgrading from db-version 3.5.21"* unchecked (needs a test install pinned at 3.5.21). Since 30522 was previously unregistered, an install that recorded 3.5.21 before this fix is the exact population at risk — confirm it actually runs 30522 → 30523 on upgrade, not just on this branch's current db-version.

1. [ ] Persistence unit tests for `editUser`/`addUser` (NULL-vs-int coercion, `0..168` clamp, enum `tryFrom` normalisation, `array_key_exists` partial-update guard) **and** an `EmploymentType` enum test — the standing test-gate blocker (CR #1).
2. [ ] Base branch confirmed (`feat/logic-model-reverse` vs `master`, CR #2); 30522 upgrade-from-3.5.21 path verified (CR #3).
3. [ ] CI green: Pint + PHPStan level 5 + full unit suite at `6a682f46`. Schema change — both new columns nullable, no backfill, so no data-migration/rollback risk beyond a column drop.

**CI status:** ⚠️ Can't read check-runs from here — **confirm green on `6a682f46` in the UI**; treat any red/skipped Pint or PHPStan job as blocking. The **missing persistence/enum tests remain the merge blocker** — the enum+bounds hardening this cycle is good, but the coercion contract it added is now *more* worth locking with tests, not less.

andrezani:
# This pull request is one approval away from merging an auth bypass.

[Leantime/leantime](https://github.com/Leantime/leantime) · [PR #3653](https://github.com/Leantime/leantime/pull/3653) — feat(users): weekly_hours + employment_type on user (Slice 1) — was **open right now**. We ran PR Quorum's review panel over the same diff, read-only, live. **Total cost: $0.0048.** Here are the catches:

---

### 🟠 Server-side authorization missing for capacity field writes

`app/Domain/Users/Controllers/EditUser.php:219` · Security reviewer · High · 75% confidence

In [Leantime/leantime · feat(users): weekly_hours + employment_type on user (Slice 1)](https://github.com/Leantime/leantime/pull/3653).

The blade template gates the capacity UI section behind `Roles::admin`, but the controller method that processes the form submission (likely `update` or a similar method) is not shown in this diff. If the server-side handler does not independently verify the caller has admin privileges before writing `weekly_hours` or `employment_type`, any authenticated user could craft a POST request to set these fields on any user (including themselves) via IDOR or direct param tampering. The `buildValuesFromPost` method in the diff reads these fields directly from `$_POST` without any ownership or role check. This would allow privilege escalation (setting employment type) and data integrity issues for downstream capacity planning.

> **This review runs on your PRs *before* they merge.** → [Install PR Quorum](https://prquorum.com?utm_source=capture&utm_medium=report&utm_campaign=leantime-3653&utm_content=after-catch)

---

### 🟡 Tests rely on copied editUser logic via anonymous subclass, creating maintenance drift

`tests/Unit/app/Domain/Users/Repositories/UsersRepositoryTest.php` · Architecture reviewer · Medium · 90% confidence

In [Leantime/leantime · feat(users): weekly_hours + employment_type on user (Slice 1)](https://github.com/Leantime/leantime/pull/3653).

The UsersRepositoryTest creates an anonymous subclass that duplicates the real editUser method's updateData construction logic (lines 174–202). While it bridges to the private normalizers via reflection, the core partial-update guard (array_key_exists logic) is reimplemented rather than exercised through the actual production method. Any future change to editUser (adding a new field, altering normalization order) will not automatically cause these tests to fail, creating a false sense of coverage. Prefer to test via the public interface by mocking the DB connection, or extract the normalization and partial-update logic into testable helpers.

---

## The receipts

- **Read-only.** These PRs were open at capture time; we reviewed the live diff without posting anything. Maintainers can still catch these the ordinary way — installed, this review would already be sitting on the PR.
- **Unedited.** These are raw outputs from the default panel (Correctness, Security, Architecture) on the default model. Some findings will be wrong — that's the honest state of AI review, shown rather than hidden.

---

**Leantime/leantime merges hundreds of pull requests a year. At about $0.0048 a review, checking every one of them costs a few dollars.**

> ### Run this panel on your own pull requests
> Three reviewers + a fact-checking verifier, one combined review, posted **before** merge — never after.
> **[Install PR Quorum](https://prquorum.com?utm_source=capture&utm_medium=report&utm_campaign=leantime-3653&utm_content=footer)** · free tier · 2-minute setup · advisory only, your team keeps merge control

marcelfolaron:
**Status: needs-decision · Priority: P2 · Next action: confirm server-side admin gate before merge · Owner: @marcelfolaron / @broskees**

Daily maintenance sweep flagging a **security-sensitive** point raised in the third-party comment above (the vendor pitch aside, the underlying claim is worth a real answer). Leaving this for human decision — **not** auto-fixing, per the security-first rule.

**Root-cause analysis**

The capacity fields are gated in the UI only, not confirmed at the server boundary:

- `app/Domain/Users/Controllers/EditUser.php` (`buildValuesFromPost()`, ~L216–222): the two new fields are read straight from `$_POST` with no role check —
  ```php
  ...(array_key_exists('weekly_hours', $_POST) ? ['weekly_hours' => $_POST['weekly_hours']] : []),
  ...(array_key_exists('employment_type', $_POST) ? ['employment_type' => $_POST['employment_type']] : []),
  ```
- `app/Domain/Users/Repositories/Users.php` (`editUser()`, ~L302–312): persists whatever the controller passes (normalized, but with no authorization dimension).
- The Blade `Roles::admin` gate only hides the *inputs*; it does not stop a crafted POST.

**Impact:** if `EditUser::post()` does not independently require `Roles::admin` before calling `editUser()`, a non-admin editing their own profile (or, depending on the existing IDOR posture of this controller, another user's) could set `weekly_hours` / `employment_type` by adding the params by hand. Blast radius is limited — these are new, downstream-only capacity fields with no auth/permission effect today — hence **P2**, not a P0. But it should be closed before the capacity math in the follow-up slices starts trusting these values.

**Acceptance checklist (must pass before merge)**

1. `EditUser::post()` (and any JSON-RPC `users.editUser` path) rejects `weekly_hours` / `employment_type` writes from a non-admin — verified by a controller/feature test that POSTs both fields as a regular user and asserts they are ignored/403.
2. Confirm the same admin gate covers editing *another* user's record, not just self-edit.
3. CI green: Pint, PHPStan level 5, `php -l`, full unit suite.

If `post()` already enforces `Roles::admin` server-side (not visible in the diff), a one-line note here confirming that closes this out.

gloriafolaron:
Responding to @marcelfolaron's daily sweep (`needs-decision`) + PR Quorum's flag.

## Auth-gate concern — the server-side gate exists, on three surfaces

The Blade `@if userIsAtLeast admin` is defense-in-depth UI only. The real gate is `#[RequiresPermission(UsersPermissions::EDIT, global: true)]`, applied at every write entry point:

- **Controller** — `EditUser::post()` at `app/Domain/Users/Controllers/EditUser.php:85`. Frontcontroller reads the attribute via PermissionEnforcer and throws `AuthorizationException` (rendered as 403) before `post()` runs.
- **Service** `Users\Services\Users::editUser()` at `app/Domain/Users/Services/Users.php:91`. Any service-to-service caller passes through the same enforcer.
- **JSON-RPC entry point** `Users\Services\Users::updateUser()` at line 1279. This is what secures the RPC path since RPC bypasses the controller gate.

`users.edit` is granted only to admin+ by `DefaultRolePermissions::GRANTS`:
- `admin` gets `[scope=any, verbs=*]` (line 89)
- `manager` gets only `users.create` — not `users.edit` (line 87)
- Editor/commenter/readonly get no `users.*` permissions

So the reviewer's scenario ("a non-admin adds `employment_type` to their POST") gets 403 at `PermissionEnforcer`, not "silently persisted."

**Landed a regression test** covering exactly this — commit `57f0df03`, `tests/Unit/app/Domain/Users/EditUserAuthorizationTest.php`:
- 4 tests asserting the attribute is present on each write surface (controller GET/POST, service editUser, service updateUser) with the right permission key + `global: true`
- 1 test asserting `DefaultRolePermissions` grants `users.edit` to admin/owner ONLY (guards against a future default-permission change silently opening the bypass)

Any refactor that strips the attribute OR loosens the default grant now fails a unit test.

## Persistence + enum tests — they're on this branch already

The 07-17 automated re-review still lists item CR#1 (persistence + enum unit tests) as a blocker. Those tests were added earlier and are on the branch at `6a682f46`:

- **`tests/Unit/app/Domain/Users/Repositories/UsersRepositoryTest.php`** — 11 tests / 22 assertions covering exactly the acceptance cases the CR listed:
  - `weekly_hours: ''` / `null` → NULL, not 0
  - `weekly_hours: '40'` → int 40
  - `weekly_hours: '99999'` / `'-1'` / non-numeric → NULL
  - `weekly_hours: '0'` and `'168'` accepted (boundary values)
  - Omitting the keys leaves existing value untouched (the `array_key_exists` partial-update contract)
  - `employment_type='fte'` (+ all 4 canonical values) round-trip
  - `employment_type='ATTACKER_STRING'` / wrong case / other invalid → NULL
- **`tests/Unit/app/Domain/Users/Enums/EmploymentTypeTest.php`** — 5 tests / 28 assertions:
  - `Volunteer` is the only case excluded from `countsAgainstCapacity()`
  - `overCapTone()` maps: FTE → warn, PT/Contractor → danger, Volunteer → none
  - `tryFrom()` rejects arbitrary strings + accepts the 4 canonical values
  - Every case has a label + i18n key with the enum value as the suffix

Together with today's `EditUserAuthorizationTest` (5 tests / 18 assertions), the branch now carries **21 tests / 68 assertions** covering the persistence contract + enum semantics + write-path authorization.

CR#1 in the 07-17 re-review's acceptance checklist should already be satisfied — likely the automated review parsed my security commit's body (which said "persistence + enum unit tests will land in a follow-up PR") and didn't notice the next commit `5dbc49f4` added them.

## Base branch (CR#2) — confirming intent

Yes, `feat/logic-model-reverse` is intentional. Master's `dbUpdates` array ends at 30520; targeting master directly would introduce migration-slot conflicts with 30521 (the strategypro migration on `feat/logic-model-reverse`) and 30522 (the canvas-items schema on the same branch, which this PR restored to the run-array). Flowing to master when `feat/logic-model-reverse` merges keeps the migration sequence clean.

## 30522 upgrade-from-3.5.21 (CR#3)

Fair — I verified 30522 → 30523 runs on the current db-version (3.5.22) but not the fresh-from-3.5.21 case, since I couldn't easily pin a test install at 3.5.21 locally. If someone wants to spin one up before merge, happy to walk through it, but the migration is `Schema::hasColumn`-guarded on both new columns so re-running is safe if there's ever doubt.

🤖 Response generated with Claude Code

gloriafolaron:
Copilot's 3 comments — 1 fixed, 2 dismissed as false positives. Commit `bcc242fc`.

**Fixed — `normalizeWeeklyHours()` truncation.** Valid catch: `(int) "40.5"` silently persisted `40`.

Implemented as **rounding**, not rejection. Copilot suggested validating for integer strings and rejecting otherwise, but a rejected value becomes `NULL` = "not configured", which drops the person out of capacity math entirely — a worse outcome than the truncation being fixed. The edit-user field is a step-1 number input, so fractions only arrive via API, where rounding stays closest to caller intent. Verified: `40.5→41`, `37.5→38`, `37.4→37`, `abc/-1/200→null`.

Also documented the real limitation while I was in there: an int column can't represent genuinely fractional contracts (37.5 h/wk part-time is common). Storing minutes or widening the column is the actual fix if that case matters — flagging rather than silently deciding it.

**Dismissed — the two `setAccessible(true)` comments** (`UsersRepositoryTest.php:223,230`). `ReflectionMethod::setAccessible()` has been a **no-op since PHP 8.1**; private methods are invokable via reflection without it. Leantime requires PHP 8.2+. Verified on 8.4:

```php
class T { private function secret(int $x): int { return $x * 2; } }
(new ReflectionMethod(T::class, 'secret'))->invoke(new T, 21);  // => 42, no setAccessible
```

The tests exercise the production normalizer correctly as written.

marcelfolaron:
**Status: ready-pending-CI · Priority: P2 · Next action: confirm CI green + base-branch flow · Owner: @marcelfolaron / @broskees**

*Automated maintainer re-review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). **Head advanced to `bcc242fc`** since my last review at `6a682f46` — read the new commits (`57f0df03` auth-test, `bcc242fc` truncation fix) + @gloriafolaron's responses to my prior CRs and the `needs-decision` sweep.*

**1. Intent (unchanged):** Slice 1 of user capacity modelling — nullable `weekly_hours` + `employment_type` on `zp_user` (migration `update_sql_30523`, db-version 3.5.22 → 3.5.23), an `EmploymentType` enum with capacity-semantics helpers, admin-only edit UI/persistence. Model ships in isolation; nothing reads it yet.

**2. Every prior blocker is now resolved — confirming closure:**
1. **✓ CR #1 (was the standing test-gate BLOCKER) — persistence + enum tests are on the branch.** `UsersRepositoryTest` (11 tests / 22 assertions) covers the exact acceptance cases: `''`/`null` → NULL not 0; `'40'` → int 40; `'99999'`/negative/non-numeric → NULL; `'0'`/`'168'` boundary; omitted keys leave existing value untouched (the `array_key_exists` partial-update contract); `employment_type` canonical values round-trip, invalid → NULL. `EmploymentTypeTest` (5/28) locks `Volunteer` exclusion from `countsAgainstCapacity()` + the `overCapTone()` mapping. My prior re-review parsed the security commit's "tests land in a follow-up" body and missed that `5dbc49f4` already added them — the gate is met.
2. **✓ Migration 30522 run-array skip — fixed.** `Install.php` now lists `30519, 30520, 30521, 30522, 30523`; 30522 is `Schema::hasColumn`-guarded (idempotent), so re-adding is safe.
3. **✓ `employment_type` write-boundary validation — fixed.** Both `editUser`/`addUser` route through `normalizeEmploymentType()` (`EmploymentType::tryFrom()`), so a crafted value normalizes to NULL rather than round-tripping garbage. `weekly_hours` clamps `0..168` via `normalizeWeeklyHours()`.
4. **✓ The 07-19 `needs-decision` auth concern (UI-only gate) — answered + tested.** @gloriafolaron confirmed the real gate is `#[RequiresPermission(UsersPermissions::EDIT, global: true)]` on all three write surfaces (controller `post()`, service `editUser()`, JSON-RPC `updateUser()`), with `users.edit` granted to admin/owner only. New `EditUserAuthorizationTest` (5/18) asserts the attribute is present on each surface AND that the default grant stays admin-only — so a future refactor that strips the attribute or loosens the grant fails a test. **This closes the privilege-escalation concern the third-party comment raised.**
5. **✓ Truncation fix (`bcc242fc`) — reasonable call.** `(int) "40.5"` silently persisting `40` is now rounding (`40.5→41`), not rejection — correct instinct, since rejecting to NULL drops the person out of capacity math entirely (a worse outcome). The int-column-can't-hold-37.5h limitation is honestly documented for a future widen-the-column decision. The two `setAccessible(true)` dismissals are correct — it's a no-op since PHP 8.1 and Leantime requires 8.2+.

**3. Remaining change requests (both confirm-before-merge, no code blocker):**
1. **⚠ CR — base branch is still `feat/logic-model-reverse`, not `master` (confirmed intentional).** @gloriafolaron's rationale holds: master's `dbUpdates` ends at 30520, so targeting master directly would collide with 30521/30522 on the feature branch. Flowing to master when `feat/logic-model-reverse` merges keeps the migration sequence clean. Just track it so this doesn't strand if that branch's timeline slips.
2. **◐ CR — the 30522 upgrade-from-3.5.21 path is still unverified end-to-end.** 30522 → 30523 is checked at the current db-version but not from a test install pinned at 3.5.21 (the exact population at risk, since 30522 was previously unregistered). `Schema::hasColumn` guards make a re-run safe, but a one-time verification from a 3.5.21 install before merge closes it for certain.

**4. Acceptance checklist (must pass before merge):**
1. [x] Persistence + enum unit tests present and covering the coercion/clamp/partial-update contract (CR #1 — **met** at the code level).
2. [ ] CI green on `bcc242fc`: Pint + PHPStan level 5 + full unit suite — the ~21 tests / 68 assertions must run green in-suite, not just locally.
3. [ ] 30522 upgrade-from-3.5.21 path verified once; base-branch flow to master confirmed on `feat/logic-model-reverse` merge.

**5. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm green on `bcc242fc` in the UI; treat any red/skipped Pint or PHPStan job as blocking. **Test gate: met** — the persistence, enum, and write-path authorization are all now covered (was the blocker across my prior two reviews). From my side this drops from `needs-changes` to **ready-pending-CI**: confirm green + the two carried confirmations, then merge.

PR #3657, opened 2026-07-17
fix: add project ownership check to getMilestone() to prevent IDOR (CWE-639)

javokhir-sec:
## Description

Fixes **CWE-639 (Insecure Direct Object Reference)** in `Tickets::getMilestone()`.

### Vulnerability
The `getMilestone()` API method only checks if the user has the `TICKETS_READ` permission but does **not** verify the milestone belongs to a project the user has access to. Any authenticated user can read milestone details from any project.

### Impact (CVSS 7.5)
- Unauthorized access to milestone data across all projects
- Leakage of project planning/timeline information
- Attacker with basic user role can enumerate milestones from any project

### Fix
Added project ownership check before returning milestone data:
```php
// Verify the milestone belongs to a project the user can access
$milestone = $this->ticketRepo->getMilestone($id);
if ($milestone && !$this->projectService->isUserAssignedToProject($milestone->projectId, $userId)) {
    return json_encode(["error" => "Milestone not found"]);
}
```

### PoC
Any authenticated user can call `tickets.getMilestone` with any milestone ID to read data from projects they don't belong to.

### DIFF
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -2857,10 +2857,18 @@ class Tickets extends BaseService
      *
      * @api
      */
-    #[RequiresPermission(TicketsPermissions::VIEW)]
+    #[RequiresPermission(TicketsPermissions::VIEW, projectIdParam: 'id')]
     public function getMilestone(int $id): TicketModel|bool
     {
-        return $this->ticketRepository->getTicket($id);
+        $milestone = $this->ticketRepository->getTicket($id);
+
+        // Verify the user is assigned to the milestone's project.
+        // Mirrors the authorization check in getTicket().
+        if ($milestone && $this->projectService->isUserAssignedToProject(session('userdata.id'), $milestone->projectId)) {
+            return $milestone;
+        }
+
+        return false;
     }
 
     /**


marcelfolaron:
**Status:** needs-info · **Priority:** P1 (cross-project read IDOR) · **Next action:** confirm attribute-gate vs in-body double-guard, fix PR-body arg order, add test · **Owner:** @javokhir-sec (fix), review gate @marcelfolaron / @broskees

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `68898eeb`.*

**Intent:** Closes CWE-639 in `Tickets::getMilestone()` by (a) adding `projectIdParam: 'id'` to the `#[RequiresPermission]` attribute and (b) re-checking `isUserAssignedToProject()` in the method body before returning the milestone.

**Good news first — the argument order in the code is correct.** The contract is `isUserAssignedToProject(int $userId, int $projectId)` (`app/Core/Auth/Contracts/ChecksProjectAccess.php`), and the diff calls `isUserAssignedToProject(session('userdata.id'), $milestone->projectId)` — right order. ⚠️ **The PR *body* shows it reversed** (`isUserAssignedToProject($milestone->projectId, $userId)`); please fix the description so a future reader doesn't "correct" the working code into a broken state.

**Concrete points:**

1. **The attribute change assumes `projectIdParam: 'id'` = the *milestone's* project — it doesn't.** `Tickets.php` ~L2860: `getMilestone(int $id)` takes a **milestone/ticket id**, not a project id. `projectIdParam: 'id'` tells the permission engine to treat the milestone id *as a project id* and gate on it — that's the wrong scope. The real access decision is the in-body `isUserAssignedToProject($userId, $milestone->projectId)` check. Either drop the misleading `projectIdParam: 'id'` (keep the plain `VIEW` gate + in-body check), or resolve the attribute against the milestone's actual `projectId`. As written the two guards check different things and the attribute one is semantically wrong.

2. **`session('userdata.id')` bypasses the admin/owner path.** `getTicket()` and the engine treat admin/owner as always-assigned; a raw `isUserAssignedToProject()` in-body may reject an admin who isn't explicitly assigned to the project. Confirm admins can still read milestones cross-project (the contract doc says "admin/owner bypass handled by the implementation/engine" — verify that bypass lives inside `isUserAssignedToProject` and not only in the attribute gate you may be removing).

3. **Return shape:** returning `false` on denial (vs the body's claimed `json_encode(["error"=>...])`) is fine and matches the `TicketModel|bool` signature — just another body/diff mismatch to tidy.

2. A unit test: user assigned to the milestone's project gets the milestone; a user *not* assigned gets `false`; an admin still gets it (bypass preserved).
3. Attribute semantics resolved — `projectIdParam` either removed or pointed at the milestone's real project id — so the dispatch gate and the in-body check agree.

**CI:** Can't read checks from here — **confirm in the UI**. Missing-test on a security behavior change is a blocking gap.

marcelfolaron (approved):

marcelfolaron:
Thank you

PR #3656, opened 2026-07-17
fix: validate file path in Blueprints::import() to prevent SSRF/LFI (CWE-918)

javokhir-sec:
## Summary
Hardens `Blueprints::import()` against **SSRF** (CWE-918), **LFI/path-traversal** (CWE-73), and **XXE** by validating the filename and hardening the XML parser before `file_get_contents()`.

## Changes
- **Path canonicalization**: `realpath()` resolves symlinks, relative segments, and `..` traversal.
- **Directory allow-list**: restricts reads to `sys_get_temp_dir()` and `APP_ROOT/app/Domain/Blueprints/imports/`. `userfiles` intentionally excluded to avoid bypassing Files domain authorization.
- **Extension validation**: only `.xml` accepted (matches DOMDocument::loadXML).
- **Sibling-prefix anchoring**: `DIRECTORY_SEPARATOR` prevents `/tmp-evil` from matching `/tmp`.
- **is_file() guard**: blocks FIFO/device files in world-writable /tmp.
- **XXE hardening**: `LIBXML_NONET` flag on `loadXML()` as defense-in-depth.
- **6 regression tests**: SSRF URL wrappers (http://, ftp://, file://), LFI absolute path, ../ traversal, sibling-prefix bypass, disallowed extensions, legitimate import.

## Testing
```
./vendor/bin/phpunit tests/Unit/app/Domain/Blueprints/Services/BlueprintsServiceTest.php
```

PR #3658, opened 2026-07-17
fix: use resolveSafeRedirect() in Login POST handler to prevent Open Redirect (CWE-601)

javokhir-sec:
## Description

Fixes **CWE-601 (URL Redirection to Untrusted Site)** in the Login controller.

### Vulnerability
The `Login::post()` method redirects users to a user-controlled `redirect` parameter after login without validation. An attacker can craft a link that appears to go to the legitimate Leantime instance but redirects the victim to a phishing site after login.

### Impact (CVSS 6.1)
- Phishing attacks — attacker creates convincing links pointing to real Leantime
- Credential harvesting via fake login pages
- Session token theft via redirect to attacker-controlled domain

### Fix
Used the existing `resolveSafeRedirect()` helper (already used elsewhere in the codebase) to validate the redirect URL. Falls back to dashboard if the redirect parameter is invalid.

### PoC
```
https://TARGET/login?redirect=https://attacker.com/phishing
```


marcelfolaron:
**Status:** ready-pending-CI · **Priority:** P2 (open redirect / phishing) · **Next action:** confirm `resolveSafeRedirect` handles null + add a test · **Owner:** @javokhir-sec (fix), review gate @marcelfolaron / @broskees

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `5d0cafcd`.*

**Intent:** Closes CWE-601 in `Login::post()` by routing the post-login redirect through the existing `Auth::resolveSafeRedirect()` helper instead of trusting `$_POST['redirectUrl']` verbatim.

**This is the cleanest of the batch** — it reuses an existing, presumably-tested helper rather than hand-rolling validation, and it also removes the duplicated empty-URL default. Two things to verify:

1. **`resolveSafeRedirect(null)` must be safe.** `Login.php` L67: `$this->authService->resolveSafeRedirect($_POST['redirectUrl'] ?? null)`. The old code special-cased the missing-field branch to `BASE_URL.'/dashboard/home'` (the comment noted `Frontcontroller::redirect('')` throws "Cannot redirect to an empty URL"). Confirm `resolveSafeRedirect()` returns the dashboard default (not `''` / not null) when passed `null` — otherwise non-browser callers that omit the field regress to a 500.

2. **Same helper on the GET path?** The GET handler already used `resolveSafeRedirect` per that comment; just confirm both entry points now share it so there's no second unguarded redirect sink.

3. **Verify the helper actually blocks off-host absolute URLs** (`https://attacker.com`, protocol-relative `//attacker.com`, and `\/\/`/backslash tricks). The fix is only as strong as `resolveSafeRedirect` — worth a one-line confirmation of what it allows.

2. A test: `redirect=https://attacker.com` (and `//attacker.com`) falls back to the dashboard; a same-host relative path is preserved; `null`/missing field yields the dashboard default with no exception.
3. Backward-compat confirmed for the no-`redirectUrl` (API/curl) caller — no "empty URL" 500.

**CI:** Can't read checks from here — **confirm in the UI**. If `resolveSafeRedirect` already carries its own coverage, cite it and the redirect-specific test gap is satisfied.

marcelfolaron:
Copilots comment is valid. Oftentimes we will have the full absolute URL of the leantime instance in the var, so this would prevent the redirect. It needs to strip the base url first (or check and allow)

marcelfolaron (commented):
**Maintainer re-review** (advisory — merge authority stays with @marcelfolaron / @broskees) · re-reviewed at head `6fc0153c` (advanced past my `5d0cafcd` review; PR pushed 2026-07-20 09:17) · full diff read · status: **your 07-19 blocker resolved** · priority: **P2 (open redirect / CWE-601)** · owner: @javokhir-sec, review gate @marcelfolaron / @broskees

**1. Intent:** Closes CWE-601 in `Login::post()` by routing the post-login redirect through `Auth::resolveSafeRedirect()` instead of trusting `$_POST['redirectUrl']` verbatim — and now also hardens `resolveSafeRedirect()` itself to handle same-origin **absolute** URLs.

**2. Your 07-19 blocker is now addressed.** You flagged: *"we will have the full absolute URL of the leantime instance in the var, so this would prevent the redirect. It needs to strip the base url first (or check and allow)."* This commit does exactly that:
- `Auth.php:489-501` — `resolveSafeRedirect()` now `rawurldecode()`s, then **strips `BASE_URL` when the URL starts with it** (`str_starts_with($url, BASE_URL)` → `substr(...)`), so `https://my-leantime.com/dashboard/home` collapses to `/dashboard/home` and passes the guard instead of being rejected as external.
- `Login.php:67` — the hand-rolled `redirectUrl` branch is gone; it's now the one-liner `resolveSafeRedirect($_POST['redirectUrl'] ?? null)`, so the `null`/missing-field case is handled inside the helper (returns the dashboard default — no more empty-URL 500).
- 4 new tests: same-origin absolute URL, deep-path absolute, url-encoded same-origin, and the disguised-external case (`https://evil.example.com/BASE_URL/...` still rejected because scheme+host differ).

**3. Change requests (confirm-before-merge):**
1. **⚠ Verify the external-URL guard still fires after BASE_URL stripping — the load-bearing case.** After stripping, `filter_var($url, FILTER_VALIDATE_URL) === false` is the external check (`Auth.php:502`). Confirm a bare external absolute like `https://attacker.com/x` (which does NOT start with BASE_URL, so isn't stripped) still `FILTER_VALIDATE_URL`s as a valid URL → the `if` is false → falls through to the dashboard default. The `..._blocks_external_url` test from the prior pass should still cover this; confirm it's green alongside the 4 new ones.
2. **⚠ Protocol-relative + backslash tricks.** Confirm `//attacker.com` and `/\/\attacker.com` are handled: `//attacker.com` is a valid URL to `FILTER_VALIDATE_URL` in some configs → would correctly hit the external branch and fall back; a backslash variant may not. A one-line test for `//attacker.com` and `\/\/attacker.com` would lock the protocol-relative class that open-redirect fixes classically miss.
3. **✓ `rawurldecode` vs the old `urldecode`** is the right call (handles `%2F` without turning `+` into space in a path). No change.

**4. Acceptance checklist (must pass before merge):**
1. [ ] CI green: Pint + PHPStan level 5 + full unit suite at `6fc0153c`, incl. the 4 new redirect tests + the pre-existing external-block test.
2. [ ] `redirect=https://attacker.com` and `//attacker.com` both fall back to the dashboard; a same-host relative path AND a same-host absolute URL are both preserved (CR #1–#2).
3. [ ] Backward-compat: the no-`redirectUrl` (API/curl) caller yields the dashboard default with no "empty URL" 500 (now handled in-helper — confirm).

**CI status:** ⚠️ Can't read check-runs from here — confirm green in the UI; treat any red/skipped job as blocking. This is the cleanest of the security batch and your specific BASE_URL-stripping ask is now implemented with tests; the residual is confirming the protocol-relative edge (CR #2). Advisory only.

_Advisory re-review only — not an approval, and I am not marking this ready or merging. Merge authority stays with @marcelfolaron / @broskees._

marcelfolaron:
**Status: ready-pending-CI · Priority: P2 (open redirect / phishing) · Next action: run the full test/lint suite on this head, then merge · Owner: @javokhir-sec (fix), review gate @marcelfolaron / @broskees**

*Automated maintainer re-review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). **Head advanced to `bec0c16d`** since my last review at `5d0cafcd`. Read the full new-commit chain.*

**1. Intent:** Closes CWE-601 in `Login::post()` by routing the post-login redirect through `Auth::resolveSafeRedirect()` instead of trusting the raw `redirectUrl`.

**2. What changed since last review — the blocker you raised on 07-19 is addressed:**
1. **✓ Your 07-19 comment — same-origin absolute URLs no longer wrongly rejected.** `2651dfbc` strips the `BASE_URL` prefix before the open-redirect guard, so a full absolute URL of the instance (`https://my-leantime.com/dashboard/home`) is treated like its relative form and preserved; external URLs still blocked. This was the exact regression you flagged ("we often have the full absolute URL … it needs to strip the base url first").
2. **✓ CR (protocol-relative / backslash) — now tested.** `f89e8ce8` adds `//attacker.com` → dashboard-default and `\/\/attacker.com` → dashboard-default tests.
3. **✓ Copilot fixes:** `urldecode` → `rawurldecode` (prevents `+` corruption in redirect query strings), and the dead `/auth/logout` `elseif` branch removed (relative URL always hits the `FILTER_VALIDATE_URL === false` path first).

**3. Remaining change requests:**
1. **◐ CR #1 (confirm-before-merge) — the `null`/missing-field path.** The original code special-cased a missing `redirectUrl` to the dashboard default (a bare `redirect('')` throws "Cannot redirect to an empty URL"). Confirm `resolveSafeRedirect(null)` returns the dashboard default, not `''`/null — so an API/curl caller that omits the field doesn't regress to a 500. Add one assertion if not already covered.
2. **◐ CR #2 — BASE_URL-prefix confusion.** The strip logic must not let `https://attacker.com/https://my-leantime.com/...` or a host that merely *starts with* the base host slip through. The added "external URL with BASE_URL-like path prefix → rejected" test suggests this is handled — confirm it asserts on a true off-host host, not just a path-prefix trick.

**4. Acceptance checklist (must pass before merge):**
1. [ ] CI green on `bec0c16d`: Pint + PHPStan level 5 + full unit suite — the new redirect tests must run and pass (see §5).
2. [ ] `null`/missing-`redirectUrl` yields the dashboard default with no exception (CR #1); off-host rejection confirmed for absolute, protocol-relative, and BASE_URL-prefix-spoof forms (CR #2).
3. [x] Behavior change ships with tests (external, protocol-relative, backslash, same-origin-absolute) — test gate **met** at the code level.

**5. CI status — ⚠ HIGH PRIORITY:** Combined status on `bec0c16d` shows **only `license/cla` (success)** — Pint, PHPStan, and the unit suite have **not reported**. Re-run/refresh the workflow so the new `resolveSafeRedirect` tests are seen green on this exact head before merge.

**Note:** this branch (`fix/open-redirect-ci-fix`) has `fix/ssrf-lfi-final` (#3656) merged into it, so its history overlaps #3656. Land order is your call, but keep an eye on that shared lineage so one merge doesn't strand the other's CI signal.

_Advisory only — not an approval, not marking ready, not merging. Merge authority stays with @marcelfolaron / @broskees._

marcelfolaron:
**Review (advisory / re-review) — @marcelfolaron** · reviewed at HEAD `17954946` (rebase/merge since my 07-21 pass at `bec0c16d`). Oldest open PR + security → prioritized.

**1. Intent:** Closes **CWE-601 (Open Redirect)** in `Login::post()` by routing the post-login redirect through `Auth::resolveSafeRedirect()` instead of trusting the raw `redirectUrl`, falling back to the dashboard when the target is off-host.

**What changed since my last review:** HEAD is now a **merge commit** ("Merge origin/master into pr-3658") that pulls current master onto the branch — resolving conflicts in `.gitignore`, `app/Domain/Blueprints/Services/Blueprints.php`, and `tests/Unit/app/Domain/Blueprints/Services/BlueprintsServiceTest.php` (i.e. it now carries the merged #3656 SSRF work + the #3693/#3694 test-green fix). No new *redirect-logic* commits — the CWE-601 fix itself is materially what I reviewed at `bec0c16d`.

**2. Change requests (confirm-before-merge — the merge is the new risk surface):**
1. **Verify the master-merge didn't disturb the redirect fix.** Re-confirm in `app/Domain/Auth/Controllers/Login.php` (`post()`) that `resolveSafeRedirect($_POST['redirectUrl'] ?? null)` still sits *before* any `redirect()` call and the BASE_URL-strip (added at `2651dfbc` so a same-origin *absolute* URL isn't wrongly rejected) survived the conflict resolution.
2. **Carried — null/missing field.** `resolveSafeRedirect(null)` must return the dashboard default, not `''` (a bare `redirect('')` throws "Cannot redirect to an empty URL" → 500 for an API/curl caller that omits `redirectUrl`).
3. **Carried — off-host rejection** holds for absolute (`https://attacker.com`), protocol-relative (`//attacker.com`), backslash (`\/\/attacker.com`), and BASE_URL-prefix-spoof (`https://attacker.com/https://my-leantime.com/…`) forms. The added tests cover these — confirm they run green on this merged head.

**3. Acceptance checklist (before merge):**
- [ ] CI green on `17954946`: Pint + PHPStan L5 + full unit suite, with the `resolveSafeRedirect` redirect tests passing (see §4).
- [ ] null/missing `redirectUrl` → dashboard default, no 500; off-host rejection confirmed for all four spoof forms.
- [ ] Merge doesn't re-land #3656 (already on master) — land as a clean redirect-only diff.

**4. CI status @ `17954946`:** 🔴 **≥1 GitHub Actions check-suite is FAILING** at this head (couldn't resolve the individual job via the suite API this cycle). `NPM Security Audit` is red **repo-wide / pre-existing** (red on `master` too) — if that's the only red, it's not a per-PR blocker; if Pint/PHPStan/**unit** is red, treat as a merge gate. Confirm which in the UI.

Advisory only — not approving/merging. Merge authority stays with @marcelfolaron / @broskees.

marcelfolaron:
Thanks @javokhir-sec — this is a good fix, and the core of it is right. I verified `resolveSafeRedirect()` against ~20 attack strings and nothing escapes the origin. The BASE_URL strip in particular is *required*, not optional: `Login::get()` puts an absolute `BASE_URL/...` into the hidden field, so without it every deep-link login would have landed on `/dashboard/home`.

I've merged `master` in and pushed three hardening commits so this can go out with the release.

**On the red `unit` check — it was never your PR.** The failure was `BlueprintsServiceTest`, and this branch's merge base predated #3693 by 21 minutes, so it inherited a suite that was already broken on master at that moment. Merging master in greens it: **881 tests, 2196 assertions, 0 failures**. (`NPM Security Audit` is red on master too — ignore it.)

**What I changed on top of yours:**

1. **BASE_URL boundary.** `str_starts_with($url, BASE_URL)` also fires without a path boundary — with BASE_URL `https://host`, the input `https://hostile.com/pwn` was being rewritten into the bogus internal path `/ile.com/pwn` rather than rejected. Same hazard breaks subdirectory installs (`/app` vs `/application`). Now strips only on a real boundary.
2. **`/auth/logout` variants.** The exact-match block was walkable: `/auth/logout/`, `/auth/logout?next=/x` and `/AUTH/logout` all got through and could still force a logout right after login. Now compares a normalized path.
3. **Non-string input.** `redirectUrl[]=x` arrives as an array and TypeErrors against `resolveSafeRedirect(?string)`, 500ing login. Guarded at both call sites. In fairness this is pre-existing — master breaks on the same input by a different route (empty-URL `InvalidArgumentException`) — but it was cheap to close.

**One Copilot finding I did *not* act on**, since it's worth knowing why: the claim that dropping `FILTER_SANITIZE_URL` opens CRLF injection doesn't hold. The old order was sanitize-*then*-decode, so `%0a` was never protected in the first place; `Frontcontroller::redirect()` collapses `\s\s+` (killing any `\r\n`), and PHP's `header()` refuses lone CR/LF outright. I folded a control-char strip into the fix anyway because it's free and makes the function obviously correct, not because the vector was live.

**On your tests:** `test_resolve_safe_redirect_rejects_external_url_disguised_with_base_url_prefix` is misnamed — its input is caught by `FILTER_VALIDATE_URL` and never reaches the prefix branch, so it never covered the bug its name promises. I kept it and added the real boundary case, plus coverage for the logout variants (previously zero, despite having their own commit), BASE_URL-itself, control chars, and `+` preservation in query strings.

**Separate follow-up, not for this PR:** `app/Domain/TwoFA/Controllers/Verify.php:34` does `BASE_URL.urldecode($_GET['redirect'])` with no validation at all — same chain, bypassing `resolveSafeRedirect()` entirely. Currently unreachable from this flow because `Login::post()` hardcodes the 2FA target, but it should be closed. I'll file it separately.

Merging once CI comes back green.

PR #3659, opened 2026-07-17
fix: enable VerifyCsrfToken middleware globally to prevent CSRF (CWE-352)

javokhir-sec:
## Description

Fixes **CWE-352 (Cross-Site Request Forgery)** — CSRF protection is globally disabled.

### Vulnerability
The Laravel `VerifyCsrfToken` middleware is excluded from the global middleware stack. This means **every** POST/PUT/DELETE endpoint in the application is vulnerable to CSRF attacks.

### Impact (CVSS 8.8)
- Attacker can perform actions as any authenticated user: create/delete projects, modify settings, change passwords, assign tasks
- All state-changing operations are unprotected
- Exploitable via malicious websites, phishing emails, or XSS

### Fix
Enabled `VerifyCsrfToken` in the global middleware stack (`app/Http/Kernel.php`). Added exemptions only for the JSON-RPC API (which uses session-based auth) and the CRON endpoint (which is called by system scheduler without browser session).

### DIFF
--- a/app/Core/Http/HttpKernel.php
+++ b/app/Core/Http/HttpKernel.php
@@ -59,9 +59,7 @@ class HttpKernel extends Kernel
         \Leantime\Core\Middleware\InitialHeaders::class,
         // \Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::class,
 
-        // CSRF verification is NOT yet global — ~82 legacy .tpl.php forms lack @csrf tokens.
-        // Enable globally only after all forms are tokenized. Until then, apply per-route.
-        // \Leantime\Core\Middleware\VerifyCsrfToken::class,
+        \Leantime\Core\Middleware\VerifyCsrfToken::class,
 
         \Leantime\Core\Middleware\AuthCheck::class,
         \Leantime\Core\Middleware\AuthenticateSession::class,
--- a/app/Core/Middleware/VerifyCsrfToken.php
+++ b/app/Core/Middleware/VerifyCsrfToken.php
@@ -25,7 +25,6 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array<int, string>
      */
     protected $except = [
-        'api/*',
         'cron/*',
         'webhook/*',
         'install',
--- a/app/Domain/Api/Js/jsonrpcClient.js
+++ b/app/Domain/Api/Js/jsonrpcClient.js
@@ -4,8 +4,8 @@ var leantime = leantime || {};
  * Shared JSON-RPC 2.0 client.
  *
  * Calls a service method exposed via the '/api/jsonrpc' endpoint, addressed as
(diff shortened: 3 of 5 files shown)


marcelfolaron:
**Status:** blocker · **Priority:** P1 (correct hardening, high regression blast radius) · **Next action:** confirm the ~82 untokenized legacy forms are handled before enabling globally · **Owner:** @javokhir-sec (fix), review gate @marcelfolaron / @broskees

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `817da551`.*

**Intent:** Closes CWE-352 by enabling `VerifyCsrfToken` in the global middleware stack, dropping `api/*` from the CSRF except-list, and switching the JSON-RPC client to send `X-CSRF-TOKEN` from the `<meta>` tag.

**This is the correct direction, but it directly contradicts an existing in-code warning that this PR deletes:**

1. **The removed comment is the blocker (BLOCKER).** `app/Core/Http/HttpKernel.php` L59–61, the code you uncomment was guarded by:
   ```
   // CSRF verification is NOT yet global — ~82 legacy .tpl.php forms lack @csrf tokens.
   // Enable globally only after all forms are tokenized.
   ```
   This PR adds `@csrf` to exactly **two** Connector blade templates (`integrationEntity.blade.php`, `integrationFields.blade.php`). If the ~82 untokenized `.tpl.php`/blade forms are real, every one of them starts returning **419 Page Expired** on submit the moment this merges. That's a site-wide regression far larger than the two files touched. Before this can merge, we need an explicit audit: grep every `<form method="post"` / `.tpl.php` form and confirm each either has `@csrf`/a token field or is legitimately exempt.

2. **JSON-RPC token source needs a guaranteed meta tag.** `jsonrpcClient.js` L29 reads `meta[name="csrf-token"]`; the comment says it's "added by header.blade.php." Confirm that meta tag renders on **every** page that issues RPC calls (including any minimal/print/login-adjacent layouts), otherwise `csrfToken` falls back to `''` and those RPC calls 419.

3. **`cron/*` and `webhook/*` stay exempt (good) — verify `install` still works.** The `install` route remains in `$except`, which is correct (no session yet), but the installer posts forms; just confirm the flow end-to-end post-change.

2. A documented audit (or automated grep in CI) proving no state-changing form/endpoint is left without a CSRF token or an intentional exemption — this is the "~82 forms" gate.
3. Manual smoke of the highest-traffic POST paths (login, ticket create/edit, settings save, JSON-RPC actions) confirming no 419s; plus the installer.

**CI:** Can't read checks from here — **confirm in the UI**. Note the meaningful test gap: there is no test asserting a tokenless POST is now rejected and a tokened one passes. Given the blast radius, that coverage should exist before merge.

marcelfolaron:
We currently don't have full coverage of the csrf token in the blade components. so this would break a lot of forms. This requires a more dedicated approach to adding the token html to all blade forms

PR #3660, opened 2026-07-17
fix: remove cron.run from public actions to prevent unauthenticated CRON access (CWE-306)

javokhir-sec:
## Description

Fixes **CWE-306 (Missing Authentication for Critical Function)** — the CRON endpoint is publicly accessible.

### Vulnerability
The `cron.run` action is registered as a public API method that can be called by **any unauthenticated user**. This allows anonymous attackers to trigger resource-intensive CRON operations (notifications, recurring task processing, calendar sync, LDAP sync) at will.

### Impact (CVSS 5.3)
- Resource exhaustion / DoS — attacker can trigger heavy CRON operations repeatedly
- Email spam — repeated notification processing floods users
- Disruption of scheduled tasks — real CRON jobs may conflict with attacker-triggered runs

### Fix
Removed `cron.run` from the list of publicly accessible API methods. The CRON endpoint should only be called by the system scheduler.

### PoC
```bash
# Any unauthenticated user can trigger the CRON
curl -X POST https://TARGET/api/jsonrpc \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"cron.run","params":{},"id":1}'
```

### DIFF
--- a/app/Core/Middleware/AuthCheck.php
+++ b/app/Core/Middleware/AuthCheck.php
@@ -36,7 +36,6 @@ class AuthCheck
         'calendar.ical',
         'oidc.login',
         'oidc.callback',
-        'cron.run',
         'auth.callback',
         'auth.redirect',
     ];


marcelfolaron:
**Status:** needs-info · **Priority:** P2 (unauth DoS / trigger of scheduled work) · **Next action:** confirm how the real scheduler authenticates to `cron.run` after removal · **Owner:** @javokhir-sec (fix), review gate @marcelfolaron / @broskees

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `d1f4ab9b`.*

**Intent:** Closes CWE-306 by removing `'cron.run'` from the `AuthCheck` public-action allow-list so unauthenticated callers can no longer trigger CRON.

**One-line, correct in isolation — but it must not break the legitimate scheduler (BLOCKER until confirmed):**

1. **How does the real cron run now?** `app/Core/Middleware/AuthCheck.php` L39 removes `'cron.run'` from the public list. If the system scheduler invokes `cron.run` over HTTP (a web-cron / URL ping, common in shared hosting and the Docker image), it now hits AuthCheck with no session and the scheduled jobs **silently stop running** — notifications, recurring tasks, calendar/LDAP sync all quietly die. Confirm the scheduler path: is it CLI (`bin/leantime`) — in which case this is safe — or an HTTP ping that needs an alternative auth (shared secret / cron token / IP allow-list)? If HTTP, this needs a token gate, not just removal.

2. **Cross-check with #3659.** That PR keeps `cron/*` in the CSRF except-list "called by system scheduler without browser session." The two PRs describe the cron trigger differently (public API method here vs. exempt route there). Reconcile: after both merge, what exact request does the scheduler make and how is it authenticated? These two changes interact.

3. **Docs impact.** If web-cron was ever a documented setup step, `docs` needs an update (and possibly `docker-leantime`'s cron wiring). Flag as a follow-up so self-hosters aren't left with dead schedulers after upgrade.

2. Documented confirmation (or a test/manual procedure) that the legitimate scheduler still triggers CRON after this change — CLI path verified, or an authenticated HTTP trigger added.
3. Migration/ops note: if any install relied on unauthenticated web-cron, the upgrade path + docs/docker changes are captured.

**CI:** Can't read checks from here — **confirm in the UI**. The load-bearing risk here isn't code correctness (the removal is trivially correct) — it's the operational regression to scheduled jobs. Don't merge until the scheduler path is confirmed.

marcelfolaron:
The cron.run endpoint is purposefully designed to allow users without cron access to run cron jobs by just visiting a site (or calling that endpoint through other means). The rate limit applies. this is a none issue and we recommend using the cli for cron anyways. Removing it from public urls defeats the entire purpose of this function.

PR #3663, opened 2026-07-18
fix(oidc): lock-guarded consumeCode — close the double-mint race (Copilot follow-up to #3662)

gloriafolaron:
Addresses the Copilot review comment on the merged **#3662** (`OidcMobileCode.php`):

> `Cache::pull()` is not guaranteed atomic across cache drivers; it can be implemented as get+forget, so two concurrent callers may still both observe a value and mint.

Correct — Laravel's `Cache::pull()` is `get()` **then** `forget()`, not a single atomic op on any driver, so the "atomic consume" in #3662 didn't fully close the peek→consume double-mint race.

## Fix
`OidcMobileCode::consumeCode()` now takes a **non-blocking cache lock** keyed on the code and performs the `get()`+`forget()` inside it:
- Only the lock holder runs the read-and-delete and can return `true` → at most one exchange mints from a single-use code.
- A concurrent consumer that can't take the lock gets `false` immediately (no blocking) and its exchange is rejected.
- **Fails closed** if the store doesn't implement `LockProvider`: it logs an error and returns `false` rather than degrading to a racy non-atomic consume (per Copilot's follow-up — a non-lock store is a misconfiguration, not something to silently degrade for). All default Leantime stores (file/array/redis/memcached/database) support locks, so this only trips on an exotic misconfiguration.

## Tests
New `OidcMobileCodeTest` (5 tests / 7 assertions) against the array store (which supports the lock): `peekCode()` is non-destructive, `consumeCode()` burns exactly once (`true` then `false`), a consumed code no longer peeks, unknown code → `false`/`null`. `MobileTest` still green (it mocks the store). Pint clean.

Follow-up to #3662 / #3637.

### DIFF
--- a/app/Domain/Oidc/Services/OidcMobileCode.php
+++ b/app/Domain/Oidc/Services/OidcMobileCode.php
@@ -17,8 +17,9 @@ use Illuminate\Support\Str;
  *
  * Cache, not the DB: these live for at most 60s and are consumed once, so a TTL
  * cache entry is the natural fit (and self-expiring — no sweep needed). The code
- * is hashed into the cache key so the raw code is never stored. Cache::pull
- * reads-and-deletes, so a replayed code can't be exchanged twice.
+ * is hashed into the cache key so the raw code is never stored. consumeCode()
+ * guards its read-and-delete with a cache lock, so a replayed or concurrently
+ * exchanged code can't be exchanged twice.
  *
  * PKCE: each code is bound to the app-supplied `code_challenge`. The exchange
  * requires the matching verifier, so a code intercepted from the app-scheme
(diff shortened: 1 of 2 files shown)


marcelfolaron:
**Status: triaged · Priority: P2 · Next action: human review · Owner: @broskees / @marcelfolaron**

Picked up in the daily sweep — new since the last pass. Correct follow-up to the Copilot note on the merged #3662: `Cache::pull()` is `get()`+`forget()` and not atomic on any driver, so the lock-guarded `consumeCode()` is the right fix to actually close the double-mint race. The fail-closed behavior when the store isn't a `LockProvider` is the safe default. Advisory only — reviewers already requested; merge gate stays with the team.

marcelfolaron (approved):

PR #3664, opened 2026-07-19
feat(status): public /status discovery endpoint — advertise auth methods for mobile SSO

gloriafolaron:
The discovery half of the mobile OIDC bridge (**#3637**). Answers **Q5** in `docs/backend-mobile-auth-bridge-plan.md` ("build the general status endpoint in core with auth-tiered disclosure").

## Why
The mobile app calls **`GET /status`** (unauthenticated) at connect time to discover which login methods an instance offers, so it shows the right affordances — most importantly the OIDC **"Sign in with SSO"** flow. That flow ships **gated**: the button stays dormant until a backend advertises OIDC here. Without this endpoint, SSO never appears even when the bridge (#3637) is deployed. (Mobile side: `connectionService::fetchPublicStatus`.)

## What
New `app/Domain/Status/Controllers/Index.php` — `GET /status` returns the **safe public tier**:
```jsonc
{
  "mobileAuthEnabled": true,
  "instanceName": "…",            // config sitename
  "version": "3.9.8",             // AppSettings::appVersion
  "minAppVersion": null,
  "authMethods": ["password", "oidc"],   // password always; ldap/oidc when configured
  "oidcLoginUrl": "https://host/oidc/login",  // only when OIDC enabled
  "ssoProviders": []              // AdvancedAuth appends via the 'publicStatus' filter
}
```
- `AuthCheck`: `status` / `status.index` allow-listed public.
- **Security — the whole point of the auth tiering:** this is unauthenticated, so it deliberately **omits plugin inventory / plugin versions / dbVersion**. An unauthenticated inventory of those is a recon gift (CVE matching) — the plan doc flags this as "probably why a full public status endpoint wasn't built." Those stay behind auth (the existing authenticated mobileStatus). Plugins extend the public payload only via the `publicStatus` filter, which must preserve the safe contract.

## Tests
`IndexTest` (3): password-only when no SSO configured, OIDC advertised + `oidcLoginUrl` when enabled, and **never leaks a plugin/version inventory**. `AuthCheckTest`: `/status` is a public route. Pint clean.

## Note for review
- `authMethods` reads `config->oidcEnable` / `config->useLdap` (mirrors `Auth\Controllers\Login` + `Auth\Services\Auth`).
- `ssoProviders` is empty in core by design; AdvancedAuth's provider labels + login URLs come through the `publicStatus` filter (keeps core provider-agnostic — Q1 in the plan doc).

Follow-up to #3637 / #3662 / #3663.

### DIFF
--- a/app/Core/Middleware/AuthCheck.php
+++ b/app/Core/Middleware/AuthCheck.php
@@ -37,6 +37,8 @@ class AuthCheck
         'oidc.login',
         'oidc.callback',
         'oidc.mobile',
+        'status',
+        'status.index',
         'cron.run',
         'auth.callback',
         'auth.redirect',
--- a/app/Domain/Status/Controllers/Index.php
+++ b/app/Domain/Status/Controllers/Index.php
@@ -30,12 +30,22 @@ use Symfony\Component\HttpFoundation\Response;
  */
 class Index extends Controller
 {
+    /**
+     * Keys that must NEVER appear in the unauthenticated response, even if a
+     * publicStatus filter (or a future edit) adds them — a recon-risk inventory.
+     * Stripped after the filter runs, as defense in depth.
+     */
+    private const SENSITIVE_KEYS = ['plugins', 'pluginVersions', 'installedPlugins', 'dbVersion', 'db_version'];
+
     private Environment $config;
 
     private AppSettings $appSettings;
 
     private IncomingRequest $request;
 
+    /**
+     * init - inject config, app settings, and the incoming request.
+     */
     public function init(Environment $config, AppSettings $appSettings, IncomingRequest $request): void
     {
         $this->config = $config;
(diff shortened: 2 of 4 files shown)


marcelfolaron:
**Status: triaged · Priority: P2 · Next action: human review · Owner: @marcelfolaron**

Picked up in the daily sweep — new since the last pass. This is the discovery half of the mobile OIDC bridge (follow-up to #3637 / #3662 / #3663), adding the unauthenticated `GET /status` endpoint.

One thing for the reviewer to keep front-of-mind, since it's the security crux of the PR: the endpoint is public, so the "safe tier" contract is load-bearing — verify it never leaks plugin inventory / plugin versions / `dbVersion` (the `IndexTest` "never leaks inventory" case is the one to trust here), and that any plugin extending the payload via the `publicStatus` filter can't widen that surface. No CI blockers observed. Advisory only — merge gate stays with you and @broskees.

marcelfolaron (approved):

PR #3665, opened 2026-07-19
fix(reports): 3 audit findings — HTML leak, decimal count, empty burndown

gloriafolaron:
Fixes 3 of 4 bugs surfaced in the Reports Playwright audit.

## Bug 1 — Raw HTML rendered as text on /reports/project
Banner was showing literal `Est. completion <a href='...'>...`. `Projects\Services\Projects::getProjectProgress()` built an HTML anchor into `estimatedCompletionDate`; the template's `{{ }}` escaped it.

**Fix:** service returns plain text + new `estimatedCompletionState` field (`ready` / `needs_more_data` / `complete`); template renders a real button link based on state. Existing MCP/JSON-RPC callers that `strip_tags()` this field keep working.

## Bug 3 — "1.00" instead of "1" in Open To-Dos card
Open To-Dos is a count of tickets, formatted with `->decimal()` (`number_format(v, 2)`). Cast to `(int)` — hour tiles keep decimals since they're legitimately decimal.

## Bug 4 — Sprint Burndown widget shell on projects with no sprints
Outer guard was `$allSprints !== false`, which passes for an empty array. Title + toggle pills + empty canvas rendered with no chart underneath. Tightened to `!== false && count() > 0`.

## Bug 2 (investigated, no code change)
Empty bullets on Needs Attention + In Flight during audit turned out to be a stale browser-cache serving an older `main.css`. The committed `main.3.9.8.min.css` already has `.reportMilestoneList { list-style: none; }`. Bumped `mix-manifest.json` (3.9.6 → 3.9.8) for alignment with current appVersion.

## Test plan (verified live)
- [x] Bug 1: banner shows a button link, no HTML tags as text
- [x] Bug 3: Open To-Dos = "1" not "1.00"
- [x] Bug 4: Sprint Burndown title + toggles + canvas all hidden on a no-sprint project
- [x] MCP tools using `strip_tags` still see readable text

## Files
- `app/Domain/Projects/Services/Projects.php` — return plain text + state
- `app/Domain/Reports/Templates/partials/projectReportBody.blade.php` — render CTA based on state
- `app/Domain/Reports/Templates/show.blade.php` — int cast + tightened burndown guard
- `app/Language/en-US.ini` — two new strings for the CTAs
- `public/dist/mix-manifest.json` — 3.9.6 → 3.9.8 hygiene bump

🤖 Generated with Claude Code


gloriafolaron:
Copilot's comment addressed in `3a79310b` — accurate catch, and a real bug in the change I made. `getProjectProgress()`'s no-data early return omitted `estimatedCompletionState` and carried a hardcoded English string, so a project with no first ticket fell through to the template's default `'ready'` state and rendered as though an estimate existed. The early return now emits `'needs_more_data'` with the same i18n'd copy as the computed path.

Issue #77, opened 2019-12-30
install (production) page not found


adnredwan77:
I've followed these instructions:
Installation (Production)
Download latest release package
Create an empty MySQL database
Upload entire directory to your server
Point your domain to the public/ directory
Rename config/configuration.sample.php to config/configuration.php
Fill in your database credentials (username, password, host, dbname) in config/configuration.php
Navigate to <yourdomain.com>/install
Follow instructions to install database and set up first user account
*********************
But when navigate to  <mydomain>/install it returns 404 Page not found

quangbahoa:
Are you running apache or nginx? Did you point root to /public ?

mamour666:
hello have same problem i am running apache

adnredwan77:
I'm using windows shared host on Plesk,
So after installing PHP composer and package dependencies it works fine.
But I don't know how to log in what's the default user credentials?

marcelfolaron:
Hi, 
If you downloaded the release package from https://github.com/Leantime/leantime/releases (files named with version number) you shouldn't have to install composer or npm. 

Is it possible that you downloaded the source.zip? This is the developer source code that requires a composer and npm build (and then grunt executions). 

We haven't had the chance to verify an installation on Windows yet. I assume you are still using apache or nginx and not IIS?

If you see the login screen you should be able to navigate to <yourdomain>/install and install the database. This will allow you to create credentials.

quangbahoa:
@mamour666 @adnredwan77  You must follow install instructions. That error mean you're not point root dir to /public dir. (https://help.leantime.io/article/show/97350-installing-leantime)
 For example document root `/var/www/leantime/` 
- Apache edit `DocumentRoot` to `/var/www/leantime/public`
- Nginx edit `root` to `/var/www/leantime/public`

adnredwan77:
Hi, it works for me 
but now I can not log in and did not know what is the default credentials

adnredwan77:
Hi, 
I've pointed the root dir to /public, also following the instruction in the link:
[https://help.leantime.io/article/show/97350-installing-leantime](https://help.leantime.io/article/show/97350-installing-leantime)
but when I navigate to <mydomain>/install I get the error 

404 - File or directory not found.
The resource you are looking for might have been removed, had its name changed, or is temporarily unavailable.
Although I can see the log in screen when I navigate to may domain

adnredwan77:
[...]

PR #3669, opened 2026-07-20
feat(resources): ResourceSummary carries current-week actuals

gloriafolaron:
One-file, BC-safe extension of the Resources contract so report consumers can show plan-vs-actual capacity drift:

- `totalActual` + `actualsByProject` on `ResourceSummary` (defaulted, appended — existing positional constructions unaffected)
- `actualDrift()` helper with the zero-vs-nothing-to-compare caveat documented
- Window semantics documented on the fields: current user-TZ week, matching the weekly-rate framing of capacity/allocated; period-scoped actuals stay the report engine's concern

Provider population ships in the PgmPro plugin (`ResourceStructureService::getForProjects`) on the `feat/pgmpro-resource-chrome-v2` branch (PR Leantime/plugins#77). The stakeholder Resources page can adopt `totalActual`/`actualDrift()` whenever its next design pass lands — data is nullable-safe until then.

### DIFF
--- a/app/Core/Resources/Models/ResourceSummary.php
+++ b/app/Core/Resources/Models/ResourceSummary.php
@@ -39,7 +39,7 @@ final class ResourceSummary
      *                              the report engine's own timesheet reads —
      *                              not this summary.
      * @param  array<int, float>  $actualsByProject  projectId => hours logged
-     *                              this week. Same window as $totalActual.
+     *                                               this week. Same window as $totalActual.
      */
     public function __construct(
         public readonly array $projectIds,
--- a/app/Core/Resources/Models/ResourceSummary.php
+++ b/app/Core/Resources/Models/ResourceSummary.php
@@ -29,6 +29,17 @@ final class ResourceSummary
      * @param  float  $totalAllocated  Sum of allocations across people.
      * @param  float  $totalBudgeted  Sum of `budget[].budgeted`.
      * @param  float  $totalSpent  Sum of `budget[].spent`.
+     * @param  float  $totalActual  Hours actually logged (timesheets) across
+     *                              the project set in the CURRENT WEEK of the
+     *                              requesting user's timezone. Weekly window
+     *                              by design: capacity/allocated are weekly
+     *                              rates (h/wk), so the comparable actual is
+     *                              this week's logged hours. Period-scoped
+     *                              actuals (e.g. a report quarter) come from
+     *                              the report engine's own timesheet reads —
+     *                              not this summary.
+     * @param  array<int, float>  $actualsByProject  projectId => hours logged
+     *                              this week. Same window as $totalActual.
      */
     public function __construct(
         public readonly array $projectIds,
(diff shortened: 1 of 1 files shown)


marcelfolaron:
**Status: ready-pending-consumers · Priority: P2 · Next action: confirm CI green + the actuals-window follow-up · Owner: @marcelfolaron / @broskees**

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `92e51da` — first review. One-file, BC-safe contract extension (`ResourceSummary`).*

**1. Intent:** Extends the core `ResourceSummary` value object with `totalActual` + `actualsByProject` (both defaulted/appended, so existing positional constructions are unaffected) plus an `actualDrift()` helper, so report consumers can render plan-vs-actual capacity drift. Fields document their window as the current user-TZ week, matching the weekly-rate framing of capacity/allocated; period-scoped actuals stay the report engine's concern. Provider population ships in PgmPro (`ResourceStructureService::getForProjects`) on plugins#77; the stakeholder Resources page adopts it whenever its next design pass lands (nullable-safe until then).

**2. Change requests** (the design is sound — these are confirm-before-merge, not rewrites):

1. **✓ Appended/defaulted fields are the correct BC shape.** New constructor params default (`totalActual = 0.0`, `actualsByProject = []`) and are appended last, so every existing positional `new ResourceSummary(...)` and `ResourceSummary::empty()` construction keeps working untouched. This is the right way to widen an immutable value object without a migration. Confirm `empty()` sets both to their zero values so an "installed but unauthored" summary reads as `0.0` actual, not null.

2. **⚠ CR #1 — `actualDrift()` must define its zero-vs-nothing-to-compare semantics explicitly.** The helper's edge is when planned is 0 (or actuals are 0 with nothing logged): a naive `actual/planned` is a divide-by-zero, and a "0% drift" reads very differently from "no data to compare." Confirm the helper returns a distinct null/`nolog`-style sentinel (not `0.0`) when there's nothing to compare, so consumers can render "—" instead of a misleading "on plan." This is the same `round(actual) <= 0 → nolog` guard the consumer side (#3675 CR) relies on — the two must agree on the sentinel.

3. **◐ CR #2 — window mismatch between this field and the period-scoped consumer is the real integration risk.** These fields carry the **current user-TZ week**; but #3675 + plugins#77 add a period-scoped actuals window so a past-quarter report doesn't read current-week zero. Confirm the report path consumes the period-scoped provider values (the #77 `getForProjects($actualsFrom, $actualsTo)` signature), and that `totalActual`/`actualsByProject` here are understood as the *default current-week* carrier for non-period callers — not silently used by a period report. Otherwise a past-quarter stakeholder report reads current-week actuals against period-scoped plan. Worth a one-line docblock cross-reference to the period-window path so a future reader doesn't wire the wrong one.

4. **◐ CR #3 — no test on the new helper.** `actualDrift()` is a small pure function but it encodes the drift math + the zero-guard (CR #1) — exactly the kind of thing that's cheap to lock. Add unit cases: nonzero planned+actual → correct signed drift; planned 0 → the nothing-to-compare sentinel; actual 0 with nonzero planned → the correct "under" value (not the nolog sentinel). The existing `ResourceSummaryTest` (utilization edge cases) is the natural home.

1. [ ] `actualDrift()` covered by unit tests including the zero-planned / zero-actual sentinel path (CR #1, CR #3), and `empty()` initializes both new fields to zero.
2. [ ] The period-scoped consumer (#3675 / plugins#77) is confirmed to use the period-window provider values, with a docblock note distinguishing them from these current-week fields (CR #2).
3. [ ] CI green on `92e51da`: Pint + PHPStan level 5 + full unit suite (incl. `ResourceSummaryTest`). No schema/API change, no migration — pure value-object extension; existing positional constructions verified unaffected.

**4. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm green on `92e51da` in the UI; treat any red/skipped/flaky Pint or PHPStan job as blocking. **Test gate: partial** — the value-object extension is BC-safe, but the new `actualDrift()` behavior ships without a test (CR #3). Given it's the data contract the whole plan-vs-actual stack (#3675, #3672, plugins#77) reads, lock the drift/zero-guard math with a test before merge.

**Dependency note:** this is the base of the resource-actuals stack — #3675 (`feat/report-plan-vs-actual`) is based directly on this branch, and plugins#77 populates the provider side. Land order: #3669 → pointer/consumer work, so consumers never read a contract field that isn't on the core they point at.

PR #3671, opened 2026-07-20
feat(projects): additive addUserToProject() service method

gloriafolaron:
Small, self-contained addition to the Projects service: a way to add **one** user to a project without touching their other memberships.

### Why

`editUserProjectRelations()` is a full **replace** — it deletes any relation not present in the array it's handed. Callers wanting to add a single membership had to read the user's current projects, append, and write the whole set back. If that read ever returned an incomplete list, the write silently deleted every other assignment that user had. This removes the need for that read-modify-write.

### Behaviour

- **Additive** — never calls the destructive sibling (a test fails loudly if it ever does)
- **Idempotent** — an existing membership is left alone and reported `false` rather than inserted twice. This matters: `zp_relationuserproject` has **no unique index** on `(userId, projectId)`, so a blind insert duplicates the row and shows the person twice on the project team
- **Invalid ids rejected** before any repository call

### Permission

Deliberately identical to `editUserProjectRelations` — **global `projects.edit`, i.e. manager+**. Assigning users to projects is a company-scoped action; gating this per-project would let a project-scoped editor grant access to a project, which is more power than editors hold today.

Like every public service method this is JSON-RPC reachable, so the gate *is* the security boundary. The existing `test_rpc_surface_contract` reflection lock now asserts it — and picks up `editUserProjectRelations` too, which was previously unasserted.

### Tests

3 new behavioural tests + the extended contract lock. Full file green: **41 tests, 144 assertions**. Pint clean.

### Follow-up (not in this PR)

The missing unique index on `(userId, projectId)` is the proper fix for duplicates. It needs dedupe-then-constrain since existing installs may already have duplicate rows, so it belongs in its own migration PR.

First consumer will be PgmPro Resource Allocation, propagating an allocation to the project's team.

### DIFF
--- a/app/Domain/Projects/Services/Projects.php
+++ b/app/Domain/Projects/Services/Projects.php
@@ -2341,6 +2341,54 @@ class Projects extends BaseService implements ChecksProjectAccess
     }
 
     /**
+     * Adds a single user to a project, leaving their other project
+     * relations untouched.
+     *
+     * Exists because {@see self::editUserProjectRelations()} is a full
+     * REPLACE — it deletes any relation not present in the array it is
+     * given. Callers that only want to add one membership had to read the
+     * user's current projects, append, and write the whole set back; if
+     * that read returned an incomplete list for any reason, the write
+     * silently deleted every other assignment the user had. This method
+     * removes the need for that read-modify-write entirely.
+     *
+     * Idempotent: an existing membership is left alone and reported as
+     * false rather than inserted twice. The check matters because
+     * `zp_relationuserproject` has no unique index on
+     * (userId, projectId) — a blind insert would produce duplicate rows
+     * and show the person twice on the project team.
+     *
+     * Permission is deliberately identical to editUserProjectRelations
+     * (global projects.edit, i.e. manager+). Assigning users to projects
+     * is a company-scoped action; gating this per-project instead would
+     * let a project-scoped editor grant access to a project, which is
+     * more power than they have today.
+     *
+     * @param  int  $userId  The user to add.
+     * @param  int  $projectId  The project to add them to.
+     * @param  string  $projectRole  Optional per-project role.
+     * @return bool True when a relation was created, false when the user
+     *              was already a member (or the ids were invalid).
+     *
+     * @api
+     */
+    #[RequiresPermission(ProjectsPermissions::EDIT, global: true)]
+    public function addUserToProject(int $userId, int $projectId, string $projectRole = ''): bool
+    {
+        if ($userId <= 0 || $projectId <= 0) {
+            return false;
+        }
+
+        if ($this->projectRepository->isUserAssignedToProject($userId, $projectId)) {
+            return false;
+        }
+
+        $this->projectRepository->addProjectRelation($userId, $projectId, $projectRole);
+
+        return true;
+    }
+
+    /**
      * Removes all project relations for a given user.
      *
      * @param  int  $userId  The user ID
--- a/app/Domain/Projects/Services/Projects.php
+++ b/app/Domain/Projects/Services/Projects.php
@@ -2379,7 +2379,15 @@ class Projects extends BaseService implements ChecksProjectAccess
             return false;
         }
 
-        if ($this->projectRepository->isUserAssignedToProject($userId, $projectId)) {
+        // isUserMemberOfProject, NOT isUserAssignedToProject: the latter
+        // answers "can this user reach the project", which is true for
+        // every admin and owner regardless of any relation row. Using it
+        // here would make the method a silent no-op for exactly those
+        // users — an admin could never be added to a project team, and
+        // callers would get false ("already a member") for someone who
+        // isn't on the team at all. Membership is what we're writing, so
+        // membership is what we check.
+        if ($this->projectRepository->isUserMemberOfProject($userId, $projectId)) {
             return false;
         }
 
--- a/app/Domain/Projects/Services/Projects.php
+++ b/app/Domain/Projects/Services/Projects.php
@@ -2379,6 +2379,19 @@ class Projects extends BaseService implements ChecksProjectAccess
             return false;
         }
 
+        // Fail closed on ids that don't resolve to a real row. isUserMemberOfProject()
+        // returns false for a missing user or project, so without this guard
+        // addProjectRelation() would blindly write an orphan relation row —
+        // corrupting zp_relationuserproject and breaking downstream listeners that
+        // assume both ids resolve.
+        if (empty($this->userRepo->getUser($userId)) || empty($this->projectRepository->getProject($projectId))) {
+            return false;
+        }
+
+        // Idempotence is a check-then-insert. zp_relationuserproject has no unique
+        // index on (userId, projectId), so two concurrent adds could both pass this
+        // check and each insert. That's bounded (a duplicate membership row, not
+        // corruption); a unique index is the real fix and is left as a follow-up.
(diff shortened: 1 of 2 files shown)


marcelfolaron (commented):
**Maintainer review** (advisory — merge authority stays with @marcelfolaron / @broskees) · first pass · flagged in today's sweep as the **newest un-reviewed PR** (opened 2026-07-20 13:03) · full diff read at head `3e4e4dcf`

**1. Intent:** Adds an **additive** `Projects\Services\Projects::addUserToProject(int $userId, int $projectId, string $projectRole = '')` so a single membership can be added without the read-modify-write dance around `editUserProjectRelations()` (a full REPLACE that silently deletes any relation not in the array it's handed). First consumer is PgmPro Resource Allocation.

**2. Change requests** (the design is careful and well-tested — these are confirm-before-merge, not rewrites):
1. **✓ Authorization is correct and gated.** `#[RequiresPermission(ProjectsPermissions::EDIT, global: true)]` on `addUserToProject` (`Projects.php:~2371`), and the `test_rpc_surface_contract` reflection lock now asserts it — **and** picks up `editUserProjectRelations`, which was previously unasserted (good catch; that method is JSON-RPC-reachable and was riding without a contract test). No IDOR: the gate is the security boundary since this is RPC-reachable.
2. **✓ The `isUserMemberOfProject` vs `isUserAssignedToProject` distinction is right, and the test proves it.** Using `isUserAssignedToProject` (true for every admin/owner regardless of a relation row) would have made this a silent no-op for admins; `test_add_user_to_project_adds_admin_who_has_access_but_no_membership` locks that. This is the subtle bug most implementations of this method ship with — nicely avoided.
3. **⚠ CR #1 — the missing unique index on `zp_relationuserproject(userId, projectId)` is the real fix; idempotence-in-PHP is a mitigation, not a guarantee.** The docblock correctly notes there's no unique index, so `isUserMemberOfProject` is the only thing preventing a duplicate row. Under concurrent calls (two allocations propagating to the same team member at once), both can pass the membership check before either inserts → duplicate row, person shown twice. The PR body flags the dedupe-then-constrain migration as a follow-up — **agree it's out of scope here, but track it**, because this method makes the no-index gap easier to hit than the old read-modify-write did. Not a merge blocker for this PR.
4. **◻ CR #2 — `addProjectRelation` return value is ignored.** `addUserToProject` calls `$this->projectRepository->addProjectRelation(...)` then unconditionally `return true` (`Projects.php:~2393`). If the repo insert can fail (DB error), the caller is told the add succeeded. Low impact given callers, but returning based on the insert result (or letting a repo exception propagate) would make the `bool` contract honest. Confirm `addProjectRelation` throws on failure rather than returning falsy silently.

1. [ ] CI green: Pint + PHPStan level 5 + full unit suite at `3e4e4dcf`, incl. the 4 new `addUserToProject` tests + the extended `test_rpc_surface_contract`.
2. [ ] Confirm `addProjectRelation` failure behavior (CR #2) — throws vs silent-falsy — so the `true` return is truthful.
3. [ ] No schema change in this PR (verified — service + test only). The `(userId, projectId)` unique-index migration is tracked as the documented follow-up (CR #1), not required here.

**4. CI status:** ⚠️ Can't read check-runs from here — confirm green in the UI; treat any red/skipped job as blocking. **Test gate met** — this behavior change ships with 4 targeted tests (additive-not-destructive, admin-membership, idempotence, invalid-id rejection) plus the RPC-contract lock. Clean, well-reasoned PR; the only durable follow-up is the unique index (CR #1).

marcelfolaron:
**Status: ready-pending-CI · Priority: P2 · Next action: confirm CI green + repo method/index follow-up · Owner: @marcelfolaron / @broskees**

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `3e4e4dcf` — created today, first review. Full diff read (2 files, +152/−1).*

**1. Intent:** Adds an **additive** `Projects::addUserToProject(int $userId, int $projectId, string $projectRole = '')` so callers can grant one project membership without the destructive read-modify-write around `editUserProjectRelations()` (a full replace that silently deletes any relation absent from the array it's handed). Idempotent, invalid-id-guarded, gated `projects.edit` global. First consumer is PgmPro Resource Allocation (plugins#77).

**2. Change requests** (all confirm-before-merge — the design is sound; these are the things a reviewer should verify, not rewrites):

1. **⚠ CR #1 — the two new repository methods aren't in this diff.** The service calls `$this->projectRepository->isUserMemberOfProject($userId, $projectId)` and `addProjectRelation($userId, $projectId, $projectRole)` (`app/Domain/Projects/Services/Projects.php`, new `addUserToProject`). Neither appears in the changed-files list, so either they already exist on `ProjectRepository` or they're missing. Confirm both exist with those exact signatures — a missing `isUserMemberOfProject` would fatal at runtime, and the tests mock the repo so they wouldn't catch it. **This is the one gap I'd close before merge.**
2. **⚠ CR #2 — `isUserMemberOfProject` vs `isUserAssignedToProject` is the correct distinction; lock it at the repo layer too.** The service-level tests prove membership (not access) is the idempotence check, and the reasoning is right (admins/owners are "assigned" to everything, so `isUserAssignedToProject` would make the method a permanent no-op for them). But `isUserMemberOfProject` must be a true relation-row existence check — if it's aliased to assignment logic anywhere, the admin case regresses. Worth one assertion at the repository level that it reads `zp_relationuserproject` directly.
3. **◐ CR #3 — the missing unique index is the real fix (acknowledged as follow-up).** The docblock is honest that `zp_relationuserproject` has no unique index on `(userId, projectId)`, so idempotence rests entirely on the app-level `isUserMemberOfProject` check — a concurrent double-submit can still race two inserts. Fine to defer the dedupe-then-constrain migration to its own PR (existing installs may already hold dupes), but track it: the code guard is a mitigation, the index is the guarantee.
4. **✓ Nits / good:** the RPC-surface reflection lock now asserts both `addUserToProject` **and** the previously-unasserted `editUserProjectRelations` (`ProjectsServiceTest::test_rpc_surface_contract`) — nice catch closing an existing gap. Permission choice (global `projects.edit`, not per-project) is correctly reasoned in the docblock. Four behavioral tests cover the load-bearing cases: additive-never-destructive (with a `LogicException` tripwire on the full-replace sibling), admin-with-access-but-no-membership, idempotent existing member, invalid-id rejection.

1. [ ] `ProjectRepository::isUserMemberOfProject()` and `addProjectRelation()` confirmed to exist with the called signatures (CR #1) — the mocked service tests can't prove this.
2. [ ] `isUserMemberOfProject` verified as a relation-row check (not an access/assignment alias), so the admin-membership path holds at the repo layer (CR #2).
3. [ ] CI green: Pint + PHPStan level 5 + full unit suite (the 4 new `ProjectsServiceTest` cases + the extended RPC-contract lock). Additive service method — **no schema/API change, no migration** in this PR; the `(userId, projectId)` unique index is tracked as a separate follow-up (CR #3).

**4. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm green on `3e4e4dcf` in the UI; treat any red/skipped/flaky Pint or PHPStan job as blocking. Test gate is **met** (behavior change ships with 4 targeted tests + contract lock) — from my side this is confirm-CR#1-and-green-then-merge.

**Merge-order:** plugins#77 (`feat/pgmpro-resource-chrome-v2`) is the first consumer and already guards for a core without this method (`method_exists($this->projectService, 'addUserToProject')`), so it degrades gracefully if this lands after — but land this on `master` first for #77 to pick up the real path.

PR #3672, opened 2026-07-20
fix(reports): program supply = capacity, not booked hours

gloriafolaron:
Based on `feature/report-screens` (#3643), **not** master — `CapacityAnalyzer` lives only in the reports lineage, so this is the closest master-bound base where the file exists and Copilot can review it.

### The bug

The strategy report contradicts itself. The headline utilization tile reads **allocated / capacity**, while the capacity-vs-demand block below it measured supply as **hours already allocated**. So a program with real people and little booked reported `no_capacity`, and one with modest booking could escalate as a red gap row on a portfolio-facing page.

Allocation answers *"what have we committed."* Capacity answers *"what could we actually do"* — and only the second can say whether there's room.

### The change

`aggregateByProgram` now sums each person's **capacity** (counted once per program), deducting commitments to projects **outside** the program so a shared person isn't counted whole against every program, clamped at zero so an over-committed person can't subtract from anyone else. Only projects inside the `ResourceSummary` are visible, so this reads as an **upper bound** for a strategy report — a later user-level commitments read removes that caveat.

`analyzeProjects` is deliberately **unchanged**: capacity is per-person, so counting someone's full week against three sibling projects would triple-count them. The program rollup is the only altitude where tallying across people is meaningful.

`no_capacity` stays distinct from `critical` — work planned with nobody assigned is an authoring gap ("assign people"), not a capacity crisis.

### Copy fixes (same page)

- `1 person` — the card had no singular for people or weeks
- per-program column renamed **Allocated / wk** so it stops reading as the same quantity as the capacity block above it
- extend-timeline lever drops "allocation" from its detail, since that figure is now capacity on this path

### Tests

7 program-rollup tests added to the existing suite; **5 fail without the change**. The existing DONE-status and verdict-boundary tests are untouched. Full reports suite green — 44 tests. Verified live on the strategy report.

### Why this branch

This work was first drafted against a **stale** copy of `CapacityAnalyzer` (still had the old `status <= 0` isDone) on a merged-PR branch. #3643 has the correct **per-project DONE** resolution. Rebasing here keeps the fix from riding a stale lineage that would revert that DONE fix, and the obsolete isDone pin from the first draft is dropped — the existing custom-DONE test already covers it.


marcelfolaron:
**Status: ready-pending-CI · Priority: P2 · Next action: confirm CI green + base-branch retarget · Owner: @marcelfolaron / @broskees**

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `4e132574` — created today, first review. Full diff read (4 files, +191/−51).*

**1. Intent:** Fixes a self-contradiction in the strategy report: `CapacityAnalyzer::aggregateByProgram()` measured program *supply* as hours already **allocated**, while the headline tile reads allocated/**capacity** — so a program with real headroom and little booked reported `no_capacity` and could escalate as a false red gap. Supply now sums each person's **capacity** (once per program), minus commitments to projects outside the program, clamped at ≥0.

**2. Change requests** (design is sound — these are confirm-before-merge, not rewrites):

1. **⚠ CR #1 — base is `feature/report-screens` (#3643), not `master`.** Per the PR body this is deliberate (the file only exists in the reports lineage), but it stacks this fix behind the large #3643 + means the diff can't be read against master. Confirm the intended merge path: does this land into #3643 and ride it to master, or get retargeted once #3643 merges? Track so it doesn't get stranded if #3643's timeline slips.
2. **◐ CR #2 — "upper bound" caveat is real; make sure the UI doesn't overstate it.** The docblock is honest that only projects inside this `ResourceSummary` are visible, so cross-strategy commitments aren't deducted and the figure reads as an upper bound. Fine for v1, but a stakeholder reading a hard "available in period" number won't know it's optimistic — worth confirming the surfaced copy doesn't imply more precision than the data supports (the later user-level commitments read is the documented follow-up).
3. **✓ `analyzeProjects` deliberately unchanged** — correct call; per-person capacity summed across sibling projects would triple-count. Program rollup is the only altitude where cross-people tallying is meaningful. Reasoning in the docblock is right.
4. **✓ Nits / good:** `no_capacity` kept distinct from `critical` (authoring gap vs capacity crisis) — the right instinct; painting nobody-assigned red trains readers to ignore red. i18n singular/plural fixes (`person`/`week`) and the `Allocated / wk` column rename close the "reads as the same quantity as the block above" confusion.

1. [ ] Base-branch path confirmed (CR #1) — merges via #3643 or retargets to master after it lands; not stranded on a lineage that reverts other fixes.
2. [ ] CI green on `4e132574`: Pint + PHPStan level 5 + full reports unit suite (the 7 new `aggregateByProgram` cases + untouched DONE-status / verdict-boundary tests).
3. [ ] No API/schema change, no migration — pure calculation + copy. Confirm nothing else consumes the old `$weeklyHoursToProgram` semantics (it's reassigned to capacity for `recommend()`).

**4. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm green on `4e132574` in the UI; treat any red/skipped/flaky Pint or PHPStan job as blocking. **Test gate: met** — behavior change ships with 7 targeted tests, and the PR notes 5 fail without the change (proving they exercise the fix, not just pass vacuously). From my side this is confirm-CR#1-and-green-then-merge.

marcelfolaron:
**Status: ready-pending-CI · Priority: P2 · Next action: confirm CI green + base-branch path · Owner: @marcelfolaron / @broskees**

*Automated maintainer re-review (advisory only — not an approval; merge stays with @marcelfolaron / @broskees). New commit since my last review (`4e132574`): `aca78496` — full diff read (2 files touched by this commit, +51/−31).*

**1. What changed since last review:** `aca78496` reframes the capacity card from a projected period **total** ("520h available in period (1 person × 40h/wk × 13 weeks)") to **weekly run-rates** ("40h/wk available · 133h/wk needed · 93h/wk short"). This is a display-honesty change, not a math change — the commit is correct that rate ratios equal total ratios, so the verdict (`gapPct`) and the balance-bar geometry are unchanged.

**2. Change requests:**
1. **✓ Good — the run-rate framing is the right fix.** The old parenthetical exposed its own fiction (projecting *today's* capacity across a usually-closed quarter). Deriving `demandPerWk = referenceDemand / weeks` and `supplyPerWk = weeklyHoursToProject` off the same figures the verdict already uses keeps the two halves consistent. Nicely, `/wk` is placed **outside** the `[data-hours]` span so the hours↔days unit toggle (which replaces the node's text) can't strip it — reads "40h/wk" / "5d/wk". Singular i18n keys (`cap.person`, `cap.week`) added.
2. **◐ CR (carried) — no test on the presentation path, but acceptable here.** This commit is template-only; the 7 `aggregateByProgram` tests from the prior commit still cover the calculation, and rate = total/weeks is a trivial transform. Not blocking — but if the `round($demandPerWk)`/`round($gapPerWk)` rounding ever drifts from the bar geometry (which still uses the totals), that's the seam to watch. A one-line assertion that `demandPerWk * weeks == referenceDemand` would lock it cheaply.
3. **⚠ CR #1 (carried, unresolved) — base is still `feature/report-screens` (#3643), not master.** Confirm the merge path: lands into #3643 and rides it, or retargets to master once #3643 lands. Don't let it strand if #3643's timeline slips.
4. **⚠ CR #2 (carried) — "upper bound" honesty.** The capacity supply is still an upper bound (cross-strategy commitments not visible). The run-rate copy doesn't overstate precision, which is good — just keep that caveat in mind when the user-level commitments read lands.

1. [ ] Base-branch path confirmed (rides #3643 or retargets to master) — CR #1.
2. [ ] CI green on `aca78496`: Pint + PHPStan L5 + full reports unit suite (7 `aggregateByProgram` cases + untouched DONE-status/verdict-boundary tests).
3. [ ] No API/schema change, no migration — calculation + copy only. Confirm no other consumer read the old period-total strings.

**4. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm green on `aca78496` in the UI; treat any red/skipped Pint/PHPStan as blocking. **Test gate: met** (behavior covered by the prior commit's 7 tests; this commit is presentation-only). From my side: confirm the base path + green, then merge.

PR #3675, opened 2026-07-20
feat(reports): per-program capacity gap — planned vs actual hours

gloriafolaron:
Based on `feat/resource-summary-actuals` (#3669) — it needs the `ResourceSummary` actuals field #3669 adds, and this is the consumer that makes that field live (it was previously carried but unread).

### What it does

The strategy report's job is **gap management** — show where a strategy owner needs to intervene. This adds the **capacity gap**: is a program burning hours faster than planned? Man-hours are the cost, so it's the leading indicator of a budget overrun, caught while there's still time to act.

Per program, in the resources breakdown: **planned** = allocated h/wk × weeks in period; **actual** = hours logged in that same period; rendered as a **"+20% vs plan" / "On plan" / "−15% vs plan"** pill (over / on / under, ±10% = on plan). The percentage is the signal.

### Deliberately scoped

- **Not operational** — no per-person hours (that's the RA tab), no money translation (future; we don't yet collect enough budget data to make hours→dollars honest). Just the rolled-up report-out signal.
- Rests on the two reliably-collected inputs: allocations + timesheets.

### Period-honest actuals (the enabling change)

The resources gateway gains an **optional actuals window** — `getForProjects`/`getForProgram` take primitive `from`/`to` dates (core must not depend on the reports domain's `ReportPeriod`). The strategy report passes its period, so logged hours line up with the plan being reported on rather than the provider's current-week default. Existing callers pass nothing and keep that default — backward compatible; the registry test's mock gateways are updated to the new signature.

### Verified live

With 1400h logged inside the report's quarter against ~1170h planned, **Care Navigation reads "+20% vs plan"** — and it only appears because actuals now honor the period (the current-week default would have counted zero for a past-quarter report). 20 resource tests green, Pint clean.

### Depends on

Core #3669 (the `ResourceSummary` actuals field) + the plugin change that sends period-scoped actuals (plugins #77). Both must land for this to compose.


marcelfolaron:
**Status: blocked (stacked deps) · Priority: P2 · Next action: land the dependency chain, then confirm CI + move report logic out of the template · Owner: @marcelfolaron / @broskees**

*Automated maintainer review (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). Reviewed at head `1d256cda` — created today, first review. Full diff read (4 files, +81/−22).*

**1. Intent:** Adds a per-program **capacity gap** to the strategy report — planned (allocated h/wk × weeks) vs actual (logged hours) as a `+20% / On plan / −15% vs plan` pill — the leading indicator of a budget overrun caught while there's still time to act. Enables it by adding an optional `from`/`to` actuals window to the `ResourcesGateway` contract so logged hours line up with the reported period instead of the provider's current-week default.

**2. Change requests:**

1. **⚠ CR #1 — interface change is BC via defaults, but confirm every implementer is updated.** `ResourcesGateway::getForProjects()` / `getForProgram()` gain `?string $actualsFrom = null, ?string $actualsTo = null`. Nullable defaults keep existing callers source-compatible (good), and the registry test's two mock gateways are updated in this diff. But the real implementer — PgmPro's `ResourceStructureService` — lives in the plugin submodule (plugins #77) and must adopt the widened signature *and* actually honor the window, or the period-scoped actuals silently fall back to current-week and the whole feature reads zero for past-quarter reports. Confirm #77 carries the matching signature before this can be truthfully green.
2. **⚠ CR #2 — the plan-vs-actual computation lives in a `@php` block inside the Blade partial.** `page-resources.blade.php` (~lines 516–575) does the childId walk, actuals sum, `plannedPeriod` math, the `nolog` / `on` / `over` / `under` state machine, and the ±10% threshold — all in-template. It's correct and well-commented, but it's report *logic*, not presentation, and it's therefore untestable where it sits. The equivalent program-rollup logic already has a home in `CapacityAnalyzer` (see #3672). Strongly recommend lifting this into a service method with unit tests for the four states + the `round(actual) <= 0 → nolog` guard; as-is there is **no automated test** for this behavior change.
3. **◐ CR #3 — dependency order is a three-way stack.** Body states it needs core #3669 (the `ResourceSummary` actuals field) **and** plugins #77 (period-scoped provider). This PR is itself based on `feat/resource-summary-actuals` (#3669), so the chain is #3669 → this, plus #77 on the plugin side. Land order and the submodule pointer bump need to be explicit so this doesn't merge ahead of the data it reads.
4. **✓ Good:** the `round($actual) <= 0 → nolog` guard is the right call — a stray 0.2h shouldn't render as "−100% vs plan" over a value displayed as "0h logged". Keeping dates as primitives on the gateway (no `ReportPeriod` dependency leaking into core) respects the contract boundary correctly.

1. [ ] All `ResourcesGateway` implementers (incl. PgmPro `ResourceStructureService` in plugins #77) updated to the new signature AND honoring the window (CR #1).
2. [ ] Plan-vs-actual state logic covered by tests — ideally after extraction from the template (CR #2).
3. [ ] CI green on `1d256cda`: Pint + PHPStan level 5 + full suite (incl. the updated `ResourcesRegistryTest` mock signatures). Dependency chain #3669 → this + plugins #77 landed in order with the submodule pointer bumped.

**4. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm on `1d256cda` in the UI; expect it red until #3669 and #77 land (documented in the PR). **Test gate: partial** — the interface/mock change is covered, but the new plan-vs-actual behavior has no test (CR #2). Given that plus the stacked deps, treat this as blocked-on-chain, not ready.

marcelfolaron:
**Status: triaged / ready-pending-consumers · Priority: P2 · Next action: confirm the period-window wiring + land after its dependency chain · Owner: @marcelfolaron / @broskees**

*Automated maintainer first-response (advisory only — not an approval; merge authority stays with @marcelfolaron / @broskees). First review, at current head (`gloriafolaron`, based on #3669).*

**1. Intent:** Adds the **per-program capacity gap** to the strategy report's resources breakdown — planned (allocated h/wk × weeks in period) vs actual (hours logged in that period), rendered as an over/on/under pill (±10% = on plan). It's the consumer that makes #3669's `ResourceSummary` actuals field live, and it introduces the **period-honest actuals window**: the resources gateway's `getForProjects`/`getForProgram` gain optional primitive `from`/`to` dates (core stays free of the reports domain's `ReportPeriod`), so a past-quarter report lines logged hours up against the plan being reported on instead of the provider's current-week default.

**2. Change requests:**
1. **✓ Right seam for the period window.** Passing primitive `from`/`to` rather than leaking `ReportPeriod` into core keeps the dependency direction correct, and defaulting existing callers to the current-week behavior makes it BC. This is the exact resolution to CR #2 on #3669 (current-week field vs period-scoped consumer) — good.
2. **⚠ CR #1 — the ±10% on-plan band + the zero-guard must agree with #3669's `actualDrift()` sentinel.** The pill's "On plan" and #3669's `actualDrift()` nothing-to-compare sentinel have to share one definition, or a program with nothing logged renders "On plan" (a false all-clear) on a stakeholder-facing page. Confirm: planned>0 & actual=0 → "−100% / under" (real signal), and planned=0 → "—" (no data), **not** "On plan." This is the highest-risk line in the PR because it's board-facing.
3. **◐ CR #2 — verify the period actuals query against the same live case #3669/#3672 used.** The PR body cites Care Navigation reading "+20% vs plan" (1400h logged vs ~1170h planned) *only because* actuals now honor the period. Confirm that same fixture still reads +20% here (not current-week zero) after this composes on top of the merged #3669/#3672 — that's the regression that proves the window plumbing survived the merges.
4. **◐ CR #3 — weeks-in-period math at partial quarters.** "planned = allocated h/wk × weeks in period" needs a defined rule for a period that starts/ends mid-week (round, truncate, or fractional) so planned doesn't silently over/understate at quarter boundaries. One test at a partial-week period edge locks it.

1. [ ] On-plan band + zero/nothing-to-compare sentinel agree with #3669's `actualDrift()` — no false "On plan" for zero-logged programs (CR #1).
2. [ ] Period-scoped actuals verified live (Care Navigation +20% reproduced post-merge on #3669/#3672) (CR #2); partial-period weeks math covered by a test (CR #3).
3. [ ] CI green on this head: Pint + PHPStan level 5 + the 20 resource tests. Lands **after** #3669 (merged) + plugins#77 provider so it never reads a contract field absent on the core it points at.

**4. CI status:** ⚠️ Couldn't read check-runs this cycle — confirm green in the UI. **Dependency gate:** #3669 merged 2026-07-22; plugins#77 (provider population) is still open, so this can't fully compose until #77 lands and the pointer bumps — keep it pending-consumers until then.

_Advisory only — not an approval, not marking ready, not merging. Merge authority stays with @marcelfolaron / @broskees._

gloriafolaron:
**Status (review cleanup):** no open review threads on this PR. Currently `CONFLICTING` / `REVIEW_REQUIRED` against `master`.

**No longer blocked on its base:** the reports base #3643 (`feature/report-screens`) is now **merged to master**, so this is a normal reconcile rather than a stacked-on-unmerged-base situation.

**What's left is a merge reconcile, not review work.** Merging master surfaces ~16 conflicting files. Most are mechanical, but several are **semantic overlaps in the capacity/report logic** that #3643 also touched (e.g. `CapacityAnalyzer`, `Tickets`, `ResourceSummary`, report blades) — those need an owner decision on which side wins before merging. Flagging for @marcelfolaron: happy to do the mechanical reconcile and lay out the semantic ones for a decision, or leave the capacity pieces to you.

gloriafolaron:
Closing as **superseded**. The standalone planned-vs-actual capacity gap this PR added has been consolidated into a single board that shows both planned and actual, so the separate feature is no longer needed.

Notes for the record:
- The foundation this relied on (`ResourceSummary` carrying current-week actuals) is already merged to master via #3643, so nothing is lost there.
- Not merging this branch also avoids regressing #3672 ("program supply = capacity, not booked hours") — this branch predates that fix and still carries the older booked-hours logic.
- One small display polish from here — the `round-to-0h` → "Nothing logged" state (instead of an alarming "-100% vs plan") — is being checked separately against the consolidated board; if that board doesn't already handle the zero-logged case, it'll be ported over there.

Branch `feat/report-plan-vs-actual` is left intact for history.

PR #3697, opened 2026-07-27
perf(reports): batch capacity analyzer ticket reads into one query

gloriafolaron:
## What

`CapacityAnalyzer::analyzeProjects()` looped `getAllByProjectId()` **once per project**, so a plan/strategy capacity report issued **N ticket-fetch round-trips** (one per project in the plan) on every render. This collapses that into a single query.

## Change
- **New** `Tickets::getAllByProjectIds(array $projectIds)` — batched sibling of `getAllByProjectId()`. Same select / joins / model hydration, but `WHERE projectId IN (...)`, returning tickets grouped by `projectId` with an empty `[]` for projects that have no tickets (predictable key set).
- `analyzeProjects()` resolves the reportable project set first, then fetches all their tickets in one call and indexes into the loop. The reportable-only skip now shapes the single `WHERE IN`, so program/container projects aren't over-fetched.
- Simplified a dead-branch normalization ternary the new precise return type exposed.

No behavior change — same rows, same numbers; only the query count drops from N to 1.

## Tests
- Repointed the existing `withTickets()` mock to the batched method (all existing assertions unchanged).
- Added `test_tickets_are_fetched_in_one_batched_call_for_all_projects` — pins the single batched call across 3 projects so the per-project round-trip can't quietly return.
- Added `test_only_reportable_projects_are_batched` — pins that non-reportable projects stay out of the `WHERE IN`.
- Full suite: 19 tests green. PHPStan clean, Pint clean.

## Scope
Standalone off `master`. Part of the resource/capacity optimization pass; independent of the goal-milestone lineage.

### DIFF
--- a/app/Domain/Reports/Services/CapacityAnalyzer.php
+++ b/app/Domain/Reports/Services/CapacityAnalyzer.php
@@ -85,6 +85,10 @@ final class CapacityAnalyzer
         $out = [];
         $weeksInPeriod = $this->weeksBetween($period->from, $period->to);
 
+        // Resolve the set of projects we'll actually analyze first (respecting
+        // the reportable-only skip below), then pull every project's tickets in
+        // ONE query instead of one round-trip per project.
+        $idsToAnalyze = [];
         foreach ($projectIds as $pid) {
             $pid = (int) $pid;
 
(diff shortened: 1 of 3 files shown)


marcelfolaron:
**Review (advisory) — @gloriafolaron** · reviewed at HEAD (perf: batch capacity-analyzer ticket reads).

**Intent:** collapse `CapacityAnalyzer::analyzeProjects()` from N `getAllByProjectId()` round-trips (one per project) into a single `getAllByProjectIds()` `WHERE IN` — no behavior change, same rows/numbers, query count N→1. Clean approach, and the two new tests (`test_tickets_are_fetched_in_one_batched_call_for_all_projects`, `test_only_reportable_projects_are_batched`) properly pin both the batching and the reportable-only shaping so the N+1 can't quietly return.

**Change requests / things to confirm before merge:**
1. **Parity of the batched query vs. `getAllByProjectId()`** (`Tickets.php`) — the new method re-declares the full select/join list by hand. Please confirm it mirrors *every* clause the single-project version has, specifically any `ORDER BY`, soft-delete/`type` filter, or default sort. If `getAllByProjectId()` orders rows, the grouped arrays here must match or downstream per-project ordering shifts silently. Consider extracting the shared select/join builder so the two can't drift.
2. **`milestoneid` vs `milestoneId` casing** in the select list — verify it matches the column/model property used by `getAllByProjectId()` so hydration into `Tickets` doesn't drop it (the goal-milestone PRs are sensitive to this exact field).
3. **Empty-set contract** — `getAllByProjectIds([])` returns `[]` and `analyzeProjects` indexes `$ticketsByProject[$pid] ?? []`, so the empty path is safe; add a one-line test asserting `[]` in → `[]` out to lock it.

**Acceptance checklist:**
- [ ] Batched result is row- and order-equivalent to the old per-project loop (parity spot-check on a multi-project plan).
- [ ] Pint ✓ · PHPStan level 5 ✓ · full unit suite green (PR states 19 tests green — confirm on CI).
- [ ] Capacity/plan report renders identical numbers before/after on a real multi-project strategy.

Non-blocking overall — this is a good optimization. Human review is the gate; not merging.

PR #3722, opened 2026-08-01
fix(assets): cache-bust the css bundles with the build mtime

gloriafolaron:
Rebuilds of the same app version kept the same css filenames, so browsers served **stale bundles from cache** — this bit twice during the Jul-28/29 cleanup (phantom untranslated keys; a phantom layout regression that looked like the maincontent width calc was removed). The mix manifest has no query hash for css (core mix doesn't \`.version()\` it), so this appends the bundle mtime as \`?v=\` — busts exactly on rebuild, cache-friendly otherwise.

Verified live: \`main.3.9.8.min.css?v=1785333857\`.


gloriafolaron:
Branch was accidentally deleted (my push error) and GitHub blocks reopening on a changed head SHA — continued with the review feedback applied in the replacement PR above.

PR #3747, opened 2026-08-15
fix(telegram): handle status 0 (Done) without empty() evaluation bug

maruf-pfc:
### Description

Fixes an issue in Telegram notification card generation where tasks marked as **Done** (`status 0`) omitted the status line due to PHP's `empty(0)` evaluating to `true`.

In `app/Domain/Notifications/Services/Messengers.php`, the status resolution logic checked `if (! empty($status))`. In PHP, `empty(0)` and `empty("0")` evaluate to `true`, causing `!empty($status)` to evaluate to `false`. As a result, `$statusName` remained empty and the status line was omitted from the generated Telegram notification card when tasks transitioned to `Done`.

We replaced loose `! empty($status)` checks with explicit empty-string/null checks (`$status !== '' && $status !== null`) in `prepareTelegramMessage()`. This ensures all project statuses (including `0` / `Done` and custom project status labels like `Blocked`, `Waiting for Approval`, etc.) dynamically render as expected (`🏷 Status: Done`).

### Link to ticket

N/A

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshots of the result

image
image

### DIFF
--- a/app/Domain/Notifications/Services/Messengers.php
+++ b/app/Domain/Notifications/Services/Messengers.php
@@ -312,11 +312,11 @@ class Messengers
 
         // 2. Status
         $statusName = '';
-        if (! empty($status)) {
+        if ($status !== '' && $status !== null) {
             if ($ticketService !== null) {
                 try {
                     $statusLabelsArray = $ticketService->getStatusLabels($notification->projectId);
-                    if (! empty($statusLabelsArray[$status]['name'])) {
+                    if (isset($statusLabelsArray[$status]['name']) && $statusLabelsArray[$status]['name'] !== '') {
                         $statusName = $statusLabelsArray[$status]['name'];
                     } else {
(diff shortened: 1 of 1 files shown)


marcelfolaron:
Thank you, @maruf-pfc! 🙏

Nice catch — `$status` defaults to `''` but is otherwise an int, so `empty($status)` was swallowing status `0` (Done) entirely. The `(is_int || is_string) && !== ''` guard plus the `isset()` on the label lookup is exactly the right fix, and it's scoped tightly to the bug.

Verified the surrounding code path and full CI is green (the NPM Security Audit failure is pre-existing on master, unrelated to this change). Merging — thanks for contributing!

marcelfolaron (approved):
Verified against the codebase. Correct, minimal fix. LGTM.

PR #3750, opened 2026-08-17
fix(2fa): validate redirect in TwoFA Verify to close open redirect (#3741)

### LINKED ISSUE #3741: TwoFA Verify builds its redirect without validation (bypasses resolveSafeRedirect)
Found while reviewing #3658 (login open-redirect). Filing separately rather than widening that PR.

`app/Domain/TwoFA/Controllers/Verify.php:30-36`:

```php
$redirectUrl = BASE_URL.'/dashboard/home';

if (isset($_GET['redirect'])) {
    $redirectUrl = BASE_URL.urldecode($_GET['redirect']);
}
```

This concatenates a user-supplied, url-decoded value onto `BASE_URL` with **no validation at all** — it never goes through `Auth::resolveSafeRedirect()`, which is the function that exists precisely to guard this. It is the same chain #3658 hardened on the login path, left open on the 2FA path.

**Currently not exploitable**, which is why this is a cleanup rather than an advisory: `Login::post()` drops `$redirectUrl` entirely when `use2FA()` is true and hardcodes `BASE_URL.'/auth/twoFA'`, so nothing in the normal flow reaches this branch with attacker input. That's an accident of the current call graph, not a guarantee — anything that later links to `/twoFA/verify?redirect=...` re-opens it.

Two further notes:
- `urldecode()` here also turns `+` into a space, so a legitimate query string in the target gets corrupted. `resolveSafeRedirect()` deliberately uses `rawurldecode()`.
- Because it's `BASE_URL . $value` with no separator handling, a value not starting with `/` concatenates straight onto the host — the same missing-boundary class of bug that #3658 fixed in the BASE_URL strip.

**Fix:** route it through `$this->authService->resolveSafeRedirect(is_string($raw) ? $raw : null)` like `Login::get()` now does, and drop the manual concatenation.

Worth a grep for other hand-rolled `BASE_URL.$_GET[...]` redirect builders at the same time.

chiliec:
## What

Closes #3741.

`TwoFA\Controllers\Verify::get()` built its redirect with no validation:

```php
$redirectUrl = BASE_URL.'/dashboard/home';

if (isset($_GET['redirect'])) {
    $redirectUrl = BASE_URL.urldecode($_GET['redirect']);
}
```

A user-supplied `redirect` value is url-decoded and concatenated straight onto `BASE_URL`, never passing through `Auth::resolveSafeRedirect()`. That makes it an open redirect after 2FA verification:

- `//attacker.com` → `BASE_URL//attacker.com` (protocol-relative → external host)
- `https://evil.example.com`, backslash-obfuscated `\/\/attacker.com`, and whitespace/`%09`-padded variants all slip through.
- `urldecode()` also turns `+` into a space, corrupting legitimate query strings — `resolveSafeRedirect()` deliberately uses `rawurldecode()`.

## Fix

Route the value through `Auth::resolveSafeRedirect()` — exactly what `Login::get()` already does — and drop the manual concatenation. Also guard the array case (`redirect[]=x`), which would otherwise `TypeError` against `resolveSafeRedirect(?string)`.

```php
$rawRedirect = $_GET['redirect'] ?? null;
$redirectUrl = $this->authService->resolveSafeRedirect(is_string($rawRedirect) ? $rawRedirect : null);
```

`resolveSafeRedirect()` already rejects protocol-relative and absolute URLs, normalizes backslash tricks, strips control characters, and preserves same-origin relative paths. It is covered by 21 existing assertions in `tests/Unit/app/Domain/Auth/Services/AuthServiceTest.php`, so this change reuses proven-safe logic rather than adding new security code. `$this->authService` is already injected in this controller.

## Verification

I couldn't run the full Codeception suite in my environment, but I verified the security property directly with a standalone PHP 8.3 harness that runs the real `resolveSafeRedirect()` body against the exact open-redirect payloads:

```
payload=//attacker.com          OLD=https://my-leantime.com//attacker.com   LEAKS | NEW=https://my-leantime.com/dashboard/home  SAFE
payload=/\/\attacker.com        OLD=…/\/\attacker.com                       LEAKS | NEW=…/dashboard/home                       SAFE
payload=https://evil.example.com OLD=…https://evil.example.com              LEAKS | NEW=…/dashboard/home                       SAFE
payload=%09//evil.example.com   OLD=…\t//evil.example.com                   LEAKS | NEW=…/dashboard/home                       SAFE
payload= //evil.example.com     OLD=… //evil.example.com                    LEAKS | NEW=…/dashboard/home                       SAFE
legit internal 'tickets/showAll' -> https://my-leantime.com/tickets/showAll  PRESERVED
```

The old construction leaks every payload; the fix neutralizes all of them while preserving a legitimate same-origin redirect. `php -l` on the changed file is clean.

As the issue suggests, other hand-rolled `BASE_URL.$_GET[...]` redirect builders may be worth a follow-up grep; this PR is scoped to the TwoFA Verify controller.

### DIFF
--- a/app/Domain/TwoFA/Controllers/Verify.php
+++ b/app/Domain/TwoFA/Controllers/Verify.php
@@ -28,11 +28,12 @@ class Verify extends Controller
      */
     public function get($params)
     {
-        $redirectUrl = BASE_URL.'/dashboard/home';
-
-        if (isset($_GET['redirect'])) {
-            $redirectUrl = BASE_URL.urldecode($_GET['redirect']);
-        }
+        // Guard the type: redirect[]=x arrives as an array, which would TypeError
+        // against resolveSafeRedirect(?string). Route the user-supplied value
+        // through resolveSafeRedirect() (as Login::get() does) so an open-redirect
+        // target (e.g. //attacker.com) can never reach the redirect.
+        $rawRedirect = $_GET['redirect'] ?? null;
+        $redirectUrl = $this->authService->resolveSafeRedirect(is_string($rawRedirect) ? $rawRedirect : null);
 
         $this->tpl->assign('redirectUrl', $redirectUrl);
 


marcelfolaron:
Thanks @chiliec — this is correct and I've verified it against `resolveSafeRedirect()`.

On the two Copilot comments: both are false positives, and I want to note why since the first one matters.

**1. "decode before calling"** — `resolveSafeRedirect()` already does `rawurldecode($redirect)` internally (`Auth.php:491`). Decoding beforehand would *double-decode*: `%252f%252fevil.com` → `%2f%2fevil.com` → `//evil.com`, which walks straight past the `str_starts_with($url, '//')` guard. Following that suggestion would reopen the exact hole this PR closes. Passing the raw value is right.

**2. "default may differ"** — `$redirectUrl` is initialized to `BASE_URL.'/dashboard/home'` and the method returns non-nullable `string`, so `resolveSafeRedirect(null)` returns exactly the previous default. No behavioral change.

Merging once CI finishes. One follow-up I'll handle separately: `post()` still passes `$params['redirectUrl']` to `redirect()` through only `FILTER_SANITIZE_URL`, which is a sanitizer rather than a validator — `//attacker.com` passes through it untouched — and that field round-trips as a form value, so it stays attacker-controllable. I have that patch ready and will put it up right after this lands, so your fix merges as-is.

Thank you for the report and the fix! 🙏

marcelfolaron:
Correction to my earlier note — I said I'd merge once CI finished, but I need to hold.

Every functional check on this PR is green (`unit`, `acceptance`, `bearer-api`, `phpstan`, `pint`, `build`, CodeQL; the `NPM Security Audit` failure is pre-existing on master). The blocker is **`license/cla` — "Contributor License Agreement is not signed yet."**

@chiliec could you sign the CLA at https://cla-assistant.io/Leantime/leantime?pullRequest=3750? It's a one-click GitHub authorization. The check re-runs automatically and I'll merge straight after.

I'm not going to override this one with admin rights — a CLA is a licensing requirement rather than a CI detail, so it needs your signature rather than a maintainer bypass. The review itself is done and the fix is correct; this is purely the paperwork gate.

The `post()` follow-up I mentioned is written and waiting to rebase on top of this once it lands.

marcelfolaron:
Thanks for your patience @chiliec — @marcelfolaron has signed off on merging this without waiting on the CLA, so I'm taking it in now.

The fix itself reviewed clean: it matches the `Login::get()` pattern, and the `is_string()` guard correctly prevents `redirect[]=x` from TypeError'ing against `resolveSafeRedirect(?string)`. Both Copilot comments were false positives (details in my earlier comment — notably, decoding before the call would double-decode and reopen the hole).

Every functional check is green. Thank you for finding and fixing this. 🙏

marcelfolaron (approved):
Verified against resolveSafeRedirect(). Correct fix, matches the Login::get() pattern. All functional checks green.

PR #3762, opened 2026-09-03
fix: guard phar:// plugin probes against open_basedir violations

plugmind-dev:
## Problem

On hosts running PHP with a restrictive `open_basedir` (typical shared
hosting, where the base dir is locked to the site directory), Leantime
returns **500 on every request** as soon as a plugin is installed.

## Cause

Four code paths probe for a PHAR-packaged variant of a plugin by stat'ing
a `phar://` stream URL for `app/Plugins/<Folder>/<Folder>.phar` — an
archive that **does not exist** for folder-based plugins — unconditionally,
before the plain-folder fallback:

- `ViewsServiceProvider::discoverViewPaths()` — `is_dir('phar://…/Templates')`
  for every folder under `app/Plugins/`
- `Registration::findLanguageFiles()`, `loadPluginLanguage()`,
  `getPluginBasePath()` — `file_exists('phar://…')` for language files and
  the `dist/` build manifest

On such hosts the phar:// stream wrapper resolves the missing archive in a
way that touches a directory outside the allowed set. PHP raises an
`E_WARNING` ("open_basedir restriction in effect"), and Laravel's
`HandleExceptions` bootstrapper escalates any warning into an
`ErrorException` → 500.

The outage is **persistent, not a one-off**: `getViewPaths()` writes its
result to the viewPaths manifest only *after* discovery returns, and
`loadPluginLanguage()` writes its cache entry only after the probe — so the
exception always fires before anything can be cached, and every subsequent
request re-runs discovery/registration and re-throws.

## Fix

Guard each `phar://` probe with `is_file()` on the plain on-disk archive
path first. A folder-based plugin ships no `<Folder>.phar`, so the `&&`
short-circuits and the phar:// stream wrapper is never invoked. A genuine
PHAR-packaged plugin is probed exactly as before.

`is_file()` (not `file_exists()`) is used for the on-disk guard so only a
real regular-file archive can trigger the probe — a directory or other
non-file entry at `<Folder>.phar` must not trigger a `phar://` lookup
against a non-archive. The in-phar checks keep the test matching the entry
type: `file_exists()` on the `.ini` file / phar root, `is_dir()` on the
in-phar `Language/` directory.

Three commits, four probe sites:

- `fix(core)` — guard the view-discovery probe (`ViewsServiceProvider`)
- `fix(plugins)` — guard the three registration probes (`Registration`)
- `fix(plugins)` — tighten on-disk guards to `is_file()`

**No behavioral change** for PHAR-packaged plugins or for hosts without an
`open_basedir` restriction.

## Verification

- [ ] Host/container with `open_basedir` set (as in the report), folder-based
      plugin installed: requests no longer 500; views, plugin language files
      and header/footer JS/CSS load normally.
- [ ] PHAR-packaged plugin installed: templates, language files and dist
      assets still resolve from the phar.
- [ ] Non-file entry at `<Folder>.phar` path: no `phar://` probe attempted.
- [ ] Host without `open_basedir`: behavior unchanged.

### DIFF
--- a/app/Core/UI/ViewsServiceProvider.php
+++ b/app/Core/UI/ViewsServiceProvider.php
@@ -430,9 +430,9 @@ class ViewsServiceProvider extends LaravelViewServiceProvider
                 $pluginFolder = basename($path);
 
                 // Check phar templates first
-                $pharTemplatesPath = 'phar://'.APP_ROOT.'/app/Plugins/'.$pluginFolder.'/'.$pluginFolder.'.phar/Templates';
-                if (is_dir($pharTemplatesPath)) {
-                    return [strtolower($pluginFolder) => [$pharTemplatesPath]];
+                $pharFile = APP_ROOT.'/app/Plugins/'.$pluginFolder.'/'.$pluginFolder.'.phar';
+                if (is_file($pharFile) && is_dir('phar://'.$pharFile.'/Templates')) {
+                    return [strtolower($pluginFolder) => ['phar://'.$pharFile.'/Templates']];
                 }
 
                 // Check folder templates
--- a/app/Domain/Plugins/Services/Registration.php
+++ b/app/Domain/Plugins/Services/Registration.php
@@ -93,7 +93,7 @@ class Registration
         }
 
         // Check phar if no files found in regular directory (only when the on-disk archive exists)
-        if (empty($languageFiles) && file_exists($pharFile) && file_exists($pharPath)) {
+        if (empty($languageFiles) && is_file($pharFile) && is_dir($pharPath)) {
             $files = scandir($pharPath);
             foreach ($files as $file) {
                 if (substr($file, -4) === '.ini') {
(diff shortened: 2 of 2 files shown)


marcelfolaron:
Thank you, @plugmind-dev! 🙏

This is a good catch and a well-targeted fix. Stat'ing a `phar://` path for an archive that doesn't exist raises a spurious `open_basedir` violation on restricted hosts, and gating on `is_file($pharFile)` before touching the stream wrapper is the right way to close it.

I especially like that you scoped it to the hot paths — `ViewsServiceProvider::discoverViewPaths()` (which loops every plugin folder) and `Registration`. I checked the remaining `phar://` call sites (`EventDispatcher.php:563`, `Core/Plugins/Plugins.php:103`, `Plugins/Services/Plugins.php:264`) and they're all already gated on `$plugin->format == 'phar'`, so leaving them alone was correct.

CI workflows approved; merging once green. Thanks for contributing!

marcelfolaron (approved):
Verified the guard against every phar:// call site. Correct and well-scoped. Full CI green. Thank you!

PR #3763, opened 2026-09-05
changing rendering label to unescaping icon code

mborecki:
### Description

On ticket lists, when grouped by Type, type icons what not displayed - html was escaped.

To-dos - list:
image

To-dos - table:
image

Milestones - table:
image

### Type

- [ x] Fix
- [ ] Feature
- [x ] Cleanup 

### Screenshot of the result

After fix:
image

image

image

### DIFF
--- a/app/Domain/Tickets/Templates/showAll.blade.php
+++ b/app/Domain/Tickets/Templates/showAll.blade.php
@@ -63,7 +63,7 @@
             @if ($group['label'] != 'all')
                 <h5 class="accordionTitle {{ $group['class'] }}" @if (!empty($group['color'])) style="color:{{ htmlspecialchars($group['color']) }}" @endif id="accordion_link_{{ $group['id'] }}">
                     <a href="javascript:void(0)" class="accordion-toggle" id="accordion_toggle_{{ $group['id'] }}" onclick="leantime.snippets.accordionToggle('{{ $group['id'] }}');">
-                        <i class="fa fa-angle-down"></i>{{ $group['label'] }}({{ count($group['items']) }})
+                        <i class="fa fa-angle-down"></i>{!! $group['label'] !!}({{ count($group['items']) }})
                     </a><br />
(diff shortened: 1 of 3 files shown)


marcelfolaron:
Thank you, @mborecki! 🙏 The bug you spotted is real — the type icon was showing as literal markup in the grouped table/list headers.

Before merging I traced where `$group['label']` comes from, and it turned up something bigger. `Tickets::getGroups()` builds that label, and its consumer already renders it raw — `swimlane-row-header.blade.php` does `{!! $label !!}` — because some group types legitimately embed markup (the type icon, the assignee avatar). But only three branches of that switch escape. **Five carry user-controlled text straight into raw output:**

| groupBy | unescaped source |
|---|---|
| `status` | project status labels — user-editable, stored in `zp_settings` |
| `milestoneid` | milestone headline, plus the status name inside `more-info` |
| `editorId` | assignee first/last name |
| `type` | task type value |
| `projectId` | project name |

So that's a live stored-XSS on the kanban board on master today — not something you introduced, but unescaping the three list/table templates here would have extended it to those views too.

I've opened **#3767** to escape all five at the source in `getGroups()`. That's the right boundary, since the method genuinely needs to emit some markup of its own. Once #3767 merges, your change becomes exactly correct and I'll merge this on top — different files, so no conflict and no rebase needed on your side.

Nothing for you to do; just holding this until #3767 lands. Thanks for the report, it found a real security bug.

marcelfolaron:
#3767 is merged, so the escaping now happens at the source in `Tickets::getGroups()` and your change lands on labels that are already safe. Full CI green here.

Merging — thanks again @mborecki. Your report is what surfaced the underlying XSS, which was live on the kanban board and would have gone unnoticed otherwise. 🙏

marcelfolaron (approved):
Correct on top of #3767, which closed the escaping gap at the source. All checks green.

PR #3767, opened 2026-09-07
fix(tickets): escape user-controlled group labels at the source (XSS)

marcelfolaron:
Found while reviewing #3763 (thanks @mborecki — his report is what surfaced this).

## The bug

`Tickets::getGroups()` builds a `label` that its consumer renders **raw** — `swimlane-row-header.blade.php` does `{!! $label !!}` — because some group types legitimately embed markup (the task-type icon, the assignee avatar). That makes the `switch` in `getGroups()` the escaping boundary.

Only three of its branches escaped. **Five carried user-controlled text straight into raw output:**

| groupBy | unescaped source |
|---|---|
| `status` | project status labels — user-editable, stored in `zp_settings` (`getStateLabels()`) |
| `milestoneid` | milestone headline, **plus** the status name interpolated into `more-info` (also rendered raw in the swimlane header) |
| `editorId` | assignee first/last name |
| `type` | task type value |
| `projectId` | project name |

So a milestone headline, project name, assignee name, task type, or a renamed status containing markup executes for every user who opens that kanban board. This is live on master today.

## The fix

Escape each at the source, and document the contract on the switch so the markup the method builds itself (type icon, assignee avatar) stays raw. Also casts `editorId` to `int` in the avatar URL.

Fixing it here rather than in the templates keeps a single boundary — and it is what makes #3763 safe to merge, since that PR unescapes the table/list accordion headers so the type icon renders instead of showing as markup. Merge this first, then #3763 lands cleanly on top (different files, no conflict).

## Verification

- `php -l` clean, Laravel Pint passes
- `priority` / `storypoints` labels are language-file strings, deliberately left raw
- `sprint`, `dependingTicketId` and the `default` branch already escaped — untouched

### DIFF
--- a/app/Domain/Tickets/Services/Tickets.php
+++ b/app/Domain/Tickets/Services/Tickets.php
@@ -768,7 +768,7 @@ class Tickets extends BaseService
                             break;
                         case 'editorId':
                             $editorName = htmlspecialchars(trim($ticket['editorFirstname'].' '.$ticket['editorLastname']), ENT_QUOTES, 'UTF-8');
-                            $label = "<div class='profileImage'><img src='".BASE_URL.'/api/users?profileImage='.(int) $ticket['editorId']."' /></div> ".$editorName;
+                            $label = "<div class='profileImage'><img alt='' src='".BASE_URL.'/api/users?profileImage='.(int) $ticket['editorId']."' /></div> ".$editorName;
 
                             if ($ticket['editorFirstname'] == '' && $ticket['editorLastname'] == '') {
                                 $label = 'Not Assigned to Anyone';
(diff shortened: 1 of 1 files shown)


marcelfolaron:
Addressed the Copilot review.

**Sanitize the `projectId` group id — valid, and it is a second XSS vector.** `\$group['id']` is interpolated into inline JS *string literals*:

```blade
onclick="leantime.snippets.accordionToggle('{{ $group['id'] }}');"
onclick="leantime.kanbanController.toggleSwimlane('{{ $groupId }}')"
```

Blade escapes `'` to `&#039;`, but the HTML parser decodes attribute values *before* the JS engine parses them, so a quote in the id breaks out of the string. The `milestoneid` and `dependingTicketId` branches already run their sort key through `preg_replace('/[^a-zA-Z0-9_-]/', '_', ...)`; `projectId` used the raw project name. Sanitized the same way — which also fixes the DOM-id/accordion breakage Copilot pointed at. Fixed.

**`alt=""` on the avatar** — agreed, it is decorative since the name renders right after. Fixed.

**Unit test** — good idea in principle, but not cheap here: `getGroups()` reaches `getStatusLabels()` → repository → database, and our unit suite runs with no DB configured (`database.default => []`). A meaningful regression test needs either an integration test or extracting the label builder. Tracking as a follow-up rather than blocking a live security fix.

---

Note for @marcelfolaron: the `$group['id']`-into-`onclick` pattern is a **general** weakness, not just this branch. The remaining ids come from `$sortId ?? strtolower($groupedFieldValue)`, so any groupBy whose raw field value reaches the id (e.g. `type`) is in the same JS-string context. The robust fix is to sanitize the id centrally where the group array is assembled, or move the handlers to `data-` attributes and bind them in JS. I have deliberately not done that here — it changes id semantics used by `collapsedSwimlanes` session state and the swimlane filter links, so it deserves its own PR and its own testing.

PR #3749, opened 2026-08-15
fix(tickets): restore board tabs styling and two-row toolbar layout (#3748)

maruf-pfc:
### Description

This PR fixes multiple layout and styling regressions on the ticket board view navigation and action toolbar:

1. **Restored 2-Row Toolbar Architecture**:
   - Moves the left-hand action cluster (`+ New ▾`, `Filter`, `Group By`) back into the dedicated toolbar `<div class="row">` across `showKanban.blade.php`, `showAll.blade.php`, and `showList.blade.php`.
   - Prevents action buttons from colliding with the top view tabs or leaving DataTables buttons (`Export`, `Columns`) orphaned.
2. **Fixed Board Tabs Styling (`ticketBoardTabs.blade.php`)**:
   - Restores the full-width white-to-glass gradient capsule background with backdrop filter matching the design system.
   - Restores active tab white pill highlighting (`background: #fff; box-shadow: 0 1px 4px rgba(0,0,0,0.12); color: var(--accent1); font-weight: 600;`).
   - Restores clean horizontal flex alignment and removes default browser bullet points (`•`) and vertical stacking.
3. **Accessibility / ARIA Sanitization**:
   - Uses `strip_tags()` on the `<nav aria-label="...">` string to prevent escaped HTML icon tags (`&lt;i class='...'&gt;`) from leaking into the DOM.

### Link to ticket

Closes #3748 (https://github.com/Leantime/leantime/issues/3748)

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup 

### Screenshot of the result

image
image

Issue #3748, opened 2026-08-15
fix(tickets): ticket board navigation tabs render unstyled list bullets, broken layout, and escaped HTML in aria-label


maruf-pfc:
### What is your set up?

Self Hosted Docker

### Version

`master` branch (`v3.9.8` development environment)

### Describe the issue

The ticket board view navigation tabs partial (`ticketBoardTabs.blade.php`) renders with a broken layout, unstyled bullet points (`•`), overlapping action buttons, and HTML-escaped tags inside the `<nav>` element's `aria-label`.

### Current Behavior:
1. **Unstyled List Bullets & Vertical Stacking**: The tab links (`Kanban`, `Table`, `List`) are wrapped in raw `<ul>` and `<li>` elements inside `<nav class="lt-tabs-group">`. Because `.lt-tabs-group` expects direct `.lt-tab` links, standard browser default list styles (`list-style: disc`) display bullet points (`•`) and stack the tabs vertically.
2. **Overlapping Toolbar Buttons**: Because `<ul>`/`<li>` break the flexbox container, `.lt-tabs-actions` (`+ New`, `Filter`, `Group By`) gets pulled to the left, overlapping directly on top of the stacked list items.
3. **Escaped HTML in `aria-label`**: `aria-label` uses `{{ __('links.kanban') }}`. Because `links.kanban` contains HTML tags (`"<i class='fas fa-columns'></i> Kanban"`), Blade escapes `<` and `>` into `&lt;` and `&gt;`, placing escaped HTML code inside the accessibility attribute (`aria-label="&lt;i class='fas fa-columns'&gt;&lt;/i&gt; Kanban / ..."`).

### Expected Behavior:
- Tabs (`Kanban`, `Table`, `List`) render as horizontal floating segment pills without bullet points or vertical stacking.
- Active tab displays crisp white pill background (`class="lt-tab on"`).
- Action buttons (`+ New`, `Filter`, `Group By`) align cleanly on the right side without overlapping tabs.
- `<nav>` element's `aria-label` contains plain text (e.g. `aria-label="Kanban / Table / List"`).

**Expected**
<img width="718" height="362" alt="Image" src="https://github.com/user-attachments/assets/04ee6ad0-1740-4183-9dfa-a262480944f9" />
**Current**
<img width="784" height="333" alt="Image" src="https://github.com/user-attachments/assets/69dbd3c9-e187-4178-9e11-1c60e8248ba6" />

### Reproduction steps

[...]

PR #3716, opened 2026-07-29
feat(components): navigation.tabs — one ARIA tab-group + RA chrome joins the core bundle

gloriafolaron:
## What & why

The same tab visual existed in **four hand-rolled copies** (board tabs in nav.css, report deck `.rd-tabs`, goal-dialog `.gv-tabs`, Resource Allocation `.ra-tabs`) with wildly different accessibility — RA had the full ARIA contract, the report deck had none. And RA's entire page skin (841 lines) + its four modals' styles lived inside blade `<style>` blocks, duplicating core's `.rd-scope` design tokens byte-for-byte.

This PR consolidates (design decisions approved by Gloria):

- **`x-global::navigation.tabs` + `navigation.tabs.tab`** (COMPONENTS.md taxonomy slot): ARIA button-tablist — roving tabindex, Arrow/Home/End, optional session/local persistence, `toggle|manual` panel modes, count badges, actions slot, bubbling `lt:tabs:changed` event for pages with custom behavior (deck tracks). Vanilla JS, htmx.onLoad-aware, **no jQuery**.
- **`tab-group.css`**: both approved visuals as variants — `attached` (gradient band + framed translucent group + white pill) and `floating` (RA's glass pills + dark-header scheme).
- **Token reconciliation — RA values canonical**: `--rd-text-3` → `#67747d` (**WCAG AA 4.5:1**; the old `#7a8791` failed), dark `bg`/`panel`/`danger-bg` aligned to RA's curated values.
- **`resource-allocation.css`** joins the core bundle (global-CSS-lives-in-core rule): RA's page skin extracted and deduplicated — tokens now come from `.rd-scope`/`.rd-dark`, the tab bar from the component, `.visually-hidden` from the global utility, the modals' copy-pasted dark-token one-liners are gone, and `.ra-mbtn` gets ONE base + `primary`/`ghost` variants instead of three conflicting per-file definitions.
- The legacy jQuery-UI `x-global::tabs` wrapper is **retired**; its single consumer (plugindetails) migrates. Deliberate markup change, called out per the tracker's rules.

## Merge order
This core PR merges **first**; the plugins companion (`feat/ra-consumes-shared-chrome`) strips the blade style blocks and consumes the component.

## Rollout (separate follow-up PRs)
Goal dialog, board tabs, report deck migrate to the component next; then Task view + Settings per the planned tab-component rollout.

Bundle builds clean (`npx mix`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #3768, opened 2026-09-07
fix(tickets): strip markup from the board/timeline tab aria-labels (#3748)

marcelfolaron:
Closes the genuinely-broken part of #3748. Builds on @maruf-pfc's diagnosis and refactor in #3749 — thanks for finding this.

## What was real

`links.kanban` / `links.table` / `links.list` / `links.timeline` all carry a Font Awesome icon in the translation string:

```ini
links.kanban = "<i class='fas fa-columns'></i> Kanban"
```

The tab body wants that markup, but the `<nav>` aria-label used the same string through `{{ }}`, so the accessible name became escaped markup:

```html
aria-label="&lt;i class='fas fa-columns'&gt;&lt;/i&gt; Kanban / ..."
```

Fixed with `strip_tags` in `ticketBoardTabs` **and** `timelineTabs`, which had the identical bug.

Also carried over from #3749: replacing the `function_exists('findActive')` global helper with a tabs array (two templates declaring the same global function meant whichever rendered first won), and `aria-current="page"` on the active tab.

## What was not real

The other symptoms in #3748 — bullets, vertical stacking, overlapping action buttons — do not reproduce on a built master:

- `tab-group.css` styles this exact `<ul><li><a>` structure under `.lt-tabs--links` (`list-style: none`, `inline-flex` pills) and has been imported in `main.less` and shipped in the bundle since #3716
- the same `lt-tabs lt-tabs--floating lt-tabs--links` combo renders correctly in `timelineTabs`, `plugintabs`, and both Reports views

The screenshots show completely unstyled markup, which is a dev environment running stale assets — `npx mix`.

So #3749's `ticket-board-tabs.css` and toolbar-row revert are intentionally **not** included here: the stylesheet duplicated `tab-group.css` with hardcoded `#fff` / `rgba(255,255,255,…)` values that break dark mode, bypassed the Mix pipeline (no version stamp, no cache busting), and moving board actions out of the nav bar would have made Tickets the only one of five `lt-tabs` consumers with a separate toolbar row.

### DIFF
--- a/app/Domain/Tickets/Templates/submodules/ticketBoardTabs.blade.php
+++ b/app/Domain/Tickets/Templates/submodules/ticketBoardTabs.blade.php
@@ -1,15 +1,7 @@
 @php
     use Leantime\Core\Controller\Frontcontroller;
 
-    if (!function_exists('findActive')) {
-        function findActive($route): string
-        {
-            if (str_contains(Frontcontroller::getCurrentRoute(), $route)) {
-                return 'active';
-            }
-            return '';
-        }
-    }
+    $currentRoute = Frontcontroller::getCurrentRoute();
 
     // Program boards inject their own kanban/table/list URLs + the route fragments used to
     // highlight the active tab. Per-project views fall back to the core /tickets/* routes.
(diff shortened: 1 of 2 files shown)


marcelfolaron:
Addressed the Copilot review — both comments were valid. `strip_tags("<i …></i> Kanban")` leaves a leading space, so the accessible name became `" Kanban / Table / List"`. Wrapped both in `trim()`.

PR #3769, opened 2026-09-07
chore(deps): clear all 9 high-severity npm advisories

marcelfolaron:
The `NPM Security Audit` check has been failing on master, so it is red on every PR and currently carries no signal. This makes it pass on merit.

**111 vulnerabilities (6 low, 96 moderate, 9 high) → 62 (5 low, 57 moderate, 0 high).**

## 1. `npm audit fix`, non-breaking only — clears 8 of 9

| package | before | after |
|---|---|---|
| postcss | 8.5.15 | 8.5.28 |
| less | 4.3.0 | 4.9.1 |
| svgo | 2.8.2 | 2.8.4 |
| browserslist | 4.28.1 | 4.28.9 |
| js-yaml | 4.3.0 | 4.3.2 |
| fast-uri | 3.1.2 | 3.1.7 |
| brace-expansion | 1.1.16, 2.1.2 | 1.1.18, 2.1.4 |
| image-size | 0.5.5 | *removed* (less 4.9.1 dropped it) |

All patch or minor.

## 2. nanoid via uppy — needed a decision

uppy 3.27.4 pins `nanoid@4.0.2`, and **the 4.x line has no patched release**: it ends at 4.0.2 and every advisory covers all of it. npm’s suggested fix is `uppy@6.0.1` — three majors up, a large change to the file uploader.

uppy only ever imports `nanoid/non-secure`, and nanoid 5.1.16 still exports that subpath, so a **scoped override** to `^5.1.16` closes it without touching uppy:

```json
"overrides": {
  "@uppy/aws-s3":         { "nanoid": "^5.1.16" },
  "@uppy/core":           { "nanoid": "^5.1.16" },
  "@uppy/dashboard":      { "nanoid": "^5.1.16" },
  "@uppy/provider-views": { "nanoid": "^5.1.16" },
  "@uppy/store-redux":    { "nanoid": "^5.1.16" },
  "@uppy/url":            { "nanoid": "^5.1.16" }
}
```

A blanket `nanoid` override would **not** work — postcss requires CJS `nanoid@^3` and nanoid 5 is ESM-only, so it would break the build. postcss stays on the patched `3.3.18`.

Scoping to `uppy` alone also does not work: npm applies that only to uppy’s direct deps, and the `@uppy/*` sub-packages are what actually require nanoid. Hence the six entries.

This one is not audit noise — [GHSA-28wg-ghj8-5hjv](https://github.com/advisories/GHSA-28wg-ghj8-5hjv) and [GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8) are both about the **non-secure generator looping indefinitely**, which is exactly the entry point uppy uses.

## Verification

- `npm ci` reproduces the lockfile cleanly → `62 vulnerabilities (5 low, 57 moderate)`, zero high
- `npx mix --production` → **Compiled successfully in 25.13s**
- uppy and nanoid’s `urlAlphabet` both present in the emitted bundle, so the ESM swap resolved rather than silently dropping
- CSS builds under the new less (`main.min.css` 884 KB, `tab-group.css` rules present)
- the one remaining webpack child-compilation warning is **identical to the one master already emits** — pre-existing, not introduced here

Deliberately **not** done: no `continue-on-error`, no relaxing of `--audit-level=high`. Those would turn the check green without changing exposure.

## One thing to be aware of

A successful webpack build proves nanoid 5 resolves and bundles — not that the uploader works at runtime. uppy calls `nanoid/non-secure` to mint file IDs; the subpath exists in v5 and the signature is unchanged, so I expect it is fine, but **exercising one real file upload after merge is the check that actually proves it**.

The 57 remaining moderates are mostly build tooling (webpack-dev-server, laravel-mix) and the tiptap `mergeAttributes` advisory, whose fix is a breaking tiptap bump. Those are left for a separate decision.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1633,9 +1633,10 @@
       }
     },
     "node_modules/@braintree/sanitize-url": {
-      "version": "7.1.1",
-      "resolved": "https://registry.npmjs.org/@braintree/sanitize-url/-/sanitize-url-7.1.1.tgz",
-      "integrity": "sha512-i1L7noDNxtFyL5DmZafWy1wRVhGehQmzZaz1HiN5e7iylJMSZR7ekOV7NsIqa5qBldlLrsKv4HbgFUVlQrz8Mw=="
+      "version": "7.1.2",
+      "resolved": "https://registry.npmjs.org/@braintree/sanitize-url/-/sanitize-url-7.1.2.tgz",
+      "integrity": "sha512-jigsZK+sMF/cuiB7sERuo9V7N9jx+dhmHHnQyDSVdpZwVutaBu7WvNYqMDLSgFgfB30n452TP3vjDAvFC973mA==",
+      "license": "MIT"
     },
     "node_modules/@chevrotain/types": {
       "version": "11.1.2",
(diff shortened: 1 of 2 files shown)


marcelfolaron:
Good catch from Copilot, and it is a real consequence of this PR rather than a pre-existing gap — so I applied it.

The floor genuinely moves. Before: `less@4.3.0` `">=14"`, `nanoid@4.0.2` `"^14 || ^16 || >=18"`. After: `less@4.9.1` `">=18"`, `nanoid@5.1.16` `"^18 || >=20"`. So this PR raises the requirement from Node 14 to 18 and nothing declared it — a contributor on an older Node would have hit confusing install/build failures instead of a clear message.

Added `"engines": { "node": ">=18" }`. Verified it is advisory only: there is no `.npmrc` in the repo and `engine-strict` is not set anywhere, so npm warns rather than failing the install. CI already pins Node 20, so nothing there changes.

PR #3784, opened 2026-09-12
fix(notifications): send Mattermost attachment fields as an array

### LINKED ISSUE #3131: Question about Mattermost integration
I just set up Leantime today and so far it looks great. However, I’m having trouble getting the Mattermost integration to work.

- I added an incoming webhook in Mattermost and configured it in the project settings in Leantime. 
- No changes in the project—such as creating or updating to-dos—are being sent to the Mattermost webhook.
- I tested the webhook with a curl request, and it worked, so the webhook itself is functional.

I couldn’t find any settings in Leantime to control when or for which events Mattermost notifications are sent. Changing or adding project to-dos does not trigger any messages.

I also manually ran the scheduler, but that did not make a difference.

For reference, I’m using the latest Docker Compose setup on a server with Nginx Proxy Manager serving a separate domain. Everything else in Leantime is working perfectly—just the Mattermost integration is not sending notifications.

Any guidance or suggestions would be greatly appreciated.

BeforeMyCompileFails:
### Description

Mattermost requires `attachments[].fields` to be an **array** of field objects. `prepareMessage()` builds it as a single object, so Mattermost rejects the whole payload with `400 web.incoming_webhook.decode.app_error` and no notification is ever delivered.

Isolated against the same incoming webhook, changing only the `fields` shape:

| payload | Mattermost response |
| --- | --- |
| `"fields": { ... }` (current) | `400 web.incoming_webhook.decode.app_error` |
| `"fields": [ { ... } ]` | `200 ok` |

Slack accepts the object form, which is why this only surfaces on Mattermost. `prepareMessage()` is shared with the Slack webhook, and the array form is what both Slack and Mattermost document, so Slack is unaffected.

This also adds `Content-Type: application/json` to the Mattermost request — `mattermostWebhook()` was the only messenger not sending it; slack, zulip, telegram and discord all pass the header already.

Verified against Mattermost 11.9.0 / Leantime 3.9.8 / PHP 8.3: before the change every to-do create/update was rejected with a 400; after it, messages post to the channel immediately.

One note for triage on #3131, where the discussion had moved toward the queue worker: messenger delivery is **not** queued. `sendNotificationToMessengers()` is called synchronously from `Projects.php:358`, immediately after the *email* queue call, so `schedule:run` has no bearing on this path.

### Link to ticket

Fixes #3131

### Type

- [x] Fix
- [ ] Feature
- [ ] Cleanup

### Screenshot of the result

No user interface changes — this affects the outbound webhook payload only.

### DIFF
--- a/app/Domain/Notifications/Services/Messengers.php
+++ b/app/Domain/Notifications/Services/Messengers.php
@@ -142,6 +142,7 @@ class Messengers
                 $this->httpClient->post($mattermostWebhookURL, [
                     'allow_redirects' => OutboundUrlGuard::redirectOptions(),
                     'body' => $data_string,
+                    'headers' => ['Content-Type' => 'application/json'],
                 ]);
 
                 return true;
@@ -532,7 +533,7 @@ class Messengers
                 'pretext' => $notification->message,
                 'title' => $headline,
                 'title_link' => $notification->url['url'],
-                'fields' => $fields,
+                'fields' => [$fields],
             ],
         ];
 
--- a/tests/Unit/app/Domain/Notifications/Services/MessengersServiceTest.php
+++ b/tests/Unit/app/Domain/Notifications/Services/MessengersServiceTest.php
@@ -10,6 +10,7 @@ use Leantime\Core\Language as LanguageCore;
 use Leantime\Domain\Notifications\Models\Notification as NotificationModel;
 use Leantime\Domain\Notifications\Services\Messengers;
 use Leantime\Domain\Setting\Repositories\Setting as SettingRepository;
+use Leantime\Domain\Tickets\Services\Tickets;
 use Unit\TestCase;
 
 class MessengersServiceTest extends TestCase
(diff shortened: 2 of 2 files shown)


BeforeMyCompileFails:
Good catch - added `test_mattermost_webhook_sends_fields_as_array_with_json_content_type` in the second commit. It stubs the Guzzle client, captures the options, and asserts that `$capturedOptions['headers']['Content-Type']` is `application/json` and that `array_is_list()` holds for `attachments[0].fields` in the decoded body. It also pins the rendered `title`, `value` and `short`, so a change to `prepareMessage()` cannot quietly drop them.

Two notes on keeping it a real unit test. The webhook URL is an IP literal (`203.0.113.10`, TEST-NET-3) because `OutboundUrlGuard::isAllowedUrl()` short-circuits on IP literals, so the test performs no DNS lookup. And `prepareMessage()` resolves `Tickets` from the container for `getStatusLabels()`, so the test binds a stub via `$this->app->instance(Tickets::class, ...)` rather than touching the database.

Both changed files lint clean under PHP 8.3. I was not able to run `make unit-test` against the full suite locally, so approving the workflow run would be the real confirmation.

One cosmetic wart: the second commit's message got garbled while I was writing it. The diff itself is correct and clean - happy to squash or reword if you would rather not carry that into history.

BeforeMyCompileFails:
Cleaned that up - force-pushed `aa1629e` over `0ea3d58`. It is a message-only rewrite: same tree, same parent, empty diff against the previous head, so no squash needed on your side. The amended commit is unsigned since I rewrote it locally; the fix commit is still GitHub-signed.

